PR #354, opened 2026-02-04
Add Qdrift and partial scheme

nabbelbabbel:


python/src/qdk_chemistry/algorithms/time_evolution/builder/partially_randomized.py
```diff
@@ -0,0 +1,391 @@
...
+    This hybrid approach benefits from:
+    - Better ε-scaling from deterministic treatment of dominant terms
+    - Reduced circuit depth from random sampling of the long tail
+    """
+
+    def __init__(self):
+        """Initialize PartiallyRandomizedSettings with default values.
+
+        Attributes:
+            num_deterministic_terms: Number of largest-weight terms to treat deterministically.
```
wavefunction91: It seems that these should be mutually exclusive. I'd err on the side of the latter, as having a fixed number could lead to unexpected results, and it would be better to encourage users to use a scalable approach. I don't think we have a concept for mutual exclusivity, do we?

python/src/qdk_chemistry/algorithms/time_evolution/builder/partially_randomized.py
```diff
@@ -0,0 +1,391 @@
...
+            return max(1, count)  # At least 1 deterministic term
+
+        # Default: top 10% of terms (at least 1, at most all but 1 for random)
+        num_det = max(1, len(terms_data) // 10)
+        # Ensure at least some terms remain for random treatment
+        if num_det >= len(terms_data):
+            num_det = max(1, len(terms_data) - 1)
+        return num_det
+
+    def _build_random_terms(
```
wavefunction91: We should have a common set of functionality that unifies this with the qDRIFT implementation as to not repeat code.

python/src/qdk_chemistry/algorithms/time_evolution/builder/qdrift.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Examples:
+        >>> from qdk_chemistry.algorithms import create
+        >>> # Create a qDRIFT builder with 500 samples
+        >>> qdrift = create("time_evolution_builder", "qdrift", num_samples=500, seed=42)
+        >>> # Use it to build time evolution for a Hamiltonian
+        >>> time_evolution = qdrift.run(qubit_hamiltonian, time=1.0)
+
+    References:
+        Campbell, E. (2019). Random Compiler for Fast Hamiltonian Simulation.
```
wavefunction91: Is there a paper DOI for this?

python/src/qdk_chemistry/algorithms/time_evolution/builder/qdrift.py
```diff
@@ -0,0 +1,242 @@
...
+        num_samples: int = self._settings.get("num_samples")
+        tolerance: float = self._settings.get("tolerance")
+
+        # Extract Hamiltonian terms and coefficients
+        paulis = list(qubit_hamiltonian.pauli_ops.paulis)
+        coeffs_raw = qubit_hamiltonian.pauli_ops.coeffs
+
+        # Convert coefficients to real values and validate Hermiticity
+        coeffs: list[float] = []
+        for idx, coeff in enumerate(coeffs_raw):
```
wavefunction91: This validation is in several places - I'm inclined to move this to a reusable utility. Maybe `QubitHamiltonian::is_hermitian`?

python/tests/test_time_evolution_qdrift.py
```diff
@@ -0,0 +1,249 @@
+"""Tests for qDRIFT randomized time evolution builder in QDK/Chemistry."""
```
wavefunction91: Why are there two sets of qdrift tests with the same heading?

nabbelbabbel: Consolidated this

python/src/qdk_chemistry/algorithms/time_evolution/builder/partially_randomized.py
```diff
@@ -0,0 +1,391 @@
...
+        self._settings.set("seed", seed)
+        self._settings.set("tolerance", tolerance)
+
+    def _run_impl(self, qubit_hamiltonian: QubitHamiltonian, time: float) -> TimeEvolutionUnitary:
+        r"""Construct the time evolution unitary using partially randomized product formula.
+
+        The algorithm:
+        1. Sort Hamiltonian terms by coefficient magnitude
+        2. Split into H_D (deterministic, large terms) and H_R (random, small terms)
+        3. Apply second-order Trotter structure with qDRIFT for the H_R block
```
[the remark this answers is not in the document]
nabbelbabbel: Fixed

python/src/qdk_chemistry/algorithms/time_evolution/builder/qdrift.py
```diff
@@ -0,0 +1,242 @@
...
+                    num_qubits=qubit_hamiltonian.num_qubits,
+                )
+            )
+
+        # Compute λ (1-norm) and probabilities
+        abs_coeffs = np.array([abs(c) for c in filtered_coeffs])
+        lambda_norm = float(np.sum(abs_coeffs))
+        probabilities = abs_coeffs / lambda_norm
+
+        # Sample N term indices according to the probability distribution
```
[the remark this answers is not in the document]
nabbelbabbel: According to my quick check merging should work for already grouped commuting terms, shuffling non-commuting terms may alter the bound. I added an option to merge grouped commuting terms.

PR #358, opened 2026-02-09
S1 entropies and mutual information in MACIS

### LINKED ISSUE #139: Add Mutual Information as a Wavefunction Attribute
We currently only support orbital entropies as our information theoretic quantity. Mutual information would also be useful, and is commonly used in the quantum information literature. This requires selective evaluation of elements of the 3- and 4-rdms, which will require some additional infrastructure.

mmoerchen:
This PR implement the direct computation of mutual information and single orbital entropies in MACIS. 
Additionally, the wavefunction class is expanded to accept pre-computed s1 entropies and mutual information at construction. 

Resolves #139


cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -173,11 +173,19 @@ class WavefunctionContainer {
...
+   * (optional)
    * @param type The type of wavefunction
    */
-  WavefunctionContainer(const std::optional<MatrixVariant>& one_rdm_spin_traced,
-                        const std::optional<VectorVariant>& two_rdm_spin_traced,
-                        WavefunctionType type = WavefunctionType::SelfDual);
+  WavefunctionContainer(
+      const std::optional<MatrixVariant>& one_rdm_spin_traced,
+      const std::optional<VectorVariant>& two_rdm_spin_traced,
+      const std::optional<Eigen::VectorXd>& single_orbital_entropies =
```
wavefunction91: Adding this feels like a slippery slope. I'd prefer that we keep constructors to a low number of parameters. Perhaps it might make more sense to encapsulate this information into another container that can be passed in here do consolidate? I suppose the same goes for the RDMs, but we can punt on that for now.

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -281,19 +297,41 @@ class WavefunctionContainer {
...
+   *
+   * @return True if mutual information matrix is available, false otherwise
+   */
+  virtual bool has_mutual_information() const;
+
+  /**
+   * @brief Get the mutual information matrix for active orbitals
+   *
+   * The mutual information is defined as:
+   * \f$ I_{ij} = s_{1,i} + s_{1,j} - s_{2,ij} \f$
```
wavefunction91: Perhaps it makes more sense to store the two-orbital entropies and evaluate this lazily?

cpp/src/qdk/chemistry/algorithms/microsoft/macis_base.hpp
```diff
@@ -161,6 +167,25 @@ inline data::Wavefunction build_wavefunction(
     }
   }
 
+  // evaluate single orbital entropies and mutual information
+  std::optional<Eigen::VectorXd> computed_s1;
+  std::optional<Eigen::MatrixXd> computed_mi;
+  if (eval_s1 || eval_mutual_info) {
```
wavefunction91: Is the cost negligible to evaluate both s1 and mi? If not, then this should be finer grained as to not trigger excessive work unnecessarily based on user request.

cpp/src/qdk/chemistry/data/wavefunction_containers/sci.cpp
```diff
@@ -34,8 +34,7 @@ SciWavefunctionContainer::SciWavefunctionContainer(
                                std::nullopt,  // two_rdm_aabb
                                std::nullopt,  // two_rdm_aaaa
                                std::nullopt,  // two_rdm_bbbb
-                               std::nullopt,  // single_orbital_entropies
-                               std::nullopt,  // mutual_information
+                               {},            // entropies
```
wavefunction91: Unnamed brace initializer should be avoided. Now sure if this shows up other places in the PR, but it should be something like `OrbitalEntropies{}`

PR #384, opened 2026-03-02
2nd order Trotter and Corresponding Error Bounds

v-agamshayit:
Added support for second-order Trotter time evolution. Implemented the Naive and commutator-based error bounds for the second-order Trotter formula and added some tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -219,3 +219,9 @@ __marimo__/
 # PYIs
 python/src/qdk_chemistry/_core/*.pyi
 python/src/qdk_chemistry/algorithms/registry.pyi
+
+# Generated at install time by scikit-build
+python/VERSION
+
+# VS Code
+.vscode/
--- a/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.cpp
+++ b/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.cpp
@@ -83,6 +83,21 @@ std::tuple<std::vector<double>, size_t> compute_cholesky_vectors(
   const size_t num_shells = obs.size();
   const size_t num_shell_pairs = num_shells * (num_shells + 1) / 2;
 
+  // Cholesky decomposition (rank is bounded by num_aos*(num_aos+1)/2)
+  const size_t max_rank = num_aos * (num_aos + 1) / 2;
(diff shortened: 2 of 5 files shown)


python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
```diff
@@ -139,10 +139,12 @@ def _run_impl(self, qubit_hamiltonian: QubitHamiltonian, time: float) -> TimeEvo
 
         """
         if self._settings.get("order") == 1:
-            return self._first_order_trotter(qubit_hamiltonian, time)
-        raise NotImplementedError("Only first-order Trotter decomposition is currently supported.")
+            return self._first_or_second_order_trotter(qubit_hamiltonian, time)
+        if self._settings.get("order") == 2:
+            return self._first_or_second_order_trotter(qubit_hamiltonian, time)
```
wavefunction91: Now that we have more than first order, we should build out proper "order dependence" in this function. I.e. create a function that can be called recursively with an order and time parameter with truncating conditions at 1 and 2. This will allow for the expression of arbitrary order Trotter in a future PR.

python/src/qdk_chemistry/utils/pauli_commutation.py
```diff
@@ -186,6 +187,77 @@ def get_commutation_checker(
     raise ValueError(f"Unknown commutation_type {commutation_type!r}; expected 'general' or 'qubit_wise'.")
 
 
+def pauli_product_label(label_a: str, label_b: str) -> str:
```
wavefunction91: We already have pauli algebra to do this.

python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
```diff
@@ -260,7 +262,8 @@ def _decompose_trotter_step(
                 mapping = self._pauli_label_to_map(label)
                 angle = coeff * time / 2
                 terms.append(ExponentiatedPauliTerm(pauli_term=mapping, angle=angle))
-
+        else:
+            raise NotImplementedError("Only orders 1 or 2 are currently supported.")
```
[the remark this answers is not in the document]
v-agamshayit: Fixed

PR #389, opened 2026-03-04
Arbitrary Order Trotter

v-agamshayit:
This pull requests adds support for arbitrary order Trotter decompositions using the usual recursive definition. The computation of the terms and angles is done bottom-up using dynamic programming, which circumvents the exponential runtime complexity associated with recursion.

The implemented naive and commutator-based error bounds are extended to accommodate arbitrary p-order Trotter formulae. We use the asymptotic error estimate of Childs et.al 2021 (Theorem 6 and Corollary 7). New tests were added to ensure that the computed high-order formulae and their error bounds are correct.

### DIFF
--- a/python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
+++ b/python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
@@ -28,6 +28,7 @@ from qdk_chemistry.data.time_evolution.containers.pauli_product_formula import (
     ExponentiatedPauliTerm,
     PauliProductFormulaContainer,
 )
+from qdk_chemistry.utils import Logger
 
 __all__: list[str] = ["Trotter", "TrotterSettings"]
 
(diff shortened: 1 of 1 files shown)


python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
```diff
@@ -138,17 +138,27 @@ def _run_impl(self, qubit_hamiltonian: QubitHamiltonian, time: float) -> TimeEvo
             TimeEvolutionUnitary: The time evolution unitary built by the Trotter decomposition.
 
         """
-        if self._settings.get("order") in {1, 2}:
+        order = self._settings.get("order")
+        if order in {1, 2} or (order > 2 and order % 2 == 0):
             return self._trotter(qubit_hamiltonian, time)
-        raise NotImplementedError("Only orders 1 or 2 are currently supported.")
+        raise NotImplementedError("Non-positive and higher odd orders are not supported.")
```
wavefunction91: This is not a clear error message. Something like `"Trotter orders must be positive and even for orders greater than 1"

python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
```diff
@@ -261,6 +276,56 @@ def _decompose_trotter_step(
                 angle = coeff * time / 2
                 terms.append(ExponentiatedPauliTerm(pauli_term=mapping, angle=angle))
 
+            # Construct order 2k formula bottom up dynamic-programming style
+            if order > 2 and order % 2 == 0:
```
wavefunction91: Shouldn't this modulo condition be covered by the check above?

python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter_error.py
```diff
@@ -72,20 +77,32 @@ def trotter_steps_naive(
...
         raise NotImplementedError(
             f"Trotter step estimation for order {order} is not yet implemented. "
-            "Only orders 1 or 2 are currently supported."
+            "Non-positive and higher odd orders are not supported."
         )
     real_terms = hamiltonian.get_real_coefficients(tolerance=weight_threshold)
     one_norm = sum(abs(coeff) for _, coeff in real_terms)
+    if order == 1:
+        return max(1, math.ceil(((2 * one_norm**2) * time**2) / target_accuracy))
+    if order == 2:
```
wavefunction91: Is there a reason to separate this from the HO expression?

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/time_evolution/builder/trotter.py
RushiGong: Is the filtered-coefficients check required for both order = 1 and higher orders? It might also be helpful to log a warning if filtering removes all coefficients and return nothing.

Issue #388, opened 2026-03-04
Propagate qubit types and counts in the Circuit


RushiGong:
We should figure out how to propagate the intention for various classes of qubits (system, ancilla, classical) within the `Circuit` object itself. Please open an issue on this.

_Originally posted by @wavefunction91 in https://github.com/microsoft/qdk-chemistry/pull/363#discussion_r2881083041_

PR #405, opened 2026-03-16
Add Hadamard test algorithm

BoqinZhang:


python/src/qdk_chemistry/algorithms/hadamard_test_generator/base.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def type_name(self) -> str:
+        """Return the algorithm type name as hadamard_test_generator."""
+        return "hadamard_test_generator"
+
+    @abstractmethod
+    def _run_impl(
+        self,
+        state_preparation: Circuit,
+        num_system_qubits: int,
```
wavefunction91: @RushiGong Is this something we need to pass. How are you handling this elsewhere?

wavefunction91 (changes requested):

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/hadamard_test_generator/hadamard_test_generator.py
```diff
@@ -0,0 +1,169 @@
...
+
+from qdk_chemistry.algorithms.hadamard_test_generator.base import HadamardTestGenerator
+from qdk_chemistry.data import Circuit
+from qdk_chemistry.utils import Logger
+from qdk_chemistry.utils.qsharp import QSHARP_UTILS
+
+__all__: list[str] = ["QiskitHadamardGenerator", "QsharpHadamardGenerator"]
+
+
+class QsharpHadamardGenerator(HadamardTestGenerator):
```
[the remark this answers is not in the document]
RushiGong: Why? It sounds not reasonable for user to do a separate function for post-processing measurement results... Why not make executor an arg and make this algorithm end-to-endl? If I run a `Hadamard test algorithm`, I would expect run the simulation and get the measured value, not just getting a circuit ... @wavefunction91

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
```diff
@@ -0,0 +1,178 @@
...
+    def type_name(self) -> str:
+        """Return the algorithm type name as hadamard_test."""
+        return "hadamard_test"
+
+    def _run_impl(
+        self,
+        state_preparation_circuit: Circuit,
+        time_evolution_unitary: TimeEvolutionUnitary,
+        shots: int,
+        unitary_power: int,
```
RushiGong: These arguments may constrain the algorithm classes. For example, `qdk_full_state_simulator` supports options for selecting the simulator type, but those are not exposed here. A similar issue exists for the circuit mapper: we currently only support `pauli_sequence`, but future mappers may require mapper‑specific configuration options. It may be more flexible to pass in the Executor or Mapper algorithm instances directly, rather than restricting the interface to fixed arguments.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
BoqinZhang: `pauli_sequence` is just the current default algorithm for mapper. They can be modified by the users. For example, in the future, if there is another supported type of `controlled_evolution_circuit_mapper`, and the user wants to use it, the user can just input its name at here.
I feel that if we input the instance of the `controlled_evolution_circuit_mapper` and `CircuitExecutor`, they have to be mandatory inputs without defaults, then calling `_run_impl` will be too complicated for the user: the two instances have to be prepared before the call.
If the user doesn't care about how to build the map and how to run the circuit, then the user can just input the four mandatory inputs. I guess it may be easier for them to use `HadamardTest`?

BoqinZhang: I have modified the interface now: the input of mapper and simulator should be the instances, but the user doesn't have to input them: if so, then the default mapper and simulator will be created inside `_run_impl`.

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
```diff
@@ -0,0 +1,176 @@
...
+        mapper: ControlledEvolutionCircuitMapper | None = None,
+        simulator: CircuitExecutor | None = None,
+        simulator_seed: int = 42,
```
RushiGong: `seed` can be removed from this level since it can be defined within `CircuitExecutor`.

python/src/qdk_chemistry/algorithms/hadamard_test/hadamard_test.py
```diff
@@ -0,0 +1,195 @@
...
+        from qdk import qsharp as _qsharp  # noqa: PLC0415
+    except ModuleNotFoundError as err:
+        raise ModuleNotFoundError(
+            "qdk.qsharp is required to convert Hadamard test bases into qsharp.Pauli values."
+        ) from err
+
+    return getattr(_qsharp.Pauli, basis.value)
+
+
+class HadamardTest(Algorithm):
```
YingrongChen, line 76: We might want a `HadamardTestCircuitBuilder` that in the `init` takes a unitary_builder and a circuit_mapper, then in `run_impl` takes a state_preparation_circuit and a qubit_hamiltonian.
see [iterative_builder](https://github.com/microsoft/qdk-chemistry/blob/main/python/src/qdk_chemistry/algorithms/phase_estimation/circuit_builder/iterative_builder.py)

python/src/qdk_chemistry/utils/qsharp/HadamardTest.qs
```diff
@@ -0,0 +1,47 @@
...
+    /// - `statePrep`: A function to prepare the initial quantum state.
+    /// - `repControlledEvolution`: A function to perform repeated controlled evolution.
+    /// - `testBasis`: Measurement basis for the control qubit. Supported values are `PauliX`, `PauliY`, and `PauliZ`.
+    /// - `control`: The index of the control qubit in the allocated register.
+    /// - `systems`: An array of indices representing the system qubits.
+    /// # Returns
+    /// A single-element result array containing the control-qubit measurement in the selected basis.
+    operation HadamardTest(
+        statePrep : Qubit[] => Unit,
+        repControlledEvolution : (Qubit, Qubit[]) => Unit,
```
YingrongChen, line 20: We will need to pass in the number of phase qubits, system qubits, and potential ancillary qubits needed by some repControlledEvolution, see [block encoding PR](https://github.com/microsoft/qdk-chemistry/blob/7f5c4715d62cc3c72093993548d751903946ea76/python/src/qdk_chemistry/utils/qsharp/IterativePhaseEstimation.qs#L22)

python/src/qdk_chemistry/algorithms/hadamard_test/hadamard_test.py
BoqinZhang: The `run_impl` has been reimplemented to resolve the conflicts. I think now the structure of it is the one you suggested.

python/src/qdk_chemistry/utils/qsharp/HadamardTest.qs
BoqinZhang: Emm, it seems that phase qubit is only used for IQPE? Does general Hadamard test algorithm need it?

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/base.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self):
+        """Initialize the Hadamard test circuit builder."""
+        super().__init__()
+
+    def type_name(self) -> str:
+        """Return the algorithm type name as hadamard_test_circuit_builder."""
+        return "hadamard_test_circuit_builder"
+
+    @abstractmethod
+    def _run_impl(
```
RushiGong, line 90: The `run` seems unclear here. I think it should take a dataclass as input, such as `state_preparation` and `unitary_representation`. The test basis can be set via settings. Also, it's unclear why the user needs to provide `num_system_qubits` and `num_ancilla_qubits`, these should be derivable from the inputs.

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/qdk_builder.py
```diff
@@ -0,0 +1,80 @@
...
+        return Circuit(
+            qsharp_factory=QsharpFactoryData(
+                program=QSHARP_UTILS.HadamardTest.HadamardTest,
+                parameter=hadamard_parameters,
+            )
+        )
+
+    def name(self) -> str:
+        """Return the name of the QdkHadamardTestCircuitBuilder algorithm."""
+        return "qdk_circuit_builder"
```
RushiGong: The naming could be simplified. `circuit_builder` seems to be redundant with algorithm type name `hadamard_test_circuit_builder`.

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
```diff
@@ -0,0 +1,172 @@
...
+        the controlled circuit mapper and the circuit executor.
+
+        """
+        super().__init__()
+        self._set_default(
+            "circuit_builder",
+            "algorithm_ref",
+            AlgorithmRef("hadamard_test_circuit_builder", "qdk_circuit_builder"),
+        )
+        self._set_default(
```
RushiGong, line 57: I think `controlled_circuit_mapper` would fit better as a setting on the circuit builder rather than the hadamard test, since it's more related to how the circuit is constructed.

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
```diff
@@ -0,0 +1,172 @@
...
+        """Return the algorithm type name as hadamard_test."""
+        return "hadamard_test"
+
+    def _run_impl(
+        self,
+        state_preparation_circuit: Circuit,
+        unitary: UnitaryRepresentation,
+        shots: int,
+        test_basis: HadamardTestBasis = HadamardTestBasis.X,
+        num_ancilla_qubits: int = 0,
```
RushiGong: See the comment above regarding the `run` arguments.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/hadamard_test/base.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        if not isinstance(test_basis, HadamardTestBasis):
+            raise TypeError("test_basis must be an instance of HadamardTestBasis.")
+        if not isinstance(unitary, UnitaryRepresentation):
+            raise TypeError("unitary must be an instance of UnitaryRepresentation.")
+        num_system_qubits = unitary.get_num_qubits()
+        if not isinstance(shots, int):
+            raise TypeError("shots must be an integer.")
+        if shots <= 0:
+            raise ValueError("shots must be a positive integer.")
```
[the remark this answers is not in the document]
BoqinZhang: This comment is WRONG!

python/tests/test_hadamard_test.py
```diff
@@ -0,0 +1,257 @@
...
+
+_QDK_BUILDER = "qdk_circuit_builder"
+_QISKIT_BUILDER = "qiskit_circuit_builder"
+
+
+def _make_hadamard_test(builder_name: str):
+    """Create a Hadamard test configured to use the given circuit builder backend."""
+    hadamard_test = create("hadamard_test")
+    hadamard_test.settings().set("circuit_builder", AlgorithmRef("hadamard_test_circuit_builder", builder_name))
+    return hadamard_test
```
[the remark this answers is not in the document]
BoqinZhang: Redundant: default seed is set in simulator

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/qdk_builder.py
```diff
@@ -0,0 +1,98 @@
...
+        Args:
+            state_preparation_circuit: Circuit that prepares the trial state on system qubits.
+            unitary: Unitary representation :math:`U` (e.g. a time-evolution unitary built with the desired power).
+
+        Returns:
+            Circuit containing compiled and rendered Q# Hadamard test artifacts.
+
+        """
+        test_basis = HadamardTestBasis(self._settings.get("test_basis"))
+        num_ancilla_qubits = self._settings.get("num_ancilla_qubits")
```
RushiGong: The `num_ancilla_qubits` is still unclear from the user's perspective. Both state preparation and unitary circuits may require ancilla qubits, and the qubit types and counts need to be carefully managed within the circuit data class #388 . Could you please remove the `num_ancilla_qubits` parameter from this PR? There will be another one to address this.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/qdk_builder.py
BoqinZhang: @YingrongChen What do you think of it? I am also confused now: maybe the ancilla qubits needed by block encoding should just be a part of the input unitary circuit, and not be an input of HadamardTest class?

YingrongChen: I agree with Rushi, sorry I wasn't aware that there's an issue on this already

BoqinZhang: extra ancilla qubits are removed from HadamardTest

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/base.py
```diff
@@ -0,0 +1,134 @@
...
+            "algorithm_ref",
+            AlgorithmRef("controlled_circuit_mapper", "pauli_sequence"),
+        )
+        self._set_default(
+            "test_basis",
+            "string",
+            HadamardTestBasis.X.value,
+            "Measurement basis for the control qubit ('X', 'Y', or 'Z').",
+            [basis.value for basis in HadamardTestBasis],
+        )
```
wavefunction91, line 43: Are these the only possible options? Is this really a discrete set? @feisuxiaozhu?

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/base.py
```diff
@@ -0,0 +1,134 @@
...
+    "HadamardTestCircuitBuilderSettings",
+]
+
+
+class HadamardTestCircuitBuilderSettings(Settings):
+    """Settings for the Hadamard test circuit builder algorithm.
+
+    Includes the nested algorithm reference for the controlled circuit mapper
+    used to synthesize the controlled evolution circuit and the measurement
+    basis for the control qubit.
```
wavefunction91: These are not necessary.

python/src/qdk_chemistry/algorithms/hadamard_test/hadamard_test.py
```diff
@@ -0,0 +1,169 @@
...
+    """Hadamard test generator.
+
+    Orchestrates the backend-agnostic Hadamard test workflow: it validates the
+    inputs, delegates the backend-specific circuit construction (including
+    mapping the target unitary into a controlled evolution circuit) to the
+    nested ``hadamard_test_circuit_builder``, and executes the resulting circuit
+    with the nested ``circuit_executor``.
+    """
+
+    def __init__(
```
wavefunction91, line 86: Why does this not init all of the settings?

python/src/qdk_chemistry/plugins/qiskit/hadamard_test_circuit_builder.py
```diff
@@ -0,0 +1,132 @@
+"""Qiskit Hadamard test circuit builder implementation."""
```
wavefunction91: What is the value add of having this as a plugin? Are the aspects to the Qiskit implementation that are not covered by our QDK implementation? If not, we can just drop it for simplicity.

wavefunction91 (changes requested):
This also needs formal documentation.

python/src/qdk_chemistry/algorithms/hadamard_test/hadamard_test.py
BoqinZhang: I think this is for using a Hadamard test instance to run different systems. `circuit_builder`, `circuit_executor` and `test_basis` are settings for the instance that may not be changed oftenly; `state_preparation_circuit`, `unitary` and `shots` are the input of a particular test. So the first three are put in the settings with default; and the second three are the input of `run` method.

python/src/qdk_chemistry/plugins/qiskit/hadamard_test_circuit_builder.py
BoqinZhang: In the aspect of functionality, it is totally the same as the Q# circuit builder. Removing it should not lose anything on our side: I think it was just a comparison reference.

python/src/qdk_chemistry/algorithms/hadamard_test/hadamard_test.py
BoqinZhang: I have added all settings in the constructor

docs/source/user/comprehensive/algorithms/hadamard_test.rst
```diff
@@ -0,0 +1,172 @@
...
+4. Measure the control qubit in a selected basis
+
+QDK/Chemistry supports control-qubit measurement in ``X``, ``Y``, and ``Z`` bases through the ``test_basis`` setting.
+For shot-based execution, the expectation estimate from measured counts is:
+
+.. math::
+
+   \langle M \rangle \approx \frac{N_0 - N_1}{N_0 + N_1}
+
+where :math:`N_0` and :math:`N_1` are the observed counts for outcomes ``"0"`` and ``"1"``.
```
wavefunction91: We should explicitly call out `\langle \psi \vert U \vert \psi \rangle` upfront and be careful not to mix up notation in the description.

docs/source/user/comprehensive/algorithms/hadamard_test.rst
```diff
@@ -0,0 +1,172 @@
...
+--------------------
+
+The algorithm returns :class:`~qdk_chemistry.data.CircuitExecutorData`, whose ``bitstring_counts`` can be converted to an expectation estimate:
+
+.. math::
+
+   \hat{m} = \frac{N_0 - N_1}{N_0 + N_1}
+
+Here :math:`N_0 + N_1` is the number of clean measurement outcomes recorded in ``bitstring_counts``.
+This can differ from requested ``shots`` (or ``total_shots``) when an executor tracks loss separately.
```
wavefunction91: Why do we have both of these?

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/hadamard_test/circuit_builder/base.py
feisuxiaozhu: Maybe use an angle that can be set freely, and set to 0 for X, pi/2 for Y? So that we get all measurement angle covered. 

But just for Hadamad test, maybe X and Y are sufficient? We don't need Z, because we will always get 0 outcome on average.

BoqinZhang: Z basis measurement is removed

python/src/qdk_chemistry/utils/qsharp/HadamardTest.qs
```diff
@@ -0,0 +1,42 @@
...
+        repControlledEvolution(control_q, system_q);
+
+        ResetAll(system_q);
+        if (testBasis == PauliX) {
+            return [MResetX(control_q)];
+        } elif (testBasis == PauliY) {
+            return [MResetY(control_q)];
+        } else {
+            fail $"Invalid measurement basis: {testBasis}. Supported values are PauliX and PauliY.";
+        }
```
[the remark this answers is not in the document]
BoqinZhang: This comment is wrong

wavefunction91 (dismissed):

wavefunction91:
@RushiGong can you re-review?

RushiGong (dismissed):

wavefunction91 (approved):

PR #431, opened 2026-03-30
Adding time-dependent Hamiltonian evolution

agamshayit-ms:
This PR adds the ability to run time evolution experiments by mapping a time-dependent Hamiltonian (represented by a list of Hamiltonians and corresponding time coordinates) into a Trotterized unitary. The time evolution unitaryis then converted to a circuit and observables are measured using the native EnergyEstimator.

This includes the optimization of the Trotter decomposition into commuting layers. Commuting layers are then decomposed into parallelizable sublayers to minimize circuit depth.

Added a notebook with periodically-kicked Ising model example


python/src/qdk_chemistry/algorithms/circuit_executor/base.py
```diff
@@ -36,13 +36,23 @@ def _run_impl(
         circuit: Circuit,
         shots: int,
         noise: QuantumErrorProfile | None = None,
+        **kwargs,
```
wavefunction91: We can't have `kwargs` in Algorithm classes - Algorithm classes need to have a single static run API. These extra arguments should be settings

python/src/qdk_chemistry/algorithms/energy_estimator/energy_estimator.py
```diff
@@ -54,6 +54,9 @@ def _run_impl(
         qubit_hamiltonian: QubitHamiltonian,
         total_shots: int,
         noise_model: QuantumErrorProfile | None = None,
+        device_backend_name: str | None = None,
+        pre_transpilation_passes: list[str] | None = None,
+        post_transpilation_passes: list[str] | None = None,
```
wavefunction91: These should be included in the configuration of the circut executtor, not the top level API

python/src/qdk_chemistry/algorithms/time_evolution/circuit_mapper/base.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @abstractmethod
+    def _run_impl(self, evolution: UnitaryRepresentation, *args, **kwargs) -> Circuit:
+        """Construct a Circuit representing the unitary for the given UnitaryRepresentation.
+
+        Args:
+            evolution: The time evolution unitary.
+            *args: Positional arguments, where the first argument is expected to be the
+                time evolution unitary.
+            **kwargs: Additional keyword arguments for concrete implementation.
```
wavefunction91: no args/kwargs: single static APIs

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,189 @@
...
+        qubit_hamiltonians: list[QubitHamiltonian],
+        times: list[float],
+        observables: list[QubitHamiltonian],
+        *,
+        state_prep: Circuit | None = None,
+        shots: int = 1000,
+        noise: QuantumErrorProfile | None = None,
+        device_backend_name: str | None = None,
+        pre_transpilation_passes: list[str] | None = None,
+        post_transpilation_passes: list[str] | None = None,
```
wavefunction91: Per other comments

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @abstractmethod
+    def _run_impl(
+        self,
+        qubit_hamiltonians: list[QubitHamiltonian],
+        times: list[float],
+        observables: list[QubitHamiltonian],
+        *,
+        state_prep: Circuit | None = None,
+        shots: int = 1000,
```
wavefunction91, line 97: This feels like a setting

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,189 @@
...
+        return "measure_simulation"
+
+    @abstractmethod
+    def _run_impl(
+        self,
+        qubit_hamiltonians: list[QubitHamiltonian],
+        times: list[float],
+        observables: list[QubitHamiltonian],
+        *,
+        state_prep: Circuit | None = None,
```
wavefunction91: This should never be optional.

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,189 @@
...
+    def type_name(self) -> str:
+        """Return the algorithm type name as measure_simulation."""
+        return "measure_simulation"
+
+    @abstractmethod
+    def _run_impl(
+        self,
+        qubit_hamiltonians: list[QubitHamiltonian],
+        times: list[float],
+        observables: list[QubitHamiltonian],
```
wavefunction91: What is the reason these should be `list`? is this to deal with the commuting groups, etc that are already in `QubitHamiltonian`?

docs/source/_static/examples/python/Untitled.ipynb
```diff
@@ -0,0 +1,124 @@
+{
```
wavefunction91: Why is this here?

python/src/qdk_chemistry/algorithms/registry.pyi
```diff
@@ -1 +1,513 @@
-# This file is a placeholder and will be replaced with generated stubs on first import
+"""Type stubs for registry.create() with all algorithm overloads."""
```
wavefunction91: This should not be getting checked in.

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
agamshayit-ms: - Observable list: For measuring multiple observables (same as in the energy_estimator class)
- Hamiltonians and times: Allows users to simulate a piecewise constant time dependent Hamiltonian. Addressed by organizing into a new `TimeDependentQubitHamiltonian` class

python/src/qdk_chemistry/algorithms/energy_estimator/energy_estimator.py
```diff
@@ -54,23 +54,19 @@ def _run_impl(
...
-        pre_transpilation_passes: list[str] | None = None,
-        post_transpilation_passes: list[str] | None = None,
     ) -> tuple[EnergyExpectationResult, MeasurementData]:
         """Estimate the expectation value and variance of the Hamiltonian.
 
         The circuit executor used to run quantum circuits is configured via
-        the ``circuit_executor`` setting (an ``AlgorithmRef``).
+        the ``circuit_executor`` setting (an ``AlgorithmRef``).  Backend-specific
+        options (device backend, transpilation passes) are configured on the
+        circuit executor's own settings.
```
wavefunction91: This does not belong in this documentation

python/src/qdk_chemistry/data/time_dependent_qubit_hamiltonian.py
```diff
@@ -0,0 +1,85 @@
...
+import itertools
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from qdk_chemistry.data.qubit_hamiltonian import QubitHamiltonian
+
+__all__ = ["TimeDependentQubitHamiltonian"]
+
+
+class TimeDependentQubitHamiltonian:
```
wavefunction91: I'm not opposed to having a time dependent abstraction for Hamiltonians, but I'm not sure that I like the current design - if we do this, I'd want it to be able to generalize to any type of time-dependent potential.

python/tests/test_evolve_and_measure.py
```diff
@@ -118,17 +132,94 @@ def test_evolve_and_measure_with_device_backend() -> None:
...
         # With device noise the expectation value should still be close to 1.0
         assert measurement[0].energy_expectation_value == pytest.approx(1.0, abs=0.5)
+
+
+# ---------------------------------------------------------------------------
+# Error-path tests for _run_impl input validation
+# ---------------------------------------------------------------------------
+
+
+class TestTimeDependentQubitHamiltonianValidation:
```
wavefunction91: If we do end up keeping the TimeDependentQubitHamiltonian, this should be in a different file.

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/trotter.py
```diff
@@ -156,24 +157,34 @@ def __init__(
...
-    def _run_impl(self, qubit_hamiltonian: QubitHamiltonian) -> UnitaryRepresentation:
+    def _run_impl(
+        self,
+        qubit_hamiltonian: QubitHamiltonian | TimeDependentQubitHamiltonian,
+    ) -> UnitaryRepresentation:
         """Construct the unitary representation using Trotter decomposition.
 
+        Delegates to :meth:`_trotter`, which handles both time-independent
+        and time-dependent Hamiltonians.
+
```
wavefunction91: Not the place for this comment.

python/src/qdk_chemistry/data/time_dependent_qubit_hamiltonian/containers/base.py
```diff
@@ -0,0 +1,43 @@
...
+if TYPE_CHECKING:
+    from qdk_chemistry.data.qubit_hamiltonian import QubitHamiltonian
+
+__all__: list[str] = ["TimeDependentQubitHamiltonianContainer"]
+
+
+class TimeDependentQubitHamiltonianContainer:
+    """Abstract base for time-dependent qubit Hamiltonian storage.
+
+    Concrete containers define *how* a time-dependent Hamiltonian is stored
```
wavefunction91: Strange emphasis

docs/source/_static/examples/python/release_notes_v1_1.py
```diff
@@ -108,7 +108,7 @@
```
wavefunction91, line 108: The 1.1 notes should be a snapshot in time and not updated. This is the [behaviour](https://github.com/microsoft/qdk-chemistry/blob/eb572e85b9b9a7cb92d5a881d0487ccb41b7d9b9/python/tests/test_docs_examples.py#L32-L37) of `main` at the moment

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "\n",
+    "params = {\"pauliExponents\": [], \"pauliCoefficients\": [], \"repetitions\": 1}\n",
+    "targets = list(range(td_hamiltonian.num_qubits))\n",
+    "state_prep = Circuit(\n",
+    "    qsharp_op=QSHARP_UTILS.PauliExp.MakeRepPauliExpOp(params),\n",
+    "    qsharp_factory=QsharpFactoryData(\n",
+    "        program=QSHARP_UTILS.PauliExp.MakeRepPauliExpCircuit,\n",
+    "        parameter={\"evo_params\": params, \"target_indices\": targets},\n",
+    "    ),\n",
+    ")\n"
```
wavefunction91: We should create a utility to handle this - put it in the state prep main folder.

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,355 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d1affab1",
+   "metadata": {},
+   "source": [
+    "## Exact evolution via matrix exponential\n",
```
wavefunction91: This is not exact for a driving field.

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,355 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3ef0a94f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "num_divisions = max(1, round(dt / 0.05))  # Trotter subdivisions within each dt step\n",
```
wavefunction91: what is 0.05 here?

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,355 @@
...
+   "execution_count": null,
+   "id": "3a876d78",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "measure_simulation.settings().set(\n",
+    "    \"circuit_executor\",\n",
+    "    AlgorithmRef(\"circuit_executor\", \"qdk_full_state_simulator\"),\n",
+    ")\n",
+    "measure_simulation.settings().set(\"shots\", 10000)\n",
```
wavefunction91: let's put 10k into a named variable to reflect intention

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "zz_noisy_qdk = measurements_noisy_qdk[0][0].energy_expectation_value\n",
+    "print(f\"Noisy QDK ⟨ZZ⟩: {zz_noisy_qdk:.6f}\")\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e0518676",
+   "metadata": {},
+   "source": [
+    "## Noisy simulation — Qiskit Aer simulator\n",
```
wavefunction91: Let's just nix the Aer simulation, it doesn't add anything to the example.

python/src/qdk_chemistry/algorithms/time_evolution/circuit_mapper/pauli_sequence_mapper.py
```diff
@@ -0,0 +1,118 @@
...
+
+class PauliSequenceMapperSettings(Settings):
+    """Settings for PauliSequenceMapper."""
+
+    def __init__(self):
+        """Initialize PauliSequenceMapperSettings with default values."""
+        super().__init__()
+
+
+class PauliSequenceMapper(EvolutionCircuitMapper):
```
wavefunction91: This name clashes with the controlled variant.

python/src/qdk_chemistry/algorithms/time_evolution/circuit_mapper/base.py
```diff
@@ -0,0 +1,46 @@
+"""QDK/Chemistry time evolution unitary circuit mapper abstractions."""
```
wavefunction91: Since this is a general piece of machinery - it should live alongside the controlled circuit mapper and remove reference to time evolution. This is simply the uncontrolled variant of controlled circuit mapper.

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+            "Circuit mapper used to convert the unitary to a circuit.",
+        )
+        self._set_default(
+            "circuit_executor",
+            "algorithm_ref",
+            AlgorithmRef("circuit_executor", "qdk_sparse_state_simulator"),
+            "Circuit executor used to run quantum circuits.",
+        )
+        self._set_default(
+            "energy_estimator",
```
wavefunction91: I would call this something more generic, like `observable_estimator`.

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+    @abstractmethod
+    def _run_impl(
+        self,
+        hamiltonian: TimeDependentQubitHamiltonian,
+        observables: list[QubitHamiltonian],
+        state_prep: Circuit,
+        *,
+        noise: QuantumErrorProfile | None = None,
+    ) -> list[tuple[EnergyExpectationResult, MeasurementData]]:
+        """Run evolve-and-measure simulation.
```
wavefunction91: We should indicate that this is effectively our concept for ODE integration.

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+            noise: Optional noise profile.
+
+        Returns:
+            A list of tuples containing ``EnergyExpectationResult`` and ``MeasurementData`` objects.
+
+        """
+
+    def _create_time_evolution(
+        self,
+        qubit_hamiltonian: QubitHamiltonian | TimeDependentQubitHamiltonian,
```
wavefunction91: This should only be `QubitHamiltonian` as the static insrtance at time (t)

python/src/qdk_chemistry/algorithms/time_evolution/circuit_mapper/pauli_sequence_mapper.py
```diff
@@ -0,0 +1,118 @@
...
+    r"""Evolution circuit mapper using Pauli product formula term sequences.
+
+    Given a time-evolution operator expressed as a Pauli product formula
+    :math:`U(t) \approx \left[ U_{\mathrm{step}}(t / r) \right]^{r}`, this mapper constructs
```
wavefunction91: Remove references to time

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+from qdk_chemistry.data.circuit import QsharpFactoryData
+from qdk_chemistry.utils.qsharp import QSHARP_UTILS
+
+__all__: list[str] = ["MeasureSimulation", "MeasureSimulationFactory", "MeasureSimulationSettings"]
+
+
+class MeasureSimulationSettings(Settings):
+    """Settings for evolve-and-measure simulation."""
+
+    def __init__(self):
```
wavefunction91, line 32: This should also take a total_time

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def _create_time_evolution(
+        self,
+        qubit_hamiltonian: QubitHamiltonian | TimeDependentQubitHamiltonian,
+        time: float,
+    ) -> UnitaryRepresentation:
+        """Create the time-evolution unitary for current settings."""
+        evolution_builder = self._create_nested("evolution_builder")
+        evolution_builder.settings().set("time", time)
+        return evolution_builder.run(qubit_hamiltonian)
```
wavefunction91, line 142: This is fine, but we need to guard on the evolution builder being an instance of `TimeEvolutionBuilder`

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        self._set_default(
+            "shots",
+            "int",
+            1000,
+            "Number of shots per observable measurement.",
+        )
+
+
+class MeasureSimulation(Algorithm):
```
wavefunction91: This name doesn't really tell me much about its intention. While we have the time, let's work out a better name

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+            "Circuit executor used to run quantum circuits.",
+        )
+        self._set_default(
+            "energy_estimator",
+            "algorithm_ref",
+            AlgorithmRef("energy_estimator", "qdk"),
+            "Energy estimator used to compute expectation values.",
+        )
+        self._set_default(
+            "shots",
```
wavefunction91: Since this is an input to `EnergyEstimator`, let's make it an input to this module

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        self._set_default(
+            "dt",
+            "float",
+            0.0,
+            "Time step for time-dependent evolution. Each step is passed to the builder",
+        )
+
+
+class EvolveAndMeasure(MeasureSimulation):
```
wavefunction91: Let's rename this to reflect it's intention. I'm not saying call it ForwardExponentialEuler, but something along those lines (with documentation to match)

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/euler_integrator.py
```diff
@@ -0,0 +1,182 @@
...
+
+class EvolveAndMeasure(MeasureSimulation):
+    """Evolve under a Hamiltonian and measure a target observable."""
+
+    def __init__(self):
+        """Initialize EvolveAndMeasure with the given settings."""
+        Logger.trace_entering()
+        super().__init__()
+        self._settings = EvolveAndMeasureSettings()
+        self._evolution_circuit: Circuit | None = None
```
wavefunction91, line 78: Why is this a member variable?

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/base.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def type_name(self) -> str:
+        """Return the algorithm type name as measure_simulation."""
+        return "measure_simulation"
+
+    @abstractmethod
+    def _run_impl(
+        self,
+        hamiltonian: TimeDependentQubitHamiltonian,
+        observables: list[QubitHamiltonian],
```
wavefunction91, line 95: Some context should be given to what we expect from users here relative to the outputs. Is it expected that each individual operator will return its own expectation value? Is the consumer allowed to group these among themselves?

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+            "float",
+            1.0,
+            "Total evolution time.",
+        )
+        self._set_default(
+            "dt",
+            "float",
+            0.0,
+            "Time step for time-dependent evolution. Each step is passed to the builder",
+        )
```
wavefunction91: These should both be in the top level settings - let's try to keep the scope of this class to be time evolution, otherwise I'm not sure what it offers over `EnergyEstimator`

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+            Combined ``UnitaryRepresentation`` for the full evolution.
+
+        """
+        dt: float = self._settings.get("dt")
+        if dt <= 0.0:
+            dt = total_time
+        if dt > total_time:
+            raise ValueError(f"dt ({dt}) must not exceed total_time ({total_time}).")
+        num_steps = max(1, round(total_time / dt))
+        dt = total_time / num_steps
```
wavefunction91: This is odd - if a user specifies `dt` they shouldn't be overridden.

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+
+        combined_container: PauliProductFormulaContainer | None = None
+        for i in range(num_steps):
+            t_mid = (i + 0.5) * dt
```
wavefunction91: It's not obvious to me that this is general

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+            h_snapshot = hamiltonian.evaluate(t_mid)
```
wavefunction91: I'm thinking this through for the example you gave - in principle it should work (and give the appropriate result) for arbitrary time steps - i.e. for the piece-wise continuous driving potential, you shouldn't have to "happen" to pick the right time step to get the right result. This can be remedied, but it will require some infra:

To first order, the propagator is given by

$$
H(t) = H_0 + f(t) H_1, \delta t = t_2-t_1
$$

$$
U(t_1,t_2) \approx \exp\left(\int_{t_1}^{t_2}\mathrm{d}t' H(t')\right) = \exp\left(\left(H_0 + \tilde f(t_1,t_2) H_1\right)\delta t\right)
$$

$$
\tilde f (t_1,t_2) = \frac{1}{\delta t} \int_{t_1}^{t_2}\mathrm{d}t' f(t')
$$

python/src/qdk_chemistry/algorithms/registry.pyi
```diff
@@ -1 +1,516 @@
-# This file is a placeholder and will be replaced with generated stubs on first import
```
wavefunction91, line 1: This should not be checked in

python/src/qdk_chemistry/data/time_dependent_qubit_hamiltonian/containers/driven.py
```diff
@@ -0,0 +1,83 @@
...
+        h0: QubitHamiltonian,
+        h1: QubitHamiltonian,
+        drive: Callable[[float], float],
+    ) -> None:
+        """Initialize the driven container."""
+        if h0.num_qubits != h1.num_qubits:
+            raise ValueError("h0 and h1 must have the same number of qubits.")
+
+        self._h0 = h0
+        self._h1 = h1
```
wavefunction91: Let's name these something more descriptive

python/src/qdk_chemistry/data/time_dependent_qubit_hamiltonian/driven.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Raises:
+        ValueError: If *h0* and *h1* have different qubit counts.
+
+    """
+
+    def __init__(
+        self,
+        h0: QubitHamiltonian,
+        h1: QubitHamiltonian,
```
wavefunction91, line 43: Same comment as the other instance

python/src/qdk_chemistry/data/unitary_representation/containers/pauli_product_formula.py
```diff
@@ -125,6 +125,51 @@ def reorder_terms(self, permutation: list[int]) -> "PauliProductFormulaContainer
...
+            (other_container.step_terms, other_container.step_reps),
+        ):
+            for _ in range(step_reps):
+                for term in step_terms:
+                    if merged and merged[-1].pauli_term == term.pauli_term:
+                        new_angle = merged[-1].angle + term.angle
+                        if abs(new_angle) > atol:
+                            merged[-1] = ExponentiatedPauliTerm(pauli_term=term.pauli_term, angle=new_angle)
+                        else:
+                            merged.pop()
```
wavefunction91, line 168: question for @swernli - if I have two identical `c exp` instructions next to each other, will the QDK compiler merge them?

python/src/qdk_chemistry/data/base.py
```diff
@@ -151,7 +151,7 @@ def __getattr__(self, name: str) -> Any:
             if attr is not None:
                 return attr
         # Forward to parent class for other attributes
-        return super().__getattr__(name)
+        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
```
wavefunction91: Why is this behaviour changed?

python/src/qdk_chemistry/data/qubit_hamiltonian.py
```diff
@@ -32,6 +32,23 @@
 __all__: list[str] = []
 
 
+def _merge_term_partitions(p0: TermPartition, p1: TermPartition, offset: int) -> TermPartition | None:
+    """Merge two partitions by concatenating groups and offsetting *p1* indices.
```
wavefunction91, line 36: I would give an example of what to expect here.

python/src/qdk_chemistry/data/qubit_hamiltonian.py
```diff
@@ -200,6 +221,82 @@ def is_hermitian(self, tolerance: float = 1e-12) -> bool:
...
+
+        return QubitHamiltonian(
+            pauli_strings,
+            coefficients,
+            encoding=self.encoding,
+            fermion_mode_order=self.fermion_mode_order,
+            term_partition=partition,
+        )
+
+    def __mul__(self, scalar: complex) -> QubitHamiltonian:
```
wavefunction91: does this work for real as well?

python/src/qdk_chemistry/utils/qsharp/CircuitComposition.qs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Applies two operations sequentially on the same system register.
+    operation ApplySequential(
+        first : Qubit[] => Unit,
+        second : Qubit[] => Unit,
+        systems : Qubit[]
+    ) : Unit {
+        first(systems);
+        second(systems);
+    }
```
wavefunction91, line 17: @swernli Is there a way to make this more general?

wavefunction91 (changes requested):

python/src/qdk_chemistry/data/unitary_representation/containers/pauli_product_formula.py
swernli: I'm not exactly sure what you mean by "merge" in this case, but I'm guessing the answer is no. As of now, the compiler in the QDK doesn't perform any gate optimizations, so it will just see those two calls and not track anything that indicates they are repeats of the same code. What is it you are hoping it might do (or not do)?

python/src/qdk_chemistry/utils/qsharp/CircuitComposition.qs
swernli: There are some fancy Q#-isms you could use, but this looks like the right way to me. Any other techniques may make the control flow and indirection complex enough that it will look odd in circuits or possibly even interfere with QIR generation.

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/euler_integrator.py
agamshayit-ms: So that users can retrieve the simulated circuit via get_circuit(). Equivalent to _iteration_circuits member variable that IQPE uses

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
```diff
@@ -0,0 +1,182 @@
...
+            step_evolution = self._create_time_evolution(h_snapshot, dt)
+            step_container = step_evolution.get_container()
+            if not isinstance(step_container, PauliProductFormulaContainer):
+                raise TypeError(f"Expected PauliProductFormulaContainer, got {type(step_container).__name__}.")
+            if combined_container is None:
+                combined_container = step_container
+            else:
+                combined_container = combined_container.combine(step_container)
```
wavefunction91: Let's make combine a global API requirement on the UnitaryRepresentationContainer

python/src/qdk_chemistry/algorithms/time_evolution/measure_simulation/evolve_and_measure.py
wavefunction91: Per offline discussion, this should be a Propagator algorithm with IO semantics

```
(TimeDependentHamiltonian, t_1, t_2) -> UnitaryRepresentation
```

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/circuit_mapper/non_controlled_pauli_sequence_mapper.py
```diff
@@ -28,12 +28,12 @@ def __init__(self):
         super().__init__()
 
 
-class PauliSequenceMapper(EvolutionCircuitMapper):
-    r"""Evolution circuit mapper using Pauli product formula term sequences.
+class NonControlledPauliSequenceMapper(CircuitMapper):
```
wavefunction91: I would likely opt to change the other one to `ControlledPauliSequenceMapper` than change this one to this oddly specific name.

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
```diff
@@ -0,0 +1,344 @@
...
+      \mathrm{ad}_{\Omega_{j_k}}(H(t))
+
+where :math:`B_k` are Bernoulli numbers.
+
+For the driven case :math:`H(t) = H_0 + f(t)\,H_1` every Magnus term
+reduces to a linear combination of nested commutators of :math:`H_0`
+and :math:`H_1` with scalar coefficients that are iterated integrals
+over the drive function.  The propagator returns
+:math:`\Omega / \delta t` so that the builder (which multiplies by
+``dt``) recovers the full exponent.
```
wavefunction91: This doesn't belong here.

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Returns :math:`H_\text{eff}`.
+
+        """
+        dt = t_end - t_start
+        h0, h1, drive = container.base_hamiltonian, container.drive_hamiltonian, container.drive
+
+        return (1.0 / dt) * integrate.quad(lambda _t: 1.0, t_start, t_end)[0] * h0 + (1.0 / dt) * integrate.quad(
+            drive, t_start, t_end
+        )[0] * h1
```
wavefunction91: $$
\frac{1}{t_2-t_1} \int_{t_1}^{t_2} \mathrm{d}t' = 1
$$

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/euler_integrator.py
```diff
@@ -1,4 +1,20 @@
-"""Hamiltonian evolution + observable measurement implementation."""
+r"""Euler integrator for time-dependent Hamiltonian simulation.
+
+This module implements an Euler integrator for solving the time-dependent
+Schrodinger equation i*dU/dt = H(t)*U on a quantum computer.
+
+The integrator divides [0, T] into uniform steps of size dt.  At each step
```
wavefunction91: I have a slight issue with uniform here - it will only be uniform if $\dfrac{T}{\delta t} \in\mathbb{N}$m otherwise there will be a cleanup time step. Make sure that this logic is consistent throughout the code.

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/euler_integrator.py
wavefunction91: Ah, OK, we need to check this once https://github.com/microsoft/qdk-chemistry/pull/471 is merged and be consistent with the parallel algorithms for execution + circuit generation.

examples/time_evolve_and_measure.ipynb
```diff
@@ -0,0 +1,259 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "64f39b5e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qdk_chemistry.algorithms.state_preparation import identity_state_prep\n",
+    "\n",
+    "state_prep = identity_state_prep(num_qubits=td_hamiltonian.num_qubits)"
```
wavefunction91, line 85: Thanks for changing this - but this should still be documented to indicate what this means in this context. All spin up?

wavefunction91 (changes requested):

examples/time_evolve_and_measure.ipynb
agamshayit-ms: Added comment in the notebook

python/src/qdk_chemistry/algorithms/time_evolution/hamiltonian_simulation/euler_integrator.py
agamshayit-ms: Fixed documentation and added residual step handling

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
agamshayit-ms: Fixed

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
agamshayit-ms: Fixed

python/src/qdk_chemistry/algorithms/circuit_mapper/non_controlled_pauli_sequence_mapper.py
agamshayit-ms: Renamed

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
```diff
@@ -1,32 +1,14 @@
-r"""Time-averaged propagator with Magnus expansion.
+r"""Time-averaged propagator via first-order Magnus expansion.
```
wavefunction91: The top level docs should not restrict to first order - this code will eventually be extended to higher order expansions, just not today.

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
```diff
@@ -61,18 +43,13 @@ def __init__(self):
 
 
 class MagnusPropagator(Propagator):
-    r"""Magnus propagator with recursive expansion.
+    r"""First-order Magnus propagator (time averaging).
```
wavefunction91: Same thing here - don't restrict documentation to things that will be generalized in the future.

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
agamshayit-ms: restored and fixed Magnus docs

python/src/qdk_chemistry/algorithms/propagator/magnus_propagator.py
agamshayit-ms: restored and fixed Magnus docs

wavefunction91 (approved):

PR #435, opened 2026-04-06
Implement sparse isometry with binary encoding

RushiGong:


python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1206 @@
...
+        assert sparse_size > 0
+        if sparse_size & (sparse_size - 1) == 0:
+            return sparse_size
+        return 1 << (sparse_size.bit_length() - 1)
+
+    def _record(self, op: tuple[MatrixCompressionType, tuple[Any, ...]]):
+        """Append an operation and update the tableau.
+
+        Args:
+            op: Operation to record, as a tuple of (MatrixCompressionType, payload).
```
YingrongChen: Is payload the same as input qubits/rows in the table?

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1206 @@
...
+           bit, with no ancilla waste.
+
+        Args:
+            rank: Number of pivot columns (size of the identity block).
+
+        """
+        if rank == 0:
+            return
+
+        logical_rows = self._apply_unary_staircase(rank)
```
YingrongChen: As we discussed, the `_apply_unary_staircase` can be combined with `_ref_to_staircase`

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1206 @@
...
+    """Raised when a matrix is not in row echelon form (REF)."""
+
+
+class MatrixCompressionType(Enum):
+    """Operations types recorded during matrix compression."""
+
+    CX = "cx"
+    SWAP = "swap"
+    TOFFOLI = "toffoli"
+    PUI_BLOCK = "pui_block"
```
YingrongChen: We can rename `PUI_BLOCK` to `SELECT` for uniformity

YingrongChen: Should we merge this `MatrixCompressionType` with `MatrixCompressionOp`?

And the logic for updating matrix based on a quantum operation could be reused both here and in the regular sparse isometry, where you update the matrix based on cx. Maybe we want to remove redundant code.

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1206 @@
...
+                    num_fixed + off,
+                    bool((be.dense_content >> (dense_size - 1 - num_fixed - off)) & 1),
+                )
+                for off in range(num_changing)
+            ]
+            rest_entries.append((i, changing_controls))
+
+        self._record((MatrixCompressionType.PUI_BLOCK, (fixed_controls, rest_entries)))
+
+    def to_gf2x_operations(
```
YingrongChen: If we could merge `MatrixCompressionOp` with `MatrixCompressionOp`, then we might not need this conversion function.

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry_binary_encoding.py
```diff
@@ -0,0 +1,297 @@
...
+            wavefunction: The target wavefunction to prepare.
+
+        Returns:
+            A Circuit object containing the quantum circuit.
+
+        """
+        Logger.trace_entering()
+
+        # Active Space Consistency Check (same as parent)
+        alpha_indices, beta_indices = wavefunction.get_orbitals().get_active_space_indices()
```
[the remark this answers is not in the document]
RushiGong: The check has been removed.

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -467,6 +512,17 @@ def gf2x_with_tracking(matrix: np.ndarray) -> GF2XEliminationResult:
...
         matrix: shape (m, n), binary (0/1) matrix
+        skip_diagonal_reduction: If True, skip the optional diagonal-to-upper-
+            staircase rank reduction (step 4).  Binary encoding handles the
+            identity pivot block natively, so the extra CX + X expansion ops
+            produced by the diagonal reduction are redundant Cliffords.
+        staircase_mode: If True, perform forward-only GF2 elimination (REF)
+            then convert the pivot block directly to upper-staircase form
+            with minimal CX ops — skipping back-substitution entirely.
+            This produces a matrix that binary encoding's diagonal path
+            recognises as ``is_diagonal_reduced``, saving ``r-1`` CX gates
```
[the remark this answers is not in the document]
RushiGong: Docstring has been updated.

python/src/qdk_chemistry/utils/binary_encoding.py
RushiGong: Yes, renamed it as `qubit_args` for clarity.

python/src/qdk_chemistry/utils/binary_encoding.py
RushiGong: `PUI_BLOCK` has been updated to `SELECT`.
The `MatrixCompressionType` has been made to an Enum and `MatrixCompressionOp` contains more information including `MatrixCompressionType`, qubit indices, control state, and lookup data.

python/src/qdk_chemistry/utils/binary_encoding.py
RushiGong: This function has been refactored and can be used directly in `SparseIsometryBinaryEncoding` for passing operations into Q# layer.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,300 @@
...
+        /// Total number of qubits.
+        numQubits : Int,
+    }
+
+    /// Apply a single matrix-compression gate to a qubit register.
+    operation ApplyMatrixCompressionOp(gate : MatrixCompressionOp, qs : Qubit[]) : Unit {
+        if gate.name == "X" {
+            X(qs[gate.qubits[0]]);
+        } elif gate.name == "CX" {
+            CX(qs[gate.qubits[0]], qs[gate.qubits[1]]);
```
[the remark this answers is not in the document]
RushiGong: Q# uses CX.

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -574,56 +578,34 @@ def _remove_duplicate_rows_with_cnot(
...
-                operations_work.append(("cnot", (row_map_work[j], row_map_work[i])))
-                rows_to_eliminate.append(j)
+            if np.array_equal(matrix_work[i], matrix_work[j]):
+                operations_work.append(("cx", (row_map_work[j], row_map_work[i])))
+                matrix_work[j] ^= matrix_work[i]
+                rows_to_eliminate.add(j)
 
                 Logger.info(
-                    f"Found duplicate row {j} identical to row {i}, adding CNOT({row_map_work[i]}, {row_map_work[j]})"
+                    f"Found duplicate row {j} identical to row {i}, adding CX({row_map_work[i]}, {row_map_work[j]})"
```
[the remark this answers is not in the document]
RushiGong: Keep the log message in `CX(control,target)` order.

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1256 @@
...
+            filtered_table,
+            address_qubits=address_qubits,
+            data_qubits=data_qubits,
+            use_measurement_and=self.measurement_based_uncompute,
+        )
+
+        gf2x_ops = list(reversed(lookup_ops))
+        and_count = sum(1 for name, _ in lookup_ops if name == "and")
+
+        return gf2x_ops, and_count
```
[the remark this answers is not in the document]
RushiGong: Resolved.

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1256 @@
...
+                translated.append(
+                    (
+                        MatrixCompressionType.MCX,
+                        (
+                            [map_idx(int(q)) for q in controls],
+                            list(ctrl_state),
+                            map_idx(int(target)),
+                        ),
+                    )
+                )
```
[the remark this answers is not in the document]
RushiGong: Resolved.

examples/state_prep_energy.ipynb
```diff
@@ -415,12 +435,12 @@
             "source": [
                 "# Estimate energy using the optimized circuit and the qubit Hamiltonian\n",
                 "estimator = create(\"energy_estimator\", algorithm_name=\"qdk\")\n",
-                "circuit_executor = create(\"circuit_executor\", algorithm_name=\"qdk_full_state_simulator\")\n",
+                "circuit_executor = create(\"circuit_executor\", algorithm_name=\"qiskit_aer_simulator\")\n",
```
wavefunction91: Why AER here?

python/src/qdk_chemistry/utils/binary_encoding.py
```diff
@@ -0,0 +1,1313 @@
...
+        self.data = np.asarray(data, dtype=np.int8)
+        if self.data.ndim != 2:
+            raise ValueError("Input data must be a 2-dimensional array")
+
+        _check_ref(self.data)
+
+        self.num_rows, self.num_cols = self.data.shape
+        self.dense_size = _dense_qubits_size(self.num_cols)
+        if self.dense_size >= self.num_rows:
+            raise ValueError(
```
YingrongChen: Do we have to raise an error here? Could we just do nothing in the binary encoding and only do dense state preparation?

RushiGong: Move this error raise into `BinaryEncodingSynthesizer`. Also added check in sparse isometry binary encoding, if already dense, just prepare from gf2x results.

python/src/qdk_chemistry/utils/qsharp/qsharp.json
```diff
@@ -0,0 +1,14 @@
+{
```
wavefunction91, line 1: Why does this need to exist?

python/tests/test_data/ozone_sparse_ci_wavefunction.wavefunction.json
```diff
@@ -0,0 +1,4309 @@
+{
```
wavefunction91: If we're going to embed something like this, it should be in HDF5 to save space. ASCII JSON is a bit heavy.

python/tests/test_helpers.py
```diff
@@ -247,3 +247,159 @@ def create_test_ansatz(num_orbitals: int = 2):
     wavefunction = Wavefunction(container)
 
     return Ansatz(hamiltonian, wavefunction)
+
+
+def _hf_determinant(n_alpha: int, n_beta: int, n_orbitals: int) -> np.ndarray:
```
wavefunction91: This lives elsewhere.

wavefunction91 (dismissed):

python/src/qdk_chemistry/algorithms/state_preparation/state_preparation.py
```diff
@@ -76,5 +70,29 @@ def algorithm_type_name(self) -> str:
         return "state_prep"
 
     def default_algorithm_name(self) -> str:
-        """Return the sparse_isometry_gf2x as default algorithm name."""
-        return "sparse_isometry_gf2x"
+        """Return the sparse_isometry as default algorithm name."""
+        return "sparse_isometry"
+
+    def create(self, name: str | None = None) -> Algorithm:
```
mmoerchen: Could we use the existing deprecation machinery here (as we did for `expectation_estimator` and `QdkExpectationEstimator`) instead of introducing a new create function?

.github/workflows/build-and-test.yaml
```diff
@@ -882,8 +882,7 @@ jobs:
       run: |
         git clone --depth 1 --branch "${QDK_BRANCH}" https://github.com/microsoft/qdk.git "${{ runner.temp }}/qdk"
         echo "QDK commit: $(git -C "${{ runner.temp }}/qdk" rev-parse HEAD)"
-        pip install --force-reinstall "${{ runner.temp }}/qdk/source/widgets" "${{ runner.temp }}/qdk/source/jupyterlab"
-        pip install --force-reinstall "${{ runner.temp }}/qdk/source/qdk_package"
+        pip install --force-reinstall --no-deps "${{ runner.temp }}/qdk/source/qdk_package"
```
mmoerchen, line 858: Why do we need to remove the install of widgets and jupyterlab?

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -128,98 +147,343 @@ def _run_impl(self, wavefunction: Wavefunction) -> Circuit:
...
+            RefTableau(gf2x_result.reduced_matrix),
+            include_negative_controls=include_negative_controls,
+            measurement_based_uncompute=self._settings.get("measurement_based_uncompute"),
+        ).synthesize(
+            num_local_qubits=n_qubits,
+            active_qubit_indices=gf2x_result.row_map,
+            ancilla_start=n_qubits,
+        )
+
+        compressed_sv = np.zeros(2**dense_size, dtype=float)
```
agamshayit-ms, line 403: This will silently discard imaginary parts of complex `StateVectorCOntainer`

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -128,98 +147,343 @@ def _run_impl(self, wavefunction: Wavefunction) -> Circuit:
...
-        from qiskit.transpiler import PassManager  # noqa: PLC0415
+        configs = []
+        coeffs_list = []
+        for idx in range(len(statevector_data)):
+            if statevector_data[idx] != 0.0:
+                bits_str = "".join(str((idx >> i) & 1) for i in range(rank))
+                configs.append(Configuration.from_bitstring(bits_str))
+                coeffs_list.append(statevector_data[idx])
+
+        coeffs_arr = np.array(coeffs_list, dtype=float)
```
agamshayit-ms, line 461: This will silently discard imaginary parts of complex `StateVectorCOntainer`

python/src/qdk_chemistry/algorithms/state_preparation/state_preparation.py
```diff
@@ -67,6 +59,8 @@ def run(self, wavefunction: Wavefunction) -> Circuit:
 class StatePreparationFactory(AlgorithmFactory):
     """Factory class for creating StatePreparation instances."""
 
+    _DEPRECATED_ALIASES: ClassVar[dict[str, str]] = {"sparse_isometry_gf2x": "sparse_isometry"}
```
agamshayit-ms: We also need to translate the old settings, not just the algorithm name. Otherwise, `SettingNotFound` is raised. 

For example, dense_preparation_method="qiskit" could map to a dense_state_prep=AlgorithmRef("state_prep", "qiskit_regular_isometry", ...), while the old transpilation keys could be forwarded or accepted with deprecation warnings.

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -128,98 +147,343 @@ def _run_impl(self, wavefunction: Wavefunction) -> Circuit:
...
+            embedding_map: Maps reduced qubit indices to full register positions.
+            n_qubits: Total number of qubits in the full register.
+            dense_algo: The nested dense state prep algorithm instance (for transpile settings).
+
+        Returns:
+            Composed Circuit operating on the full qubit register.
+
+        """
+        factory = dense_circuit._qsharp_factory  # noqa: SLF001
+        if factory is not None and getattr(factory.program, "__name__", None) == _QSHARP_DENSE_PREP_PROGRAM:
```
agamshayit-ms: `StatePreparation` with valid `Circuit` may have another name, but is forced through `_compose_qiskit`, causing a runtime error.

examples/factory_list.ipynb
```diff
@@ -1,2457 +1,2457 @@
 {
- "cells": [
```
agamshayit-ms, line 2: why the 2k+ lines diff? notebooks seem identical, can this be removed?

examples/state_prep_energy.ipynb
```diff
@@ -1,470 +1,470 @@
 {
```
agamshayit-ms, line 1: why the 468 lines diff? notebooks seem identical, can this be removed?

agamshayit-ms (changes requested):

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+            int: Maximum batch size as the largest power-of-2 not exceeding the sparse row count.
+
+        """
+        sparse_size = self.tableau.num_rows - self.tableau.dense_size
+        assert sparse_size > 0
+        if sparse_size & (sparse_size - 1) == 0:
+            return sparse_size
+        return 1 << (sparse_size.bit_length() - 1)
+
+    def _record(self, op: tuple[str, Any]):
```
agamshayit-ms: Could `circuit` store `MatrixCompressionOp` directly? `_record()` and `_to_operations()` duplicate dispatch over a tuple DSL; one typed representation would make export just reverse + remap.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+
+
+class MatrixCompressionType(CaseInsensitiveStrEnum):
+    """Supported operation types for matrix compression."""
+
+    X = "X"
+    CX = "CX"
+    SWAP = "SWAP"
+    CCX = "CCX"
+    MCX = "MCX"
```
agamshayit-ms: is "MCX" actually used anywhere?

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+    SELECT = "SELECT"
+    SELECT_AND = "SELECT_AND"
+
+
+_QSHARP_OPERATION_CODES = {
```
agamshayit-ms: Could the Q# wire code live on `MatrixCompressionType` instead? The parallel `_QSHARP_OPERATION_CODES` map can drift from the enum and Q# dispatch.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+        target_bits = np.zeros(self.tableau.num_rows - dense_size, dtype=bool)
+        target_bits[len(self.batch)] = True
+        for s_qubit in np.flatnonzero(sparse_col_data != target_bits):
+            self._record((MatrixCompressionType.CX, (ctrl_row, dense_size + int(s_qubit))))
+
+        self.batch.append((current_col, self.batch_index))
+        self.bijection.append((self.batch_index, current_col))
+        self.batch_index += 1
+
+    def _clear_sparse_bits(self):
```
agamshayit-ms, line 600: Could this collapse to `_clear_sparse_bits()` plus one lookup-builder helper? The flush → synthesize → lookup chain is otherwise single-use plumbing.

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
```diff
@@ -59,9 +59,170 @@ def convert(self, qir: pyqir.Module) -> QuantumCircuit:
...
         # Build the circuit
-        self.run(qir)
+        self._walk_blocks(entry_point.basic_blocks[0])
         return self._circuit
 
+    # =========================================================================
+    # Control flow
+    # =========================================================================
+
+    def _walk_blocks(self, block: "pyqir.BasicBlock") -> None:
```
agamshayit-ms, line 75: Could this land as a prerequisite QIR PR? Core binary encoding emits no branches; only `measurement_based_uncompute` needs this walker to preserve four classically controlled CZ corrections.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,354 @@
...
+            ApplyControlledOnBitString(ctrlStateBools, X, controlQubits, qs[gate.qubits[numControls]]);
+        } else {
+            fail "Unsupported adjointable matrix-compression operation.";
+        }
+    }
+
+    /// ``ancillaPool`` is a list of pre-initialised |0⟩ qubits that
+    /// SparseOneHotSelect may borrow as helpers (avoids allocating new qubits).
+    /// Pass an empty array when no pool is available (e.g. for GF2+X ops).
+    operation ApplyMatrixCompressionOp(gate : MatrixCompressionOp, qs : Qubit[], ancillaPool : Qubit[]) : Unit {
```
agamshayit-ms: Could SELECT and SELECT_AND share the qubit-extraction path, passing `gate.kind == 6` as the uncompute mode?

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,354 @@
...
+        for row in data {
+            for bit in row {
+                if bit { return false; }
+            }
+        }
+        return true;
+    }
+
+    /// Apply X to each target qubit where the corresponding data bit is true.
+    operation WriteOneHotData(data : Bool[], target : Qubit[]) : Unit {
```
agamshayit-ms: Could these helpers use standard `ApplyPauliFromBitString` and `AND`? They match the current write and measurement-based uncompute behavior.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,354 @@
...
+            }
+        }
+    }
+
+    /// Singly-controlled recursion for SparseOneHotSelect.
+    ///
+    /// Uses ``ancillaPool[0]`` as the helper qubit (must be |0⟩ on entry,
+    /// restored on exit) and passes ``ancillaPool[1...]`` to recursive calls.
+    /// Falls back to ``use helper = Qubit()`` when the pool is empty.
+    operation SparseOneHotSCS(
```
agamshayit-ms, line 134: Could helper allocation and AND-vs-CCNOT selection be factored around one recursion body? This removes the pooled/allocated × coherent/measured duplication.

python/tests/test_utils_binary_encoding.py
```diff
@@ -0,0 +1,780 @@
...
+    MatrixCompressionType,
+    RefTableau,
+    _BinaryEncodingSynthesizer,
+)
+from qdk_chemistry.algorithms.state_preparation.sparse_isometry import gf2x_with_tracking
+
+from .test_helpers import create_random_bitstring_matrix
+
+
+def _dense_register_width(num_cols: int) -> int:
```
agamshayit-ms, line 25: These tests exercise copied local formulas rather than production code. Could they assert `synthesize()` output, tableau replay, bijection, and zero-row invariants instead?

python/tests/test_utils_binary_encoding.py
```diff
@@ -0,0 +1,780 @@
...
+    """Local helper mirroring the inlined RefTableau.dense_size calculation."""
+    return 1 if num_cols < 2 else math.ceil(math.log2(num_cols))
+
+
+def _int_to_bits(val: int, nbits: int) -> list[bool]:
```
agamshayit-ms: These tests exercise copied local formulas rather than production code. Could they assert `synthesize()` output, tableau replay, bijection, and zero-row invariants instead?

python/tests/test_utils_binary_encoding.py
```diff
@@ -0,0 +1,780 @@
...
+    """Local helper mirroring the inlined _int_to_bits."""
+    return [bool((val >> i) & 1) for i in range(nbits - 1, -1, -1)]
+
+
+def _bits_to_int(bits) -> int:
```
agamshayit-ms, line 30: These tests exercise copied local formulas rather than production code. Could they assert `synthesize()` output, tableau replay, bijection, and zero-row invariants instead?

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,354 @@
...
+    /// Controlled variant: apply CX(ctl, target[i]) for each true bit.
+    operation ControlledWriteOneHotData(ctl : Qubit, data : Bool[], target : Qubit[]) : Unit {
+        for i in 0..Length(data) - 1 {
+            if data[i] { CX(ctl, target[i]); }
+        }
+    }
+
+
+    /// AND gate with measurement-based adjoint uncomputation.
+    operation MeasurementBasedAND(a : Qubit, b : Qubit, target : Qubit) : Unit is Adj {
```
agamshayit-ms: Could these helpers use standard ApplyPauliFromBitString and AND? They match the current write and measurement-based uncompute behavior.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Apply X to each target qubit where the corresponding data bit is true.
+    operation WriteOneHotData(data : Bool[], target : Qubit[]) : Unit {
+        for i in 0..Length(data) - 1 {
+            if data[i] { X(target[i]); }
+        }
+    }
+
+    /// Controlled variant: apply CX(ctl, target[i]) for each true bit.
+    operation ControlledWriteOneHotData(ctl : Qubit, data : Bool[], target : Qubit[]) : Unit {
```
agamshayit-ms: Could these helpers use standard ApplyPauliFromBitString and AND? They match the current write and measurement-based uncompute behavior.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+                for off in range(num_changing)
+            ]
+            rest_entries.append((i, changing_controls))
+
+        select_ops: list[tuple[str, Any]] = []
+        self._flush_pui_lookup_block(select_ops, dense_size, fixed_controls, rest_entries)
+        for op in select_ops:
+            self._record(op)
+
+    def _flush_pui_lookup_block(self, ops, sbs, fixed_controls, rest_entries):
```
agamshayit-ms: Could this collapse to `_clear_sparse_bits()` plus one lookup-builder helper? The flush → synthesize → lookup chain is otherwise single-use plumbing.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+
+        chunked_ops, chunked_count = [], 0
+        for chunk in chunks:
+            sub_ops, sub_count = self._synthesize_single_pui_lookup_block(sbs, fixed_controls, chunk)
+            chunked_ops.extend(sub_ops)
+            chunked_count += sub_count
+
+        ops.extend(chunked_ops if chunked_count <= mono_count else mono_ops)
+
+    def _synthesize_single_pui_lookup_block(self, sbs, fixed_controls, rest_entries):
```
agamshayit-ms, line 647: Could this collapse to `_clear_sparse_bits()` plus one lookup-builder helper? The flush → synthesize → lookup chain is otherwise single-use plumbing.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+                + _BinaryEncodingSynthesizer._scs_toffoli_cost(right)
+            )
+        if not right_empty:
+            return split_cost + _BinaryEncodingSynthesizer._scs_toffoli_cost(right)
+        if not left_empty:
+            return split_cost + _BinaryEncodingSynthesizer._scs_toffoli_cost(left)
+        return 0
+
+    @staticmethod
+    def _lookup_select(table_dict, address_qubits, data_qubits, *, use_measurement_and=False):
```
agamshayit-ms: Could this collapse to `_clear_sparse_bits()` plus one lookup-builder helper? The flush → synthesize → lookup chain is otherwise single-use plumbing.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+                f"leaving no spare rows in a {tableau.num_rows}-row matrix)."
+            )
+        self.include_negative_controls = include_negative_controls
+        self.measurement_based_uncompute = measurement_based_uncompute
+        self.batch: list[tuple[int, int]] = []
+        self.batch_index: int = 0
+        self.circuit: list[tuple[str, Any]] = []
+        self.bijection: list[tuple[int, int]] = []
+
+    def synthesize(
```
agamshayit-ms, line 273: is "MCX" actually used anywhere?

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+            num_local_qubits: Number of system (local) qubits.
+            active_qubit_indices: Mapping from local row indices to global qubit indices.
+            ancilla_start: Global qubit index where ancilla qubits begin.
+
+        Returns:
+            list[MatrixCompressionOp]: Circuit operations in reversed order with remapped qubit indices.
+
+        """
+
+        def map_idx(idx: int) -> int:
```
agamshayit-ms: is "MCX" actually used anywhere?

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,840 @@
...
+            elif op_type is MatrixCompressionType.CCX:
+                target, ctrl1, ctrl2 = qubit_args
+                ops.append(
+                    MatrixCompressionOp(op_type, [map_idx(int(ctrl1)), map_idx(int(ctrl2)), map_idx(int(target))])
+                )
+            elif op_type in {MatrixCompressionType.SELECT, MatrixCompressionType.SELECT_AND}:
+                data_table, addr_qubits, dat_qubits = qubit_args
+                qubits = [map_idx(int(q)) for q in addr_qubits] + [map_idx(int(q)) for q in dat_qubits]
+                ops.append(MatrixCompressionOp(op_type, qubits, control_state=len(addr_qubits), lookup_data=data_table))
+            elif op_type is MatrixCompressionType.MCX:
```
agamshayit-ms: is "MCX" actually used anywhere?

python/src/qdk_chemistry/algorithms/state_preparation/state_preparation.py
RushiGong: Added in commit `063f5f0`.

examples/factory_list.ipynb
RushiGong: Resolved in commit `063f5f0`.

examples/state_prep_energy.ipynb
RushiGong: Resolved in commit `063f5f0`.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
```diff
@@ -0,0 +1,371 @@
...
+
+    /// Returns a callable that applies binary-encoding composition.
+    function MakeComposeBinaryEncodingOp(
+        denseParams : StatePreparationParams,
+        embeddingMap : Int[],
+        binaryEncodingOps : MatrixCompressionOp[],
+        gaussianEliminationOps : MatrixCompressionOp[],
+        ancillaPool : Int[],
+    ) : Qubit[] => Unit {
+        ComposeBinaryEncoding(denseParams, embeddingMap, binaryEncodingOps, gaussianEliminationOps, ancillaPool, _)
```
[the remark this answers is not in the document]
RushiGong: This doesn't reproduce, and neither option works. `test_ancilla_overflow_standard_qpe_qubit_allocation` builds Standard QPE from
binary-encoding state prep and calls `.estimate()`, which forces Q# compilation. It
passes.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
RushiGong: Yes, but some current tests require running statevector validation using qiskit circuit.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/plugins/qiskit/conversion.py
```diff
@@ -131,3 +137,70 @@ def _configuration_to_statevector_index(configuration: data.Configuration, n_bit
...
+    operations: X, CX, SWAP, CCX, SELECT, and SELECT_AND.
+
+    Args:
+        circuit: Qiskit QuantumCircuit to append gates to (modified in place).
+        ops: List of MatrixCompressionOp to apply.
+
+    """
+    for op in ops:
+        name = op.name.upper()
+        if name == "X":
```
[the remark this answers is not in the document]
RushiGong: `MatrixCompressionType`is a StrEnum, so op.name is a str, upper() returns 'CX', no AttributeError.

python/src/qdk_chemistry/utils/qsharp/src/BinaryEncoding.qs
RushiGong: Updated in [614c8b2](https://github.com/microsoft/qdk-chemistry/pull/435/commits/614c8b21f94bd80986c8b943580502776e68310d)

python/tests/test_utils_binary_encoding.py
RushiGong: Removed in [614c8b2](https://github.com/microsoft/qdk-chemistry/pull/435/commits/614c8b21f94bd80986c8b943580502776e68310d)

python/tests/test_utils_binary_encoding.py
RushiGong: This serves as helper functions. The tests you mentioned have been already covered.

python/tests/test_utils_binary_encoding.py
RushiGong: This serves as helper functions. The tests you mentioned have been already covered.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Updated in [aacc540](https://github.com/microsoft/qdk-chemistry/pull/435/commits/aacc5402414f926eb6fa3ad62c8e25d216c2e8be)

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
RushiGong: Added check and warning for imaginary part in `063f5f0`

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
RushiGong: Added check and warning for imaginary part in `063f5f0`

.github/workflows/build-and-test.yaml
RushiGong: The original step failed: `widgets` and `jupyterlab` can't be pip-installed from source here. Both use `hatch-jupyter-builder`, whose build hook shells out to `npm build` and expects prebuilt JS assets, this job only sets up Python.

The job's purpose is to test our branch against qdk `main`'s core functionality; nothing in the chemistry test suite touches the widget or JupyterLab.

.github/workflows/build-and-test.yaml
```diff
@@ -882,8 +882,7 @@ jobs:
...
         echo "Installed qdk from branch: ${QDK_BRANCH}"
```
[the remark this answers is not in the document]
RushiGong: The job installed the released version of qdk first (including widget), then reinstalled qdk main's core functionality. qdk.widgets is available.

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
```diff
@@ -33,54 +33,135 @@
...
 
-__all__: list[str] = []
+from ._binary_encoding_utils import MatrixCompressionOp, MatrixCompressionType, RefTableau, _BinaryEncodingSynthesizer
 
+__all__: list[str] = ["SparseIsometryStatePreparationSettings"]
 
-class SparseIsometryGF2XStatePreparationSettings(StatePreparationSettings):
-    """Settings for SparseIsometryGF2XStatePreparation."""
+
+def _is_qsharp_dense_factory(factory: QsharpFactoryData | None) -> bool:
```
agamshayit-ms, line 69: This still fails for a custom `StatePreparation` returning `Circuit(qsharp=...)`; both modes fall through to Qiskit and fail. Minimal fix: validate `dense_state_prep` as one of the two supported implementations and document that restriction, or add an explicit composition capability to the contract.

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
agamshayit-ms: `_lookup_select()` is still single-use from this method; could its table construction be inlined here?

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,802 @@
...
+        )
+        gf2x_ops = list(reversed(lookup_ops))
+        toffoli_cost = sum(
+            self._scs_toffoli_cost(op.lookup_data, root=True)
+            for op in lookup_ops
+            if op.name in (MatrixCompressionType.SELECT, MatrixCompressionType.SELECT_AND)
+        )
+        return gf2x_ops, toffoli_cost
+
+    def _to_operations(
```
agamshayit-ms: MCX is removed, but the unreachable `idx >= num_local_qubits` path remains. Minimal fix: map every index through `active_qubit_indices` and remove `num_local_qubits`/`ancilla_start` from `synthesize()` and `_to_operations()`.

python/tests/test_interop_qiskit_qir.py
```diff
@@ -62,3 +65,88 @@ def test_qir_to_qiskit_conversion():
     assert circuit2.num_qubits == 3
     assert circuit2.num_clbits == 3
     assert circuit2.count_ops() == {"h": 1, "x": 1, "y": 1, "cx": 1, "cz": 1, "t": 1, "s": 1, "measure": 3}
+
+
```
agamshayit-ms, line 69: These branch tests now compile under the Base-profile default and fail with `UseOfDynamicBool`. Could `_qsharp_qir()` use a local `Adaptive_RIF` context? Both tests pass with that profile.

python/tests/test_state_preparation_binary_encoding.py
```diff
@@ -0,0 +1,859 @@
...
+    """Tests for :meth:`SparseIsometryStatePreparation.create_dense`.
+
+    ``create_dense`` returns the dense-loading stage alone, on the same full-width
+    register as ``run``, so the isometry cost can be recovered by subtracting the two
+    resource estimates.
+    """
+
+    ISOMETRY_OP_KEYS = ("expansionOps", "binaryEncodingOps", "gaussianEliminationOps")
+
+    @pytest.mark.parametrize("binary_encoding", [False, True], ids=["no_binenc", "binenc"])
```
agamshayit-ms, line 424: Could we add a custom Q# `StatePreparation` regression test? It currently falls through to Qiskit and fails.

agamshayit-ms (changes requested):

python/src/qdk_chemistry/algorithms/state_preparation/sparse_isometry.py
RushiGong: Added check for composable Circuit object in [51533cb](https://github.com/microsoft/qdk-chemistry/pull/435/commits/51533cbb9d9b8e7a6be00fb8cb762dd4e932e65d)

python/tests/test_interop_qiskit_qir.py
RushiGong: Fixed in [51533cb](https://github.com/microsoft/qdk-chemistry/pull/435/commits/51533cbb9d9b8e7a6be00fb8cb762dd4e932e65d)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [51533cb](https://github.com/microsoft/qdk-chemistry/pull/435/commits/51533cbb9d9b8e7a6be00fb8cb762dd4e932e65d)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Fixed in [51533cb](https://github.com/microsoft/qdk-chemistry/pull/435/commits/51533cbb9d9b8e7a6be00fb8cb762dd4e932e65d)

python/tests/test_state_preparation_binary_encoding.py
RushiGong: Added the test in [51533cb](https://github.com/microsoft/qdk-chemistry/pull/435/commits/51533cbb9d9b8e7a6be00fb8cb762dd4e932e65d)

python/src/qdk_chemistry/algorithms/state_preparation/state_preparation.py
```diff
@@ -6,22 +6,11 @@
 # --------------------------------------------------------------------------------------------
 
 from qdk_chemistry.algorithms.base import Algorithm, AlgorithmFactory
-from qdk_chemistry.data import Circuit, Settings, Wavefunction
+from qdk_chemistry.data import Circuit, Wavefunction
 
 __all__: list[str] = []
 
 
-class StatePreparationSettings(Settings):
```
agamshayit-ms, line 14: Doesn't it cause backward-compatibility problems to remove this class? is there another way to deprecate it if that's the intention?

python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -261,6 +261,37 @@ def _resolve_algorithm_type(algorithm_type: str) -> str:
...
+            f"Algorithm '{algorithm_name}' of type '{algorithm_type}' is deprecated and will be "
+            f"removed in a future release; use '{new_name}' instead.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        return new_name
+    return algorithm_name
+
+
 def create(algorithm_type: str, algorithm_name: str | None = None, **kwargs) -> Algorithm:
```
agamshayit-ms, line 295: This calls update to do the translation from old keys values. But when a user runs something like `prep.settings().set(key, value)`, `update()` is not called. `set()` needs to handle translation as well, otherwise we may run into errors.

python/src/qdk_chemistry/utils/qsharp/src/StatePreparation.qs
```diff
@@ -6,16 +6,19 @@ namespace QDKChemistry.Utils.StatePreparation {
...
+    import QDKChemistry.Utils.BinaryEncoding.ApplyAdjointableCompressionOp;
+
 
     /// A struct to hold parameters for state preparation.
     /// - `rowMap`: An array of integers representing the mapping of qubits to rows in the state vector.
     /// - `stateVector`: An array of doubles representing the amplitudes of the quantum state.
-    /// - `expansionOps`: An array of arrays of integers representing the operations to expand the state preparation (e.g., CNOTs, X gates).
+    /// - `expansionOps`: An array of MatrixCompressionOp representing the operations to expand the state preparation (e.g., CX, X gates).
     /// - `numQubits`: The number of qubits to allocate for the state preparation.
     struct StatePreparationParams {
```
agamshayit-ms, line 18: This changes `expansionOps` type from `Int[][]` to `MatrixCompressionOp[]`; old `QSHARP_UTILS` calls will fail. Can we retain backward compatibility?

python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -310,12 +341,11 @@ def create(algorithm_type: str, algorithm_name: str | None = None, **kwargs) ->
     algorithm_type = _resolve_algorithm_type(algorithm_type)
     if algorithm_name is None:
         algorithm_name = ""
+    algorithm_name = _resolve_algorithm_name(algorithm_type, algorithm_name)
```
agamshayit-ms, line 344: `_resolve_algorithm_name` is only called here. Could `inspect_settings()` and `print_settings()` use it too? Both currently reject the deprecated name.

python/src/qdk_chemistry/plugins/qiskit/conversion.py
```diff
@@ -52,18 +56,20 @@ def create_statevector_from_wavefunction(wavefunction: data.Wavefunction, normal
...
-    num_qubits = num_orbs * 2
+    # Get determinants and coefficients
+    determinants = wavefunction.get_active_determinants()
+    coefficients = wavefunction.get_coefficients()
+
+    # Size the register from the wavefunction's own encoding so it matches the
+    # determinant bit-width. For standard spin-half wavefunctions this equals
+    # 2 * num_active_orbitals; reduced spinless subspaces use one qubit per mode.
+    config_set = wavefunction.get_configuration_set()
+    num_qubits = config_set.num_modes() * determinants[0].bits_per_mode()
```
agamshayit-ms, line 72: Can we guard against empty `determinants`? It may be that there are no active determinants, we need to check before accessing `determinants[0]`

python/tests/test_state_preparation_binary_encoding.py
```diff
@@ -0,0 +1,1009 @@
...
+        prep = create(
+            "state_prep",
+            "sparse_isometry",
+            binary_encoding=binary_encoding,
+            dense_state_prep=AlgorithmRef("state_prep", name),
+        )
+        with pytest.raises(ValueError, match="sparse isometry cannot compose"):
+            getattr(prep, stage)(ozone_wf)
+
+    @pytest.fixture(scope="class")
```
agamshayit-ms: This scope gives `PytestRemovedIn10Warning` for some reason. Can we replace it with a fixture like in lines 95 or 105 of this file?

agamshayit-ms (changes requested):

python/src/qdk_chemistry/utils/qsharp/src/StatePreparation.qs
RushiGong: Good catch — this is a real break, and Q# has no overloading, so there's no free fix. Laying out the tradeoff so we can decide:

**Who breaks:** any external Q# caller passing `expansionOps : Int[][]` to `MakeStatePreparationCircuit` (or constructing `StatePreparationParams` directly). It is documented surface — `docs/source/user/comprehensive/data/circuit.rst:75` names it explicitly. In-repo callers all pass `expansionOps: []`, which is why nothing here caught it.

**Why the type changed:** the old `Int[][]` encoding was positional and self-limiting — length 2 meant CNOT, length 1 meant X, anything else was `fail`. Binary encoding needs SELECT/SELECT_AND/MCX with control states and lookup tables, which that encoding cannot express.

**Options:**
1. Accept the break — cheapest, but silently breaks downstream Q#.
2. Add a legacy shim, e.g. `MakeStatePreparationCircuitLegacy(..., expansionOps : Int[][], ...)` that maps `[c,t] -> CX`, `[t] -> X` and delegates. Preserves compat at the cost of a second entry point to maintain and eventually deprecate.

Leaving this open pending a call — I don't want to lock in either without agreement. My mild preference is (2) if this is considered public surface, since the shim is ~10 lines and mechanical.

RushiGong: The Q# functions serve only for backend to support python layer functions. We don't expect Q# functions called directly by users. I don't think we should make effort to maintain Q# layers backward compatibility.

python/src/qdk_chemistry/algorithms/state_preparation/state_preparation.py
RushiGong: Added back in [e7e6fbe](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e7e6fbe7b8374e7f980bac59c24deb8b431fd07e) and it will raise deprecated warning if this is called.

python/src/qdk_chemistry/algorithms/registry.py
RushiGong: Added the `set()` in [e7e6fbe](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e7e6fbe7b8374e7f980bac59c24deb8b431fd07e)

python/tests/test_state_preparation_binary_encoding.py
RushiGong: Updated in [e7e6fbe](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e7e6fbe7b8374e7f980bac59c24deb8b431fd07e)

python/src/qdk_chemistry/plugins/qiskit/conversion.py
RushiGong: Added check for empty determinants in [e7e6fbe](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e7e6fbe7b8374e7f980bac59c24deb8b431fd07e)

python/src/qdk_chemistry/algorithms/registry.py
RushiGong: Added in `inspect_setting()` and `print_settings()` in commit [e7e6fbe](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e7e6fbe7b8374e7f980bac59c24deb8b431fd07e)

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
```diff
@@ -59,9 +59,170 @@ def convert(self, qir: pyqir.Module) -> QuantumCircuit:
             self._circuit = QuantumCircuit()
 
         # Build the circuit
-        self.run(qir)
+        self._walk_blocks(entry_point.basic_blocks[0])
```
agamshayit-ms: Do we need to check `basic_blocks` isn't empty? Something like
```
basic_blocks = entry_point.basic_blocks
if not basic_blocks:
    raise UnsupportedQIROperationError("QIR entry point has no function body.")
```

agamshayit-ms (changes requested):

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
RushiGong: Fixed in [57251dc](https://github.com/microsoft/qdk-chemistry/pull/435/commits/57251dc36c7e07d3473ab2f40e8db75a1b4c029f)

python/tests/test_helpers.py
```diff
@@ -247,3 +247,96 @@ def create_test_ansatz(num_orbitals: int = 2):
...
+    Returns:
+        Binary matrix of shape ``(2 * n_orbitals, n_dets)`` where rows are
+        qubits and columns are determinants.
+
+    """
+    n_alpha = n_electrons // 2
+    n_beta = n_electrons - n_alpha
+    rng = np.random.default_rng(seed)
+
+    hf = np.zeros(2 * n_orbitals, dtype=np.int8)
```
agamshayit-ms: Can you replace this with `Configuration.canonical_hf_configuration()`?

agamshayit-ms (changes requested):

python/tests/test_helpers.py
RushiGong: Updated in [5e3c850](https://github.com/microsoft/qdk-chemistry/pull/435/commits/5e3c8505bee865ed85ab08951618b70a7768f43c)

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
```diff
@@ -0,0 +1,757 @@
...
+            self._record(op)
+
+    def _synthesize_single_pui_lookup_block(self, sbs, fixed_controls, rest_entries):
+        """Lower one PUI sub-block into lookup ops.
+
+        Args:
+            sbs: Sparse block start row index (equal to dense_size).
+            fixed_controls: List of (row, polarity) pairs that are constant across the chunk.
+            rest_entries: List of (offset, changing_controls) pairs for this chunk.
+        c
```
agamshayit-ms: What's this c for?

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
```diff
@@ -59,9 +62,173 @@ def convert(self, qir: pyqir.Module) -> QuantumCircuit:
             self._circuit = QuantumCircuit()
 
         # Build the circuit
-        self.run(qir)
+        basic_blocks = entry_point.basic_blocks
+        if not basic_blocks:
+            raise UnsupportedQIROperationError("QIR entry point has no function body.")
```
agamshayit-ms, line 67: Can we add a test that shows this is raised?

agamshayit-ms (changes requested):

python/src/qdk_chemistry/algorithms/state_preparation/_binary_encoding_utils.py
RushiGong: Removed in [e297d89](https://github.com/microsoft/qdk-chemistry/pull/435/commits/e297d89505e9c54348741c1b7d0a43ff59494a56)

python/src/qdk_chemistry/plugins/qiskit/_interop/qir.py
RushiGong: Added in [2b19cdc](https://github.com/microsoft/qdk-chemistry/pull/435/commits/2b19cdc0b7c19a1d7ebed63db4bd3b71a13dbdd1)

agamshayit-ms (approved):

PR #437, opened 2026-04-07
Transpile to IBM fake backends

agamshayit-ms:
Added the ability to pass a device backend string to the Qiskit Aer circuit executor. If received as an argument, the circuit is transpiled to the backend and the simulator uses the backend's noise profile.


wavefunction91:
@v-agamshayit are we going to close this in favor of #431 ?

PR #440, opened 2026-04-14
Store Cholesky MO integrals and save them in serialization

rainli323:
Cholesky Hamiltonian used to save AO based integrals and save 4 center integrals during serialization. This PR did the following:

1. Cholesky Hamiltonian is no longer a derived class of canonical four center Hamiltonian
2. Cholesky Hamiltonian stores MO based integrals instead of AO, and it does not store the full 4-center integral (note changes in constructor breaks original API)
3. Cholesky Hamiltonian writes MO based 3-center integrals to disk when from_json and from_hdf5 is called.
4. The setting `store_cholesky_vectors` is changed to `store_ao_cholesky_vector`. It means we no longer want to store the result from CholeskyHamiltonianFactory inside CanonicalFourCenterContainer. We also added the ability to optionally store AO based Cholesky vector in the Hamiltonian container. This is done with optional parameter in constructing.

### DIFF
--- a/cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
+++ b/cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
@@ -42,6 +42,8 @@ class CholeskyHamiltonianContainer : public HamiltonianContainer {
    * energy)
    * @param inactive_fock_matrix Inactive Fock matrix for the selected active
    * space
+   * @param ao_cholesky_vectors Optional AO Cholesky vectors for potential reuse
+   * (default: std::nullopt)
    * @param type Type of Hamiltonian (Hermitian by default)
    *
    * @throws std::invalid_argument if orbitals pointer is nullptr
@@ -51,6 +53,7 @@ class CholeskyHamiltonianContainer : public HamiltonianContainer {
       const Eigen::MatrixXd& three_center_integrals,
       std::shared_ptr<Orbitals> orbitals, double core_energy,
       const Eigen::MatrixXd& inactive_fock_matrix,
+      std::optional<Eigen::MatrixXd> ao_cholesky_vectors = std::nullopt,
       HamiltonianType type = HamiltonianType::Hermitian);
 
   /**
@@ -74,6 +77,8 @@ class CholeskyHamiltonianContainer : public HamiltonianContainer {
    * the selected active space
    * @param inactive_fock_matrix_beta Inactive Fock matrix for beta spin in the
    * selected active space
+   * @param ao_cholesky_vectors Optional AO Cholesky vectors for potential reuse
+   * (default: std::nullopt)
    * @param type Type of Hamiltonian (Hermitian by default)
    *
    * @throws std::invalid_argument if orbitals pointer is nullptr
@@ -86,6 +91,7 @@ class CholeskyHamiltonianContainer : public HamiltonianContainer {
       std::shared_ptr<Orbitals> orbitals, double core_energy,
       const Eigen::MatrixXd& inactive_fock_matrix_alpha,
       const Eigen::MatrixXd& inactive_fock_matrix_beta,
+      std::optional<Eigen::MatrixXd> ao_cholesky_vectors = std::nullopt,
       HamiltonianType type = HamiltonianType::Hermitian);
 
   /**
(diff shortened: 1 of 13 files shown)


cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
```diff
@@ -120,10 +109,67 @@ class CholeskyHamiltonianContainer
 
   /**
    * @brief Get the type of the underlying container
-   * @return String "cholesky" identifying this container type
+   * @return String identifying the container type (e.g.,
+   * "canonical_four_center", "cholesky")
```
wavefunction91: This should only ever return `"cholesky"`

cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
```diff
@@ -8,109 +8,98 @@
...
- * This class stores molecular Hamiltonian data for quantum chemistry
- * calculations, specifically designed for active space methods. It contains:
- * - One-electron integrals (kinetic + nuclear attraction) in MO representation
- * - Two-electron integrals (electron-electron repulsion) in MO representation
- * - Cholesky vectors in AO representation for reconstructing integrals
- * - Molecular orbital information for the active space
- * - Inactive Fock matrix
- * - Core energy contributions from inactive orbitals and nuclear repulsion
+ * @brief Contains a molecular Hamiltonian using three-center
+ * integrals.
```
wavefunction91:
```suggestion
 * @brief Contains a molecular Hamiltonian expressed using three-center
 * integrals.
```

cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
```diff
@@ -140,24 +186,75 @@ class CholeskyHamiltonianContainer
...
+                  std::shared_ptr<Eigen::MatrixXd>>
+      _three_center_integrals;
+
+  /**
+   * Lazily computed four-center integrals cache (built on first access).
+   * Stores (aaaa, aabb, bbbb) as flattened arrays [norb^4].
+   * Uses shared_ptr so restricted case can share the same data for all
+   * channels.
+   */
+  mutable std::optional<std::tuple<std::shared_ptr<Eigen::VectorXd>,
```
wavefunction91: If these are shared ptrs, why are they `optional`?

cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.cpp
```diff
@@ -804,17 +776,17 @@ std::shared_ptr<data::Hamiltonian> CholeskyHamiltonianConstructor::_run_impl(
...
-      if (!_settings->get<bool>("store_cholesky_vectors")) {
+      if (_settings->get<bool>("store_ao_cholesky_vectors")) {
         return std::make_shared<data::Hamiltonian>(
-            std::make_unique<data::CanonicalFourCenterHamiltonianContainer>(
-                H_active, moeri_aaaa, orbitals,
-                structure->calculate_nuclear_repulsion_energy(), dummy_fock));
+            std::make_unique<data::CholeskyHamiltonianContainer>(
+                H_active, L_mo, orbitals,
+                structure->calculate_nuclear_repulsion_energy(), dummy_fock,
+                Eigen::MatrixXd(L_ao)));
```
wavefunction91: Why are we making an extra copy of `L_ao`?

cpp/src/qdk/chemistry/data/hamiltonian_containers/cholesky.cpp
```diff
@@ -274,27 +594,192 @@ CholeskyHamiltonianContainer::from_json(const nlohmann::json& j) {
   }
 }
 
+void CholeskyHamiltonianContainer::to_fcidump_file(const std::string& filename,
```
wavefunction91: Why is this a whole new function? It should just call the base implementation with the cached 4 center integrals. This may require the method of access to change upstream.

wavefunction91 (changes requested):

wavefunction91 (approved):

PR #453, opened 2026-04-21
Add ROHF GDM

### LINKED ISSUE #142: Add ROHF to QDK SCF
We support ROHF via the pyscf implementation, but not in the QDK implementation. This degrades user experience for those expecting feature parity. ROHF should be added to the QDK SCF to address this issue.

BoqinZhang:


cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -74,6 +74,242 @@ static void apply_orbital_rotation(RowMajorMatrix& C, const int spin_index,
...
+  RowMajorMatrix F_A = J_alpha_mo - J_beta_mo - 0.5 * (K_alpha_mo - K_beta_mo);
+  RowMajorMatrix Q = J_alpha_mo - J_beta_mo - (K_alpha_mo - K_beta_mo);
+
+  RowMajorMatrix F_sum = F_I + F_A;
+  generalized_fock_mo.block(0, 0, num_closed_orbitals, num_molecular_orbitals) =
+      2.0 * F_sum.block(0, 0, num_closed_orbitals, num_molecular_orbitals);
+  generalized_fock_mo.block(num_closed_orbitals, 0, num_open_orbitals,
+                            num_molecular_orbitals) =
+      (F_I + Q).block(num_closed_orbitals, 0, num_open_orbitals,
+                      num_molecular_orbitals);
```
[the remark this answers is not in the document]
BoqinZhang: No. We cannot generate ROHF gradient from the spin-blocked Fock matrix: we have to know the value of `J` and `K` matrix for each spin.
For ROKS, currently we don't have a plan to implement it. We need to add code to throw error message when user selects ROKS.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -194,6 +233,168 @@ void SCFAlgorithm::update_density_matrix(RowMajorMatrix& P,
...
+                                         RowMajorMatrix& effective_fock,
+                                         RowMajorMatrix& total_density) {
+  QDK_LOG_TRACE_ENTERING();
+  const int num_atomic_orbitals = static_cast<int>(C.rows());
+  const int num_molecular_orbitals = static_cast<int>(C.cols());
+  if (num_atomic_orbitals != num_molecular_orbitals) {
+    throw std::runtime_error(
+        "ROHF build requires number of atomic orbitals to equal number of "
+        "molecular orbitals!");
+  }
```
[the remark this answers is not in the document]
BoqinZhang: Yes. Currently ROHF calculation doesn't support the case in which NMO < NAO. The code should report error once it happened.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -7,9 +7,11 @@
 #include <qdk/chemistry/scf/config.h>
 
 #include <cmath>
+#include <cstdint>
 #include <limits>
 #include <qdk/chemistry/utils/logger.hpp>
 #include <stdexcept>
+#include <vector>
```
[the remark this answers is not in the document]
BoqinZhang: `#include <lapack.hh>` is in line 24

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -30,8 +32,44 @@
 #include <nvtx3/nvtx3.hpp>
 #endif
 
+#include <blas.hh>
```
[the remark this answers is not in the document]
BoqinZhang: `#include <lapack.hh>` is in line 24

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf/scf_impl.cpp
```diff
@@ -1035,9 +1046,18 @@ void SCFImpl::write_gradients_(const std::vector<double>& gradients,
 
 std::pair<double, RowMajorMatrix>
 SCFImpl::evaluate_trial_density_energy_and_fock(
-    const RowMajorMatrix& P_matrix, const std::source_location& loc) const {
+    const RowMajorMatrix& P_matrix, double* J_out, double* K_out,
+    const std::source_location& loc) const {
   QDK_LOG_TRACE_ENTERING();
 
+  // TODO: This function is called within iterate_ which runs only on rank 0,
```
agamshayit-ms: Do we keep TODO comments?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -194,6 +261,168 @@ void SCFAlgorithm::update_density_matrix(RowMajorMatrix& P,
   }
 }
 
+std::pair<const RowMajorMatrix&, const RowMajorMatrix&>
+SCFAlgorithm::try_get_rohf_convergence_matrices(const SCFImpl& scf_impl) {
```
agamshayit-ms: Does this function need to be public?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -194,6 +261,168 @@ void SCFAlgorithm::update_density_matrix(RowMajorMatrix& P,
...
+  temp_rhs.transposeInPlace();
+  info = lapack::getrs(lapack::Op::NoTrans, matrix_dim, matrix_dim, Ct.data(),
+                       matrix_dim, ipiv.data(), temp_rhs.data(), matrix_dim);
+  if (info != 0) {
+    throw std::runtime_error("getrs failed while solving C^T X = M^T");
+  }
+
+  effective_fock = temp_rhs.transpose();
+  if (!effective_fock.isApprox(effective_fock.transpose())) {
+    effective_fock = 0.5 * (effective_fock + effective_fock.transpose().eval());
```
agamshayit-ms, line 366: Should we document this behavior somehow? or check that they are close within some threshold? This could potentially mask an error

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
BoqinZhang: Yes. This function is called at line 626 of `algorithms/microsoft/scf/src/scf/scf_impl.cpp`, which is outside of class `SCFAlgorithm`.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
BoqinZhang: A warning is added to remind the user

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf/scf_impl.cpp
BoqinZhang: Removed TODO comment here and added explanation in error message

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf/scf_impl.cpp
```diff
@@ -1056,12 +1076,22 @@ SCFImpl::evaluate_trial_density_energy_and_fock(
...
                  omega);
+
+  if (J_out != nullptr) {
+    std::memcpy(J_out, J_matrix.data(),
+                sizeof(double) * static_cast<size_t>(J_matrix.size()));
+  }
+  if (K_out != nullptr) {
+    std::memcpy(K_out, K_matrix.data(),
+                sizeof(double) * static_cast<size_t>(K_matrix.size()));
+  }
```
wavefunction91: This grafts two C-style out-pointers onto a function that already returns `{energy, F}`, and `memcpy`s `J_matrix.size()` doubles into them with no check that `J_out`/`K_out` are actually that large — a silent buffer overflow if a caller  mis-sizes the buffer. The only caller (ROHF `GDMLineFunctor::eval`) already has `build_jk_matrices` available. Please drop the `J_out`/`K_out` params (here and in the `scf_impl.h:235` / `ks_impl` overrides) and have the ROHF path call  `build_jk_matrices` for its J/K, or return them in the result. Passing `double*` instead of `RowMajorMatrix&` also throws away all shape information.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -601,10 +1250,15 @@ void GDM::iterate(SCFImpl& scf_impl) {
   const auto& F = scf_impl.get_fock_matrix();
 
   const auto* cfg = ctx_.cfg;
+  // TODO: This function is called within iterate_ which runs only on rank 0,
+  // but build_jk_matrices requires participation from all MPI ranks.
+  if (cfg->mpi.world_size > 1) {
+    throw std::runtime_error(
+        "Temporary limitation: evaluate_trial_density_energy_and_fock is not "
+        "supported with MPI world_size > 1.");
+  }
```
wavefunction91, line 1273: The thrown text names `evaluate_trial_density_energy_and_fock`, but the comment right above (correctly) blames `build_jk_matrices`, which is what actually requires all ranks here. This is a copy-paste of the guard in `scf_impl.cpp:1053`;  the message contradicts the code. Either fix the message to name `build_jk_matrices`, or lift this `world_size > 1` limitation into a single shared check instead of guarding it in two layers.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_algorithm.cpp
```diff
@@ -30,8 +32,44 @@
 #include <nvtx3/nvtx3.hpp>
 #endif
 
+#include <blas.hh>
+
 namespace qdk::chemistry::scf {
 
+void compute_atba_gemm(const double* A, const double* B, double* C, int m,
```
wavefunction91: `compute_atba_gemm` is defined here but declared in `scf_helper.h`, which this TU doesn't include (nor is it declared in `scf_algorithm.h`). The definition is therefore never checked against the declaration at compile time — a signature  drift becomes a link error instead of a compile error. Please declare it in `scf_algorithm.h` next to this definition and include that header from the users.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/scf_helper.h
```diff
@@ -5,11 +5,44 @@
...
+ * The intermediate workspace is provided by caller to avoid repeated
+ * allocations across consecutive calls.
+ *
+ * @param[in] A Pointer to the A matrix block.
+ * @param[in] B Pointer to the B matrix block.
+ * @param[out] C Pointer to output C matrix block.
+ * @param[in,out] workspace Temporary buffer of at least m*n doubles.
+ * @param[in] layout BLAS matrix layout (RowMajor by default).
+ */
+void compute_atba_gemm(const double* A, const double* B, double* C, int m,
```
wavefunction91: Renaming `line_search.h` -> `scf_helper.h` just to host this GEMM declaration turns it into a junk-drawer header (a line-search template in `impl` plus an unrelated BLAS helper in the outer namespace). If `compute_atba_gemm` is meant to  be shared, give it its own `.h/.cpp`; otherwise keep it with `scf_algorithm`. Please don't repurpose the line-search header.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -378,26 +728,28 @@ class GDM {
...
   double last_accepted_energy_;
-  int gdm_step_count_;        // GDM iteration counter
-  int num_density_matrices_;  // Number of density matrices (1 for restricted, 2
-                              // for unrestricted)
+  int gdm_step_count_;  // GDM iteration counter
+
+  // Number of spin blocks (1 for restricted, 2 for unrestricted)
+  const int num_orbital_spin_blocks_;
+  // Number of density matrices (1 for closed shell, 2 for open shell)
+  const int num_density_matrices_;
```
wavefunction91: `GDM::num_density_matrices_` is initialized (L751) but never read anywhere in the `GDM` class — only `num_orbital_spin_blocks_` is used. Please remove it. (The `GDMLineFunctor` copy at L465 is used; this class-level one is dead.)

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -157,14 +461,18 @@ class GDMLineFunctor {
   const std::vector<int>& rotation_size_;
 
   // Value parameters
+  const int num_orbital_spin_blocks_;
```
wavefunction91: Carrying both `num_orbital_spin_blocks_` (RHF 1 / UHF 2 / ROHF 1) and `num_density_matrices_` (RHF 1 / UHF 2 / ROHF 2) side by side  is a footgun, and the GDM-class copy (L741) is already dead.  Consider deriving these from `scf_orbital_type` in one small helper/struct instead of recomputing the mapping in `GDM`, `GDMLineFunctor`, and the base class.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -615,44 +1269,25 @@ void GDM::iterate(SCFImpl& scf_impl) {
...
-    const int rotation_size = num_occupied_orbitals * num_virtual_orbitals;
-    RowMajorMatrix F_MO =
-        C.block(num_molecular_orbitals * i, 0, num_molecular_orbitals,
-                num_molecular_orbitals)
-            .transpose() *
-        F.block(num_molecular_orbitals * i, 0, num_molecular_orbitals,
-                num_molecular_orbitals) *
-        C.block(num_molecular_orbitals * i, 0, num_molecular_orbitals,
-                num_molecular_orbitals);
+  if (cfg->scf_orbital_type == SCFOrbitalType::RestrictedOpenShell) {
```
wavefunction91, line 1289: This ROHF branch rebuilds a full alpha/beta J/K from the SCF density every iteration, whereas RHF/UHF reuse the existing Fock `F` with no two-electron build. If the separate-spin J/K genuinely can't be recovered from the effective Fock,  please add a one-line comment saying so; otherwise reuse the J/K already built in `SCFImpl::iterate_`.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/core/scf_algorithm.h
```diff
@@ -113,6 +115,20 @@ class SCFAlgorithm {
...
+   * For ROHF, some algorithms evaluate convergence using an effective Fock
+   * matrix and total density matrix rather than the spin-blocked SCFImpl
+   * matrices. This helper returns references to those matrices.
+   *
+   * @param[in] scf_impl SCF implementation object
+   * @return Pair of references to convergence Fock and density matrices
+   * @throws std::logic_error If ROHF convergence matrices are unavailable
+   */
+  std::pair<const RowMajorMatrix&, const RowMajorMatrix&>
+  try_get_rohf_convergence_matrices(const SCFImpl& scf_impl);
```
wavefunction91: `try_get_rohf_convergence_matrices` reads like a non-throwing "try" API (bool / `std::optional`), but it throws `std::logic_error` for non-ROHF and returns a pair of references. The `try_` prefix is misleading and feels very C. Suggest  `rohf_convergence_matrices()`, or `build_rohf_convergence_matrices()` since it rebuilds the cache. The `get_rohf_convergence_*` siblings at L183/L188/L193 read fine.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/gdm.cpp
```diff
@@ -181,35 +489,59 @@ double GDMLineFunctor::eval(const Eigen::VectorXd& x) {
...
-        cached_C_
-            .block(num_molecular_orbitals_ * i, 0, num_molecular_orbitals_,
-                   num_occupied_orbitals)
-            .transpose();
+    const double occupation_factor =
+        (scf_orbital_type_ == SCFOrbitalType::Restricted) ? 2.0 : 1.0;
+    auto P_block = cached_P_.block(num_atomic_orbitals_ * i, 0,
+                                   num_atomic_orbitals_, num_atomic_orbitals_);
+
+    double* C_block_data = nullptr;
```
wavefunction91: Taking `C_block.data()` from a `Block` declared inside each if/else arm and using it after the arm closes works only because it aliases `cached_C_` — it reads as a dangling-pointer footgun. The branch only selects offset `0` vs  `num_atomic_orbitals_ * i`; collapse it to `const int c_row = (rohf) ? 0 : num_atomic_orbitals_ * i;` and a single `cached_C_.data() + c_row * num_molecular_orbitals_` expression.

wavefunction91 (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/core/scf_algorithm.h
```diff
@@ -7,11 +7,42 @@
...
+ *
+ * @param[in] A Pointer to the A matrix block.
+ * @param[in] B Pointer to the B matrix block.
+ * @param[out] C Pointer to output C matrix block.
+ * @param[in,out] workspace Temporary buffer of at least m*n doubles.
+ * @param[in] layout BLAS matrix layout (RowMajor by default).
+ */
+void compute_atba_gemm(const double* A, const double* B, double* C, int m,
+                       int n, std::vector<double>& workspace,
+                       blas::Layout layout = blas::Layout::RowMajor);
```
wavefunction91: Have it take this signature

```
/// C = ALPHA * A**H * B * A + BETA * C
similarity_transform(blas::Layout layout, int64_t N, int64_t K, double ALPHA, const double* A, int64_t LDA, const double* B, int64_t LDB, double BETA, double* C, int LDC, std::vector<double>* workspace = nullptr);
```

wavefunction91: Also, have this live somewhere other than this file - this is a general BLAS utility that could likely be useful elsewhere.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/line_search.h
```diff
@@ -9,7 +9,9 @@
 #include <cstddef>
 #include <stdexcept>
 
-namespace qdk::chemistry::scf::impl {
+namespace qdk::chemistry::scf {
+
+namespace impl {
```
wavefunction91: Why did this change?

wavefunction91 (changes requested):

BoqinZhang:

wavefunction91 (approved):

PR #457, opened 2026-04-24
Computing <S^2> for wavefunction containers

### LINKED ISSUE #202: Add computation of <S^2> for Wavefunction types
<S^2> is a common diagnostic for Wavefunction types, particularly single reference wavefunctions. It would be useful to compute this in the API, including fall throughs for exact / trivial cases (e.g. RHF).

agamshayit-ms:
Added a function to compute <S^2> for arbitrary wavefunction containers using 1- and 2-RDMs consistent with the chemist's convention: Gamma(p,q,r,s) = <p^\dagger r^\dagger s q>.
Added some CASCI and SCI tests to verify correctness.


cpp/src/qdk/chemistry/data/wavefunction.cpp
```diff
@@ -577,6 +577,99 @@ Eigen::MatrixXd WavefunctionContainer::get_mutual_information() const {
...
+    for (int i = 0; i < norbs; ++i) {
+      for (int j = 0; j < norbs; ++j) {
+        int idx = i * norbs * norbs * norbs + j * norbs * norbs + j * norbs + i;
+        using ValueType =
+            std::decay_t<decltype(vec[static_cast<Eigen::Index>(0)])>;
+        if constexpr (std::is_same_v<ValueType, std::complex<double>>) {
+          sum += vec[idx].real();
+        } else {
+          sum += vec[idx];
+        }
```
wavefunction91: *nit*: Encapsulate this into something like a `smart_real` which passes through on FP types. Repeating this is a bit silly. Also, `if constexpr` is alright, but template partial overloads are probably cleaner and a bit easier for the compiler to understand.

cpp/tests/test_s_squared.cpp
```diff
@@ -0,0 +1,660 @@
...
+      std::nullopt,                        // no spin-traced 1-RDM
+      std::make_optional(one_rdm_aa),      // alpha 1-RDM
+      std::make_optional(one_rdm_bb),      // beta 1-RDM
+      std::nullopt,                        // no spin-traced 2-RDM
+      std::make_optional(two_rdm_aabb),    // mixed-spin 2-RDM
+      std::make_optional(two_rdm_aaaa),    // alpha-alpha 2-RDM
+      std::make_optional(two_rdm_bbbb)));  // beta-beta 2-RDM
+
+  double s_squared = wf.compute_s_squared();
+  EXPECT_NEAR(s_squared, 0.0, 1e-12);
```
wavefunction91: Use the "standard" FP reference tolerances for these.

cpp/tests/test_s_squared.cpp
```diff
@@ -0,0 +1,660 @@
...
+// Force unrestricted orbitals to restricted (MACIS requires restricted)
+static std::shared_ptr<Orbitals> make_restricted(
+    std::shared_ptr<Orbitals> orbitals) {
+  if (orbitals->is_restricted()) return orbitals;
+  return std::make_shared<Orbitals>(
+      orbitals->get_coefficients().first, orbitals->get_energies().first,
+      orbitals->get_overlap_matrix(), orbitals->get_basis_set(),
+      std::make_tuple(orbitals->get_active_space_indices().first,
+                      orbitals->get_inactive_space_indices().first));
+}
```
wavefunction91: We have ROHF for a reason

cpp/tests/test_s_squared.cpp
```diff
@@ -0,0 +1,660 @@
...
+  coeffs(0) = 1.0;
+  Wavefunction::DeterminantVector dets = {Configuration("ud")};
+
+  auto wf = Wavefunction(
+      std::make_unique<CasWavefunctionContainer>(coeffs, dets, orbitals));
+
+  EXPECT_THROW(wf.compute_s_squared(), std::runtime_error);
+}
+
+// ---------------------------------------------------------------------------
```
wavefunction91, line 362: I like these integration tests, although I might put them in a different file. The one I would also like to see is a spin-contaminated UHF state with a non-trivial FP expectation value

wavefunction91 (changes requested):

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -395,6 +395,27 @@ class WavefunctionContainer {
...
+   *     \left(\gamma^\alpha_{ii} + \gamma^\beta_{ii}\right)
+   *     - \sum_{ij} \left( \Gamma^{\alpha\beta}_{ijji}
+   *     + \tfrac{1}{4} \Gamma^{\alpha\alpha}_{ijji}
+   *     + \tfrac{1}{4} \Gamma^{\beta\beta}_{ijji} \right)
+   *     - \tfrac{1}{2} \sum_{ij} \Gamma^{\alpha\beta}_{iijj}
+   *
+   * @return The expectation value \f$\langle S^2 \rangle\f$
+   * @throws std::runtime_error if spin-dependent 1-RDMs or 2-RDMs are not
+   *         available
+   */
```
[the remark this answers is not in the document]
agamshayit-ms: When calling has_one_rdm_spin_traced(), if it is not initialized, we construct it from the spin-dependent RDM. If the spin-dependent RDM doesn't exist, then we error out. Therefore, the code functions as intended

wavefunction91:
@v-agamshayit can you revive this with the new WFN infrastructure?

cpp/tests/test_s_squared_integration.cpp
```diff
@@ -0,0 +1,230 @@
...
+TEST(SSquaredCAS, StretchedH2_Triplet) {
+  std::vector<Eigen::Vector3d> coords = {{0., 0., 0.}, {0., 0., 5.0}};
+  std::vector<std::string> symbols = {"H", "H"};
+  auto h2 = std::make_shared<Structure>(coords, symbols);
+  auto wfn = run_cas_with_rdms(h2, 0, 3, "sto-3g", 2, 0);
+  EXPECT_NEAR(wfn->compute_s_squared(), 2.0, testing::rdm_tolerance);
+}
+
+// ---------------------------------------------------------------------------
+// Tests using SCF + SCI (ASCI) with RDMs from MACIS
```
mmoerchen: I think these SCI tests may actually be running FCI. If that's the case, are they effectively duplicates of the existing FCI tests?

mmoerchen (commented):
Can we combine `test_s_squared.cpp` and `test_s_squared_integration.cpp` into one file?
I'm not sure the current split buys us much.

wavefunction91 (approved):

PR #461, opened 2026-04-28
Add native cube generator backed by GauXC OrbitalEvaluator

ConradJohnston:
## Why

`generate_cubefiles_from_orbitals` went through PySCF. PySCF is pinned
`sys_platform != 'win32'` in `python/pyproject.toml`, and `utils/cubegen.py`
imported `pyscf.tools.cubegen` at module scope, so the module was unimportable
on Windows and cube generation was simply unavailable there. It also meant
orbital visualization pulled in a third-party quantum chemistry package for
something the library can evaluate itself.

## How

Adds a native `CubeGenerator` in `qdk::chemistry::utils` that evaluates orbitals
and densities on a `CubeGrid` using GauXC's `OrbitalEvaluator`, writes Gaussian
cube files through `GauXC::write_cube`, and is exposed to Python. It is now the
default backend for `generate_cubefiles_from_orbitals`, which gives Windows cube
generation for the first time. `backend="pyscf"` still works wherever PySCF is
installed.

The switch is behavior preserving: both backends place grid points identically,
with the origin at the nuclear bounding box corner minus the margin and the step
equal to the padded extent divided by `n - 1` along each axis, and both use the
same atomic orbital ordering.

GauXC's cube-generation support is now upstream, so the dependency moves off the
temporary fork and pins `wavefunction91/GauXC` at merge commit
`162e4562552323a871af17ae4acd73b71071bd24`.

Two dependency patches were needed to build across platforms:

- A vcpkg overlay for the existing `libaec` 1.1.6, retargeting the archive from
  the intermittently failing GitLab URL to libaec's canonical GitHub mirror.
  Recipe and version are unchanged.
- A clang-cl-only, idempotent patch for GauXC's vendored Gau2Grid helper, which
  calls `exit()` without including `<stdlib.h>` on that preprocessor path.

Neither affects cube-generation behavior or adds a runtime dependency.

## API

```cpp
CubeGenerator gen(basis_set);
auto field = gen.orbital(mo_coeff, "orbital.cube", grid);
auto rho = gen.density(D, "density.cube", grid);
```


cpp/tests/test_cube_generator.cpp
```diff
@@ -0,0 +1,149 @@
...
+// license information.
+
+#include <gtest/gtest.h>
+
+#include <Eigen/Dense>
+#include <cmath>
+#include <cstdint>
+#include <limits>
+#include <memory>
+#include <numbers>
```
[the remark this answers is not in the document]
ConradJohnston: Fixed in efc2b76d: the test now computes pi with std::acos(-1.0) and no longer includes the C++20-only <numbers> header.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,206 @@
...
+  std::shared_ptr<data::BasisSet> basis_set;
+  GauXC::BasisSet<double> gauxc_basis;
+  GauXC::Molecule gauxc_mol;
+  GauXC::OrbitalEvaluator evaluator;
+  int32_t nbf;
+
+  explicit Impl(std::shared_ptr<data::BasisSet> bs)
+      : basis_set(std::move(bs)),
+        gauxc_basis(to_gauxc_basis(require_basis_set(basis_set))),
+        gauxc_mol(to_gauxc_mol(*require_basis_set(basis_set).get_structure())),
```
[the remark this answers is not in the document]
ConradJohnston: Fixed in efc2b76d: the already-validated basis set is reused when constructing the GauXC molecule.

cpp/cmake/patches/gauxc-clang-cl-gau2grid-stdlib.cmake
```diff
@@ -0,0 +1,13 @@
...
+set(_helper "external/gau2grid/generated_source/gau2grid_helper.c")
+file(READ "${_helper}" _content)
+
+if(NOT _content MATCHES "#include <stdlib.h>\n#include <math.h>")
+  string(REPLACE "#include <math.h>"
+                 "#include <stdlib.h>\n#include <math.h>"
+                 _content "${_content}")
+  file(WRITE "${_helper}" "${_content}")
+  message(STATUS "Patched Gau2Grid helper for clang-cl stdlib declarations")
+endif()
```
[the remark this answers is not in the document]
ConradJohnston: Hardened in efc2b76d: the script now detects stdlib.h regardless of include ordering or line endings, avoids duplicate insertion, and emits explicit diagnostics if the pinned Gau2Grid layout is incompatible.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,206 @@
...
+    if (std::size_t(C_a.cols()) <= p)
+      throw std::out_of_range("generate_orbital_cubes: index OOB.");
+    const auto filename = prefix + [&] {
+      char index[32];
+      std::snprintf(index, sizeof(index), "%04zu.cube", p);
+      return std::string(index);
+    }();
+    auto path = (std::filesystem::path(output_dir) / filename).string();
+    gen.orbital(C_a.col(p), path, grid,
+                "Orbital " + std::to_string(p) + " (alpha)");
```
RushiGong: Some observations on inconsistency with Python layer `cubegen.py`: 
- The Python version writes both `_a` and `_b` cubes, while here seems alpha only.
- The Python version does `f"orbital_{p + 1:04d}"` for labeling the orbital index, but here does `%04zu` on the 0-based index.

Do we want this to be consistent with the Python layer?

ConradJohnston: Thanks — good catch. I've split these into two, since I think they warrant different answers.

**Spin channels — fixed** (90798ee8). `generate_orbital_cubes` now branches on `orbitals->is_restricted()`: unrestricted wavefunctions write separate `_a` and `_b` cubes (with "(alpha)"/"(beta)" comments), restricted wavefunctions write a single suffix-less spatial cube. That matches `cubegen.py`'s behaviour. Added unit tests covering both paths plus out-of-range index handling.

**Index labeling — intentionally kept 0-based.** I'd argue `cubegen.py` is the outlier rather than the reference here:

- `generate_orbital_cubes`' own `indices` argument is 0-based, so `p + 1` labeling would mean a caller passing `indices={0}` gets `orbital_0001.cube` — the function would disagree with itself.
- The rest of the codebase is 0-based: `coupled_cluster.py` frozen-orbital indices ("0-based"), `spin_channel_indices.hpp` ("flat zero-based vector"), and the phase-estimation iteration docs.

So rather than propagating `p + 1` into C++, I've filed #639 to make the Python layer 0-based. Keeping it separate because that change renames output files consumed by `tutorial_choose_active_space` and existing users, so it deserves its own PR + changelog note instead of riding along here.

I've also documented the labeling and spin-suffix behaviour in the `generate_orbital_cubes` header comment. Happy to revisit if you'd prefer the C++ side match Python's 1-based labels in the interim.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+  const auto st = qdk.get_structure();
+  const Eigen::MatrixXd& coords = st->get_coordinates();
+  const bool sph = qdk.get_atomic_orbital_type() == data::AOType::Spherical;
+
+  GauXC::BasisSet<double> basis;
+  for (std::size_t ia = 0; ia < qdk.get_num_atoms(); ++ia) {
+    CA center{coords(ia, 0), coords(ia, 1), coords(ia, 2)};
+    for (const auto& sh : qdk.get_shells_for_atom(ia)) {
+      if (sh.has_radial_powers())
```
agamshayit-ms: his also rejects ordinary named bases: `from_basis_name()` fills electron-shell `rpowers` with zeros, so even STO-3G makes this condition true. Please handle that representation and add a named-basis test.

cpp/cmake/third_party.cmake
```diff
@@ -120,12 +120,19 @@ set(GAUXC_ENABLE_CUDA ${QDK_CHEMISTRY_ENABLE_GPU} CACHE BOOL "Enable gauxc CUDA
...
+set(_gauxc_patch_args "")
+if(CMAKE_C_COMPILER_ID MATCHES "Clang" AND CMAKE_C_SIMULATE_ID STREQUAL "MSVC")
+  set(_gauxc_patch_args FETCHCONTENT_ARGS
+      PATCH_COMMAND "${CMAKE_COMMAND}" -P "${CMAKE_CURRENT_LIST_DIR}/patches/gauxc-clang-cl-gau2grid-stdlib.cmake"
+  )
+endif()
 handle_dependency(gauxc
   GIT_REPOSITORY https://github.com/wavefunction91/gauxc.git
-  GIT_TAG f05cd68e1fd549cc45a318e6d039f49d044d3e1d
+  GIT_TAG 162e4562552323a871af17ae4acd73b71071bd24
```
agamshayit-ms, line 136: This pin is bypassed when `handle_dependency` finds an installed GauXC. An older supported installation configures successfully but then fails because `gauxc/external/cube.hpp` is missing; both packages report version `1.0.0`. We can feature-check pre-installed GauXC/force the pinned fetch.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,63 @@
...
+  explicit CubeGenerator(std::shared_ptr<data::BasisSet> basis_set);
+  ~CubeGenerator() noexcept;
+  CubeGenerator(CubeGenerator&&) noexcept;
+  CubeGenerator& operator=(CubeGenerator&&) noexcept;
+
+  CubeField orbital(const Eigen::VectorXd& mo_coeff, const std::string& outfile,
+                    const CubeGrid& grid,
+                    const std::string& comment = "") const;
+
+  CubeField density(const Eigen::MatrixXd& density_matrix,
```
agamshayit-ms, line 196: Please document the expected AO ordering and whether this matrix is spin-resolved or spin-summed.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+  g.spacing[1] = ny > 1 ? extent[1] / double(ny - 1) : 0.0;
+  g.spacing[2] = nz > 1 ? extent[2] / double(nz - 1) : 0.0;
+  return g;
+}
+
+std::size_t CubeGrid::num_points() const {
+  if (nx == 0 || ny == 0 || nz == 0)
+    throw std::invalid_argument("CubeGrid: dimensions must be positive.");
+  constexpr auto max =
+      static_cast<std::size_t>(std::numeric_limits<int64_t>::max());
```
agamshayit-ms: GauXC orbital evaluation is limited to `INT_MAX` points, but this allows int64_t counts and allocates first. Please reject oversized grids before allocation.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
agamshayit-ms, line 1: The PR says it replaces PySCF, but Python still calls pyscf.tools.cubegen. Please wire the native path or narrow the PR description

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Good catch, this was a real defect. Fixed in `a89dc566`.

`has_radial_powers()` is `rpowers.size() > 0`, i.e. "array present", not "has a non-zero radial power". Since `get_basis_for_nuclear_charge` pushes a zero per primitive for ordinary electron shells, every named basis tripped it. Now testing the physical condition:

```cpp
if (sh.has_radial_powers() && (sh.rpowers.array() != 0).any())
```

The suite was green because all existing tests used the 4-arg `Shell` ctor (`rpowers` size 0), which is a representation `from_basis_name` never produces. Your named-basis test ask is what actually closes that gap.

Chasing it also showed the guard was never protecting against ECPs: real projectors live in `_ecp_shells_per_atom`, which `to_gauxc_basis` doesn't traverse. So ECP bases were only ever blocked by this bug. The genuine consequence is interpretive, in that an ECP cube is valence-only, so I've documented that on `CubeGenerator` and the cube comment now records `valence-only: ECP replaces N core electrons`.

Tests (suite now 15): named STO-3G evaluation (confirmed failing before the fix), explicit zero rpowers accepted, non-zero rpowers still rejected, plus ECP evaluation and comment annotation.

cpp/cmake/third_party.cmake
ConradJohnston: Confirmed and fixed in `d5f15241`. Leaving this open for your call on the approach.

You're right that a version floor can't help here. Discovery runs before FetchContent in `handle_dependency`, so `GIT_TAG` is only a fallback, and since both revisions advertise `1.0.0` the stale install is a perfect match for any constraint the pinned one would satisfy. The distinguishing feature (`cube.hpp`) is invisible to `find_package`, so configure succeeds and the failure surfaces much later as a missing header attributed to `cube_generator.cpp`.

I went with forcing the pinned fetch rather than feature-checking, mainly on component-governance grounds: gauXC is pinned to an unreleased merge commit recorded in `cgmanifest.json`, so honouring a system install means building something other than the declared artifact even in the case where it does have `cube.hpp`. Feature-checking is the better general answer, but it needs a post-discovery hook in `handle_dependency`, which is shared by every dependency, so that is more blast radius than this one problem warrants right now.

Implemented as an opt-in `FORCE_FETCH` flag. Without it the discovery path is byte-for-byte unchanged. Discovery is still attempted when `QDK_ALLOW_DEPENDENCY_FETCH` is off, since a system package is then the only way to satisfy the dependency at all.

Verified a full configure takes the forced-fetch path (`Skipping discovery for gauxc - building pinned source`) and completes successfully, which also exercises every other `handle_dependency` call. I did not reproduce your stale-install scenario directly, as there is no older GauXC on this machine, so if you have that environment handy a confirmation would be worth having.

Happy to switch to the feature-check if you'd rather keep system-package reuse.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+  Eigen::Vector3d spacing{0.2, 0.2, 0.2};
+  std::size_t nx = 80, ny = 80, nz = 80;
+
+  static CubeGrid from_basis_set(const data::BasisSet&, std::size_t nx = 80,
+                                 std::size_t ny = 80, std::size_t nz = 80,
+                                 double margin = 3.0);
+  std::size_t num_points() const;
+};
+
+// Evaluates molecular orbitals and densities on a `CubeGrid` and optionally
```
mmoerchen: Please use Doxygen-style comments here to keep the API documentation consistent with the rest of the codebase.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+namespace qdk::chemistry::data {
+class BasisSet;
+class Wavefunction;
+}  // namespace qdk::chemistry::data
+
+namespace qdk::chemistry::utils {
+
+using CubeField = std::vector<double>;
+
+struct CubeGrid {
```
mmoerchen, line 38: Could you add docstrings to this struct? If this type is intended as an implementation detail rather than part of the public API, it might make sense to place it in an internal/detail namespace instead.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+  explicit CubeGenerator(std::shared_ptr<data::BasisSet> basis_set);
+  ~CubeGenerator() noexcept;
+  CubeGenerator(CubeGenerator&&) noexcept;
+  CubeGenerator& operator=(CubeGenerator&&) noexcept;
+
+  CubeField orbital(const Eigen::VectorXd& mo_coeff, const std::string& outfile,
+                    const CubeGrid& grid,
+                    const std::string& comment = "") const;
+
+  CubeField density(const Eigen::MatrixXd& density_matrix,
```
mmoerchen, line 196: Please add documentation to each public function

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+// The ECP is an operator in the Hamiltonian, so it influences the SCF
+// coefficients but never enters the evaluation of a basis function at a grid
+// point; the valence shells are ordinary contracted Gaussians that are
+// evaluated exactly. The consequence is one of interpretation rather than
+// accuracy: the resulting field is a *valence-only* quantity. Core density is
+// absent near ECP-carrying nuclei, the nuclear cusp there is smoothed, and a
+// density cube integrates to the valence electron count rather than the total.
+// This matches the behaviour of PySCF's cubegen. When the basis reports ECP
+// electrons, the generated cube files record this in their comment line.
+class CubeGenerator {
```
mmoerchen, line 126: Is this class required in the public interface, or is it purely an implementation detail that could be moved to the source file?

If it can be moved to the source file, could we then simplify the implementation by removing the PIMPL pattern altogether?

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+  std::unique_ptr<Impl> _impl;
+};
+
+// Writes one cube file per requested orbital into `output_dir`, returning the
+// paths written. Orbital `indices` are zero-based, matching the numbering used
+// throughout qdk-chemistry, and the emitted file names embed that same
+// zero-based index (`<label_prefix>%04zu`). For restricted wavefunctions a
+// single spatial cube is written per index (no spin suffix); for unrestricted
+// wavefunctions the alpha and beta channels are written as `_a`/`_b` cubes.
+std::vector<std::string> generate_orbital_cubes(
```
mmoerchen, line 229: Please use Doxygen-style comments here to keep the API documentation consistent with the rest of the codebase.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,250 @@
...
+  if (!outfile.empty())
+    GauXC::write_cube(outfile, _impl->gauxc_mol, g, field.data(),
+                      annotate_comment(*_impl->basis_set, comment));
+  return field;
+}
+
+std::vector<std::string> generate_orbital_cubes(
+    const data::Wavefunction& wfn, const std::vector<std::size_t>& indices,
+    const std::string& output_dir, const CubeGrid& grid,
+    const std::string& prefix) {
```
mmoerchen, line 291: Do you think we should sanitize the prefix string to avoid generating filenames like `orbital_.cube0001.cube`?

cpp/include/qdk/chemistry/utils/cube_generator.hpp
```diff
@@ -0,0 +1,76 @@
...
+    const data::Wavefunction&, const std::vector<std::size_t>& indices,
```
mmoerchen: Why does this function require the wavefunction as input? Wouldn't the orbitals alone be sufficient?

cpp/tests/test_cube_generator.cpp
```diff
@@ -0,0 +1,361 @@
...
+using qdk::chemistry::data::OrbitalType;
+using qdk::chemistry::data::Shell;
+using qdk::chemistry::data::StateVectorContainer;
+using qdk::chemistry::data::Structure;
+using qdk::chemistry::data::Wavefunction;
+using qdk::chemistry::utils::CubeGenerator;
+using qdk::chemistry::utils::CubeGrid;
+using qdk::chemistry::utils::generate_orbital_cubes;
+
+std::shared_ptr<BasisSet> make_hydrogen_basis(
```
mmoerchen, line 39: Can we reuse the structures already provided in `ut_common.hpp`, or add these structures there as well?

cpp/tests/test_cube_generator.cpp
```diff
@@ -0,0 +1,361 @@
...
+                static_cast<std::size_t>(std::numeric_limits<int64_t>::max())) {
+    grid.nx = static_cast<std::size_t>(std::numeric_limits<int64_t>::max());
+    grid.ny = 2;
+    EXPECT_THROW(grid.num_points(), std::overflow_error);
+  }
+}
+
+TEST(CubeGeneratorTest, EvaluatesHydrogenOrbitalAndDensity) {
+  CubeGenerator generator(make_hydrogen_basis());
+  CubeGrid grid;
```
mmoerchen: Is there a reason not to use the existing grid helper here? If not, can we remove the helper function instead?

cpp/tests/test_cube_generator.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+  generator.orbital(coefficients, output.string(), single_point_grid());
+
+  const std::string comment = first_line_of(output);
+  EXPECT_NE(comment.find("valence-only"), std::string::npos) << comment;
+  EXPECT_NE(comment.find("28 core electrons"), std::string::npos) << comment;
+  std::filesystem::remove(output);
+}
+
+TEST(CubeGeneratorTest, EcpAnnotationPreservesCallerComment) {
```
mmoerchen, line 330: Can we add a test which checks real ECP behavior like for AgH (it's available in `ut_common.hpp`) in `def2-SVP` or something similar?

mmoerchen (dismissed):

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Documented in `a709532e`, on both `orbital` and `density`, since `mo_coeff` carries the same implicit ordering contract.

Verifying this turned out to be worth doing carefully. `sort_shells_inplace` sorts with a comparator that ignores `atom_index`, so I first suspected the canonical ordering interleaved atoms by angular momentum while `to_gauxc_basis` emits atom-major, which would have been a silent permutation bug rather than a documentation gap. It isn't: there is no separate flat shell vector, and `BasisSet::get_shell` walks `_shells_per_atom` in atom order, so the canonical ordering is atom-major and the two agree. Documented as such, plus a test that pins it to an analytic value so it stays true.

Also recorded that s and p shells are always Cartesian while only `l > 1` honours `AOType`, which matches `convert_to_libint_shell`'s `angular_momentum >= 2` rule, so the convention is consistent across the codebase. A p shell is therefore always three components in x, y, z order.

On spin: the matrix is applied exactly as supplied, with no scaling, symmetrisation, or spin summation, so the caller decides the physical quantity. Pass Da + Db for the total density or a single block for a spin density. I called out the failure mode explicitly, since a restricted spatial density matrix missing its factor of two gives a field that is uniformly half the total density, and because both matrices have identical dimensions no shape check can catch it.

I stopped short of adding a spin-channel enum. It cannot be validated against the matrix contents, so it would imply a guarantee the code cannot make.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Confirmed and fixed in `5c011a8d`.

You're right on both counts. I had guarded against overflowing `int64_t` because that is what `GauXC::CubeGrid` stores `nx/ny/nz` in, but I never checked what gauXC does with the value downstream, and the error message asserted a limit I hadn't verified. The operative constraint is in `check_orbital_args`:

```cpp
// ldo is passed to blas::gemm, whose LDC parameter is int.
if( ldo > static_cast<size_t>(std::numeric_limits<int>::max()) )
  GAUXC_GENERIC_EXCEPTION( ctx + ": ldo exceeds BLAS int range." );
```

and the `CubeGrid` overload passes `ldo = grid.num_points()`, so `INT_MAX` is the real ceiling. The guard was permissive by roughly a factor of four billion.

The allocation ordering point follows from that rather than needing a separate fix: `num_points` is already called from `to_gauxc_grid` before the `CubeField` is constructed, so correcting the constant puts the rejection back ahead of the allocation. Worth noting the old behaviour could be worse than `bad_alloc`, since on Linux with overcommit the vector construction can succeed and the process is then OOM killed during zero initialisation, with no exception to catch.

One thing I did not do, having considered it: I left the two failure cases as a single check rather than separating arithmetic overflow from unsupported size. With the limit at `INT_MAX`, `nx > max / ny` already catches anything that could overflow `size_t`, so a separate overflow branch would be unreachable. Kept `std::overflow_error`, which matches the `factorial` precedent in `algorithms/microsoft/utils.cpp` for a result exceeding a representable range, and put the actual number in the message.

Test pins the boundary: `INT_MAX` points is accepted, `INT_MAX + 1` is refused, and a 4000 cubed grid is refused. That last one is the useful regression, since 6.4e10 points sits far below the old `int64_t` bound and so was previously accepted, then allocated about 512 GB.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Wired the native path rather than narrowing the description.

`baef5c05` adds pybind11 bindings for `CubeGenerator` and `CubeGrid`. `22690b0c` makes native the default in `generate_cubefiles_from_orbitals`; `backend="pyscf"` is kept as an escape hatch and is no longer imported at module scope.

The switch moves no grid point: both backends use origin `min - margin` and step `extent / (n - 1)`, and share AO ordering. `test_native_backend_matches_pyscf_backend` compares origin, axes, shape, and every field value.

Side benefit: PySCF is pinned `sys_platform != 'win32'`, so this module was previously unimportable on Windows. Cube generation now works there. Description updated.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Done in `7bf0f1c3`. The whole header is Doxygen now, matching `logger.hpp` and the rest of `utils/`.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Documented in `7bf0f1c3`: each field, the point layout, and that lengths are in Bohr.

Kept it public rather than moving to a detail namespace. `baef5c05` binds it to Python, so callers construct a `CubeGrid` and hand it to `CubeGenerator`.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Done in `7bf0f1c3`. `@brief`, `@param`, `@return` and `@throws` on each.

I read the `@throws` lists off the implementation rather than guessing, so `from_basis_set` lists `invalid_argument` for a zero dimension or negative margin and `runtime_error` for a missing structure, and `num_points` lists `overflow_error` at the gauXC point limit.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Done in `7bf0f1c3`, converted with the rest of the header.

cpp/tests/test_cube_generator.cpp
ConradJohnston: Added in `55277999`. Every ECP test here was hand-built, so nothing covered a named basis going through `from_basis_name`, which is the same gap that hid the radial-power regression.

AgH in def2-SVP is unambiguous: 48 electrons all-electron, minus the 28-electron Ag ECP leaves 20. The test runs SCF, builds the AO density matrix from the converged occupations, and integrates the density cube. It gets 19.997 on a 60^3 grid with a 6 Bohr margin against a tolerance of 0.05, which is tight enough to catch a dropped factor of two.

The coarse grid is fine because the ECP removes the Ag core cusp, leaving only the hydrogen 1s. I checked 60/100/140 points and margins 6 and 8; the integral moves by less than 0.004.

The second test checks the valence-only comment reports 28 core electrons for a `from_basis_name` basis. Both also pin that projector shells contribute no AOs, since a leak would grow `nbf` and the density matrix would fail the shape check. 1.3s including the SCF.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: Agreed, changed in `71aa1176`. It takes `const data::Orbitals&` now.

The function called `wfn.get_orbitals()` on the first line and never used the wavefunction again, and the wider type shut out anyone holding just an `Orbitals`. It also dereferenced the returned `shared_ptr` with no null check, which a reference makes impossible.

The tests make the point better than I can: both fixtures were building a `Configuration` and a `StateVectorContainer` only to satisfy the signature. That is gone, along with three includes and three usings, and the same in the Python test.

All call sites are new in this PR, so nothing outside it changes. The pybind11 arg name and docstring moved too. 19 C++ and 17 Python tests pass.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Yes, and it turned out worse than the cosmetic case. Fixed in `355ca222`.

`std::filesystem::path::operator/` discards its left operand when the right side is absolute, so a prefix starting with `/` silently ignored `output_dir` and wrote wherever the prefix pointed. A prefix with `..` escaped the directory, and one with a separator targeted a subdirectory `create_directories` never made, so the write failed while the returned path still named a file that did not exist.

A prefix with a path separator or `..` now throws `invalid_argument`. A trailing `.cube` is dropped rather than rejected, since `orbital_.cube` obviously means `orbital_0000.cube`. Case insensitive. Two tests cover both.

While in there I also fixed a comment claiming `cubegen.py` labels files 1-based. That stopped being true earlier in this branch when the Python layer moved to zero-based labels.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,250 @@
...
+      const int l = sh.get_angular_momentum();
+      if (l < 0)
+        throw std::invalid_argument(
+            "CubeGenerator: ECP potential shells are unsupported.");
+      const auto np = static_cast<int32_t>(sh.exponents.size());
+      PA alpha{}, coeff{};
+      if (static_cast<std::size_t>(np) > alpha.size())
+        throw std::invalid_argument(
+            "CubeGenerator: shell exceeds gauXC's primitive limit.");
+      for (int i = 0; i < np; ++i) {
```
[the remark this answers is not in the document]
ConradJohnston: Good catch. Fixed in 28b32552: the check now compares sh.exponents.size() directly and the narrowing happens afterwards.

cpp/tests/test_cube_generator.cpp
ConradJohnston: Done in bc14d004. All six single-hydrogen structures now call testing::create_hydrogen_structure(). The basis sets stay local since they are deliberately minimal test bases rather than anything reusable.

cpp/tests/test_cube_generator.cpp
ConradJohnston: Done in bc14d004, it now uses single_point_grid(). The helper is used in nine other places, so it stays.

cpp/include/qdk/chemistry/utils/cube_generator.hpp
ConradJohnston: CubeGenerator is public API rather than an implementation detail. It is exposed to Python as qdk_chemistry.utils.CubeGenerator, and it is the stateful handle that holds the converted GauXC basis and evaluator so you can evaluate many orbitals on many grids without redoing the conversion each time. generate_orbital_cubes is just the convenience wrapper over it.

The PIMPL is what keeps it publishable. Impl holds GauXC::BasisSet, GauXC::Molecule and GauXC::OrbitalEvaluator by value, so without it the public header would have to include GauXC headers and every consumer of qdk-chemistry would need GauXC on its include path. As it stands the header includes only Eigen/Core, memory, string and vector.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  paths.reserve(indices.size() * (restricted ? 1u : 2u));
+
+  auto emit = [&](const Eigen::VectorXd& coeff, const std::string& filename,
+                  const std::string& comment) {
+    auto path = (std::filesystem::path(output_dir) / filename).string();
+    gen.orbital(coeff, path, grid, comment);
+    paths.push_back(std::move(path));
+  };
+
+  for (auto p : indices) {
```
agamshayit-ms, line 329: Could all requested indices be validated before creating or writing output? With one MO, indices {0, 1} writes orbital_0000.cube and then throws, leaving a partial batch behind.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,290 @@
...
+                                 const std::string& outfile,
+                                 const CubeGrid& grid,
+                                 const std::string& comment) const {
+  if (C.size() != _impl->nbf)
+    throw std::invalid_argument("orbital: mo_coeff length mismatch.");
+  auto g = to_gauxc_grid(grid);
+  CubeField field(g.num_points());
+  _impl->evaluator.eval_orbital(g, C.data(), field.data());
+  if (!outfile.empty())
+    GauXC::write_cube(outfile, _impl->gauxc_mol, g, field.data(),
```
agamshayit-ms, line 267: Could comments reject or strip multiline comments? Passing "first\ninjected" shifts the generated and atom header lines, producing a cube file that standard parsers cannot read.

python/src/qdk_chemistry/utils/cubegen.py
```diff
@@ -49,23 +47,61 @@ def generate_cubefiles_from_orbitals(
...
 
+        backend: Which evaluator to use, either ``"native"`` (default) or ``"pyscf"``.
+
+            The native backend evaluates the orbitals in process with gauXC and needs no
+            third-party quantum chemistry package. The ``"pyscf"`` backend delegates to
+            ``pyscf.tools.cubegen`` and is kept for comparison and for falling back if a
+            difference is ever suspected. PySCF is not installed on Windows, so the
+            ``"pyscf"`` backend is unavailable there and raises ``ImportError``.
+
+            Both backends place grid points identically: the origin is the nuclear bounding
```
agamshayit-ms, line 60: backend="pyscf" does not support Cartesian bases: O/cc-pVDZ converts from 15 qdk-chemistry AOs to 14 PySCF AOs and fails. Could we propagate AOType or reject/document Cartesian use?

cpp/cmake/third_party.cmake
agamshayit-ms: `FORCE_FETCH` fixes fetch-enabled builds. The remaining issue is the explicit no-fetch fallback: it accepts GauXC 1.0, then compilation fails because `cube.hpp` is missing. Could this fallback feature-check the cube API and fail during configure?

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,290 @@
...
+const data::BasisSet& require_basis_set(
+    const std::shared_ptr<data::BasisSet>& basis_set) {
+  if (!basis_set)
+    throw std::invalid_argument("CubeGenerator: basis set cannot be null.");
+  if (!basis_set->get_structure())
+    throw std::invalid_argument("CubeGenerator: basis set has no structure.");
+  return *basis_set;
+}
+
+GauXC::BasisSet<double> to_gauxc_basis(const data::BasisSet& qdk) {
```
agamshayit-ms, line 112: Can this include basis parameters validation similar to `cpp/src/qdk/chemistry/algorithm/microsoft/utils.cpp:49`? Otherwise, bad exponents can be passed onto GauXC

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  };
+
+  for (auto p : indices) {
+    if (std::size_t(C_a.cols()) <= p)
+      throw std::out_of_range("generate_orbital_cubes: index OOB.");
+    const auto stem = stem_prefix + index_string(p);
+    // Restricted: a single spatial cube with no spin suffix. Unrestricted:
+    // separate alpha (`_a`) and beta (`_b`) cubes, mirroring `cubegen.py`.
+    if (restricted) {
+      emit(C_a.col(p), stem + ".cube",
```
agamshayit-ms, line 334: Do we always want restricted orbitals to be labeled `alpha`?

agamshayit-ms (changes requested):

cpp/cmake/third_party.cmake
ConradJohnston: Fixed in `b79e4abd`. `handle_dependency` now accepts a `REQUIRED_HEADER` feature check, and GauXC requires `gauxc/external/cube.hpp`. With fetching disabled, a stale GauXC installation now fails during configure instead of reaching compilation. I also exercised both missing-header and present-header package configurations.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Fixed in `b79e4abd`. All requested indices are validated against both applicable spin blocks before constructing the generator, creating the output directory, or writing any cube. A regression with `{0, 1}` and one available MO verifies that no directory or partial file is left behind.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Fixed in `b79e4abd`. Cube comments containing either `\n` or `\r` are rejected with `invalid_argument` before evaluation or writing. The regression verifies that the malformed request creates no file.

python/src/qdk_chemistry/utils/cubegen.py
ConradJohnston: Fixed in `b79e4abd` by rejecting `backend="pyscf"` for Cartesian bases before importing or converting through PySCF, with guidance to use the native backend. The API documentation and a Cartesian d-shell regression cover the restriction. Simply enabling `mol.cart` would not be sufficient because the Cartesian normalization conventions also differ.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Fixed in `b79e4abd`. Conversion to GauXC now rejects empty shells, mismatched primitive arrays, non-finite or non-positive exponents, non-finite coefficients, and shells beyond GauXC’s primitive limit. Regression coverage includes zero, negative, infinite, and NaN parameters.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Fixed in `b79e4abd`. Restricted spatial orbitals are now labeled `(restricted)` rather than `(alpha)`. Alpha and beta labels remain only on unrestricted output, matching the file suffix behavior.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,316 @@
...
+        if (!std::isfinite(sh.exponents[i]) || sh.exponents[i] <= 0.0)
+          throw std::invalid_argument(
+              "CubeGenerator: shell exponents must be finite and positive.");
+        if (!std::isfinite(sh.coefficients[i]))
+          throw std::invalid_argument(
+              "CubeGenerator: shell coefficients must be finite.");
+        alpha.at(i) = sh.exponents[i];
+        coeff.at(i) = sh.coefficients[i];
+      }
+      basis.emplace_back(GauXC::PrimSize(np), GauXC::AngularMomentum(l),
```
agamshayit-ms, line 158: This theoretically allows 0-coefficient basis functions through to GauXC. We can guard against that, with somethig like
```
if (!std::all_of(basis.back().coeff().begin(), basis.back().coeff().begin() + np, [](double c) { return std::isfinite(c); })) throw std::invalid_argument("CubeGenerator: shell has zero contraction norm.");
```

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/utils/cube_generator.cpp
ConradJohnston: Fixed in `53a19bd2` with a stronger check than `isfinite`, which was already enforced per coefficient. The generator now computes the same normalized contraction norm GauXC computes and rejects non-finite or non-positive results before constructing the GauXC shell. This catches both an all-zero coefficient vector and exact cancellation such as duplicate primitives with coefficients `{1, -1}`. Both cases have regression coverage.

vcpkg-overlay/ports/libaec/portfile.cmake
```diff
@@ -0,0 +1,41 @@
...
+)
+
+set(_target_file "libaec_shared-targets")
+if(BUILD_STATIC)
+    set(_target_file "libaec_static-targets")
+endif()
+file(READ "${CURRENT_PACKAGES_DIR}/share/libaec/${_target_file}.cmake" libaec_targets)
+string(REGEX REPLACE " (SHARED|STATIC) IMPORTED" " \\1 IMPORTED \${libaec_maybe_global}" libaec_targets "${libaec_targets}")
+file(WRITE "${CURRENT_PACKAGES_DIR}/share/libaec/${_target_file}.cmake" "set(libaec_maybe_global \"\")
+if(CMAKE_VERSION VERSION_LESS 3.18)
```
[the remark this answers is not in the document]
ConradJohnston: Fixed in `d072273d`. The targets rewrite now runs only when the generated file does not already start with the `libaec_maybe_global` marker, so a repeated portfile execution leaves both the marker block and imported-target placeholder unchanged. I verified the transformation by applying it twice to a representative targets file and asserting that only one placeholder remains.

cpp/src/qdk/chemistry/utils/cube_generator.cpp
```diff
@@ -0,0 +1,350 @@
...
+// meant.
+std::string sanitize_label_prefix(const std::string& prefix) {
+  if (prefix.find('/') != std::string::npos ||
+      prefix.find('\\') != std::string::npos)
+    throw std::invalid_argument(
+        "generate_orbital_cubes: label prefix cannot contain a path "
+        "separator.");
+  if (prefix.find("..") != std::string::npos)
+    throw std::invalid_argument(
+        "generate_orbital_cubes: label prefix cannot contain '..'.");
```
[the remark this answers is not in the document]
ConradJohnston: Fixed in `1e42b17a`. Prefix validation now uses `std::filesystem::path` decomposition rather than hard-coded separator searches. Any native root path or parent path is rejected, so Windows drive-relative prefixes such as `C:escape_`, UNC/rooted paths, and nested paths cannot replace or escape `output_dir`. The Windows-only regression includes both a drive root-name and a backslash-separated path.

agamshayit-ms (approved):

PR #470, opened 2026-05-06
Add Base Container for Determinant based Wavefunctions

mmoerchen:


cpp/include/qdk/chemistry/data/wavefunction_containers/cc.hpp
```diff
@@ -97,20 +97,32 @@ class CoupledClusterContainer : public WavefunctionContainer {
...
-   * @brief Not implemented for CC wavefunctions
+   * @brief Get CI coefficients generated lazily from CC amplitudes
+   *
+   * Truncates the expansion at fourth order. Result is cached after first
+   * computation.
+   *
+   * @return Reference to vector of CI coefficients
    */
-  const VectorVariant& get_coefficients() const override;
+  const VectorVariant& get_coefficients() const;
```
wavefunction91, line 109: One of the main results of this PR should be that amplitudes should be treated differently than state vector coefficients.

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -588,6 +573,55 @@ class WavefunctionContainer {
...
 
+/**
+ * @brief Base class for wavefunctions represented as a linear expansion in
+ *        Slater determinants
+ *
+ * Guarantees that determinants and their coefficients are stored and directly
+ * accessible. Use this type to dispatch on wavefunctions with an explicit
+ * determinantal representation.
+ */
+class DeterminantalWavefunctionContainer : public WavefunctionContainer {
```
wavefunction91, line 584: 1. This should live in it's own header
2. We're going to generalize this concept in a future PR to take non-determinantal statevectors. I'd prefer that this indicates something like a `StateVectorContainer` or similar.

cpp/include/qdk/chemistry/data/wavefunction_containers/cas.hpp
```diff
@@ -198,29 +198,30 @@ class CasWavefunctionContainer : public WavefunctionContainer {
    */
   void clear_caches() const override;
 
-  /**
-   * @brief Convert container to JSON format
-   * @return JSON object containing container data
-   */
-  nlohmann::json to_json() const override;
-
```
wavefunction91, line 206: Is there a reason to prefer to the `_to_json_impl` design pattern rather than the previous motif?

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -588,6 +573,55 @@ class WavefunctionContainer {
...
+   * @return Vector of all coefficients (real or complex)
+   */
+  virtual const VectorVariant& get_coefficients() const = 0;
+
+  /**
+   * @brief Get coefficient for a specific determinant
+   * @param det Determinant to look up (active space only)
+   * @return Scalar coefficient (real or complex)
+   */
+  virtual ScalarVariant get_coefficient(const Configuration& det) const = 0;
```
wavefunction91, line 606: Things to do with `Configuration` should no longer live in the public `Wavefunction` API as not all instances support it. This is the wrong place to comment on it, but I don't want to page through this file to find the right line.

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -588,6 +573,55 @@ class WavefunctionContainer {
...
+  /**
+   * @brief Get all determinants in the wavefunction
+   * @return Vector of determinants, in the same order as ``get_coefficients``
+   */
+  virtual const DeterminantVector& get_active_determinants() const = 0;
+
+  /**
+   * @brief Get number of determinants in the expansion
+   */
+  virtual size_t size() const = 0;
```
wavefunction91, line 617: Size has to be polymorphic from the top-level API - currently `size` downcasts to the determinant container, which will fail for unnecessary reasons.

python/src/pybind11/data/wavefunction.cpp
```diff
@@ -140,37 +140,27 @@ This enum allows tagging wavefunctions based on their mathematical role:
...
-           "Get orbital occupations for active orbitals")
-      .def("get_active_orbital_occupations",
-           &WavefunctionContainer::get_total_orbital_occupations,
-           "Get orbital occupations for all orbitals")
       .def("get_active_orbital_occupations",
            &WavefunctionContainer::get_active_orbital_occupations,
            "Get orbital occupations for active orbitals only")
+      .def("get_total_orbital_occupations",
+           &WavefunctionContainer::get_total_orbital_occupations,
+           "Get orbital occupations for all orbitals")
```
wavefunction91, line 163: Why change this convention?

cpp/src/qdk/chemistry/data/wavefunction.cpp
```diff
@@ -1788,11 +1849,59 @@ void WavefunctionContainer::to_hdf5(H5::Group& group) const {
...
   } catch (const H5::Exception& e) {
     throw std::runtime_error("HDF5 error: " + std::string(e.getCDetailMsg()));
   }
 }
 
+void DeterminantalWavefunctionContainer::_to_hdf5_impl(H5::Group& group) const {
+  QDK_LOG_TRACE_ENTERING();
+
+  // Store restrictedness flag
+  bool is_restricted = get_orbitals()->is_restricted();
```
wavefunction91, line 1864: This should be guarded on a null check for the orbitals instance

wavefunction91 (changes requested):

python/src/pybind11/data/wavefunction.cpp
mmoerchen: `get_active_orbital_occupations` was defined twice (with active and total occupations), and `get_orbital_occupations` didn't follow the `active/total` pattern we usually use

cpp/include/qdk/chemistry/data/wavefunction.hpp
mmoerchen: Okay, but what should `size()` return for a non-determinantal wavefunction?
I move the size method to the `StateVectorContainer`, which is why I downcast.

cpp/include/qdk/chemistry/data/wavefunction_containers/cas.hpp
mmoerchen: This way, each class only has to handle the serialization for its specific data structures instead of implementing the logic of common data (like RDMs, entropies, etc.) as well.

cpp/include/qdk/chemistry/data/wavefunction_containers/cc.hpp
mmoerchen: Yes, CC and MP2 now only inherit from the base container and aren't required to provide these methods. They do have CI coefficients implemented, so I kept it.

Should we nix these functions to keep the API cleaner, or rename it to make it more explicit that it involves computation (e.g., compute_ci_expansion())?

mmoerchen:
See #523

PR #479, opened 2026-05-13
Add higher-order spin Hamiltonians for lattice models

wavefunction91:


eimrek:
hi @wavefunction91, I took a look, looks good. Some preliminary high level comments:

* I think it would be useful to have a general spin Hamiltonian creation method, that allows for any arbitrary terms (bilinear, such as the Dzyaloshinskii-Moriya term, biquadratic, ring exchange, ...). And then Heisenberg and Ising could be wrappers around this.
* Also it would be good to be able to set Nth nearest neighbor interactions in a convenient way.
* I wonder if perhaps it would be good to decouple the spin hamiltonian definition, and the conversion to a `QubitHamiltonian`. Related to this, I am wondering if perhaps these spin hamiltonian creation methods should just return a `Hamiltonian` object (that could be abstracted to support both fermionic and spin degrees of freedom, if it doesn't). This is also relevant for supporting hamiltonians that mix spin and fermionic DoFs (like t-J or Kondo).

Happy to discuss, and also to contribute to this PR. I'll think more about these points and come up with some more concrete proposals.

Edit: but of course these comments are a bit out of scope of this PR. Also happy to get this done, and then discuss a potential higher-level refactor in a separate PR. let me know.

PR #489, opened 2026-05-22
Small dev container refactor

### LINKED ISSUE #485: Dev container suggestions
Some suggestions for the dev container setup (happy to prep a PR if agreed):

## `venv.sh` static tasks

As `venv.sh` is ran every time a container is created, static (and potentially slow) tasks are better to be moved to the Dockerfile, these include:

- setting up the venv
- Installing pip, node, rust, ipykernel, ...

## Building C++ separately from python

At the moment

```bash
QDK_UARCH=native pip install -v .[all]
```
builds C++ and python bindings together.

But it might be nicer already here to install C++ separately and link the python build to it, which should allow for faster development.

(At the moment I'm basically doing this manually anyway and ignoring the current install step.)

eimrek:
fixes #485

1st commit addresses point 1 (moving static installs to the image build).

For point 2, I set up the separate C++ and python installation that seems to work well for me. But I do understand that different users might want to set up their environment in different ways, so I could also revert this change (and use the single `pip install` command).

`devcontainer-lock.json` also seems to be committed conventionally, added that as well.


eimrek:

wavefunction91 (approved):

.devcontainer/scripts/post_create.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# Post-create step: install QDK Chemistry from the mounted source.
+set -euo pipefail
+source "$HOME/qdk_chemistry_venv/bin/activate"
+export CMAKE_BUILD_PARALLEL_LEVEL=4
+
+# Build C++ and install to /usr/local
+cmake -S cpp -B cpp/build -G Ninja
+cmake --build cpp/build
+sudo cmake --install cpp/build
```
wavefunction91: Just noticed this - this will install to a global location. This will preclude local builds of the code

wavefunction91 (dismissed):

.devcontainer/scripts/post_create.sh
eimrek: thanks for the note. switched to a non-sudo install.

eimrek:
@wavefunction91 sorry for the delay, testing the full build took a while.

Apart from the non-sudo c++ qdk-chemistry install, few other small changes:

I added the `CMAKE_BUILD_PARALLEL_LEVEL=4` already in the Dockerfile as I was often forgetting to set it before running `pip install` and getting OOM. users can always override this in their local env.

I also set `-DCMAKE_DISABLE_FIND_PACKAGE_macis=ON` for the first install, which will prevent a full rebuild on further local installs.

`ccache` seems to help with some optimizations, and the python `CMakeList` makes use of it, so I installed it in the image.

.devcontainer/scripts/post_create.sh
```diff
@@ -0,0 +1,17 @@
...
+# full rebuild on reconfigure.
+cmake -S cpp -B cpp/build -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="$HOME/.local" \
+    -DCMAKE_DISABLE_FIND_PACKAGE_macis=ON
+cmake --build cpp/build
+cmake --install cpp/build
+
+# Install python library
+cd ./python
+pip install -v .[all]
```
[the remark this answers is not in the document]
eimrek: Not correct. cmake reads the env variable since 3.22 (https://cmake.org/cmake/help/v3.22/variable/CMAKE_BUILD_TYPE.html)

nabbelbabbel:
> @wavefunction91 sorry for the delay, testing the full build took a while.
> 
> Apart from the non-sudo c++ qdk-chemistry install, few other small changes:
> 
> I added the `CMAKE_BUILD_PARALLEL_LEVEL=4` already in the Dockerfile as I was often forgetting to set it before running `pip install` and getting OOM. users can always override this in their local env.
> 
> I also set `-DCMAKE_DISABLE_FIND_PACKAGE_macis=ON` for the first install, which will prevent a full rebuild on further local installs.
> 
> `ccache` seems to help with some optimizations, and the python `CMakeList` makes use of it, so I installed it in the image.

`CMAKE_BUILD_PARALLEL_LEVEL=4` should not be hardcoded anywhere. This will break for anyone using less then 4 cores or just generally smaller machines. If we decide to cap it somewhere then we need a working check of number of true cores and only apply `min(4, <number of cores>)`.  Personally, I think we should just not hardcode this. 

As memory requirements on compile are an issue in practice, I'd rather remove the parallelization and have that require an explicit request if we want to ensure maximum stability.

eimrek:
thanks @nabbelbabbel for the feedback

Ok, sure, happy to revert the `CMAKE_BUILD_PARALLEL_LEVEL=4`. But I think it's important to have a default value, either 1  or `min(4, <number of cores>)` (as you suggest) or something else. Defaulting to all cores causes an inevitable OOM/crash.

But if we do set the value in the docker image, users can always override this in their container.

Also note that for installing the cpp dependencies, i think we hardcode 4 cores for libint: https://github.com/microsoft/qdk-chemistry/blob/ba51df713242f91322c201e89712ec22a5e4eab5/.devcontainer/scripts/install_cpp_dependencies.sh#L40

mmoerchen:
> thanks @nabbelbabbel for the feedback
> 
> Ok, sure, happy to revert the `CMAKE_BUILD_PARALLEL_LEVEL=4`. But I think it's important to have a default value, either 1 or `min(4, <number of cores>)` (as you suggest) or something else. Defaulting to all cores causes an inevitable OOM/crash.
> 
> But if we do set the value in the docker image, users can always override this in their container.
> 
> Also note that for installing the cpp dependencies, i think we hardcode 4 cores for libint:
> 
> https://github.com/microsoft/qdk-chemistry/blob/ba51df713242f91322c201e89712ec22a5e4eab5/.devcontainer/scripts/install_cpp_dependencies.sh#L40

`CMAKE_BUILD_PARALLEL_LEVEL=2` is also exported in `.devcontainer/scripts/venv.sh` to prevent OOM on smaller machines.

eimrek:
@wavefunction91 @nabbelbabbel @mmoerchen 

Took another stab at this:

Now there is a helper script/function `parallel_jobs_for_memory <memory_per_job_gb>` that is now used in `install_cpp_dependencies.sh` and also sets a sensible `CMAKE_BUILD_PARALLEL_LEVEL` for the `post_create.sh` command also interactive shell env (e.g. in vscode terminal). Users can, of course, override this value.

.devcontainer/Dockerfile
lorisercole, line 1: Is Microsoft's Container Registry accessible by standard users? If not, we may want to point to the Docker Hub image by default.

INSTALL.md
```diff
@@ -125,6 +125,10 @@ Alternatively, click the green button in the bottom-left corner of VS Code and s
 
 After the initial build, restart VS Code and reopen in the container to ensure the Python virtual environment is properly loaded.
 
+### Step 4: Develop
+
+The dev container builds the C++ library and links it to the Python package in separate steps, as shown in `.devcontainer/scripts/post_create.sh`. After changing the source code, you need to rebuild and install accordingly.
```
lorisercole: Does "rebuild and install accordingly" mean running the `post_create.sh` script again from within the container? I'd suggest it clearly (and perhaps we can call this script `build.sh`, if it makes it clearer).

.devcontainer/Dockerfile
```diff
@@ -89,3 +95,23 @@ USER $USERNAME
...
+ENV VIRTUAL_ENV=/home/$USERNAME/qdk_chemistry_venv \
+    PATH=/home/$USERNAME/qdk_chemistry_venv/bin:$PATH
+
+RUN python -m venv "$VIRTUAL_ENV" \
+    && pip install --no-cache-dir --upgrade pip setuptools wheel \
+    && pip install --no-cache-dir \
+    "ipykernel>=6.0" \
+    "pandas>=2.0.0" \
+    "pre-commit>=4.6,<5" \
+    "rdkit" \
```
lorisercole: This list needs to be updated. See the updated [pyproject.toml](https://github.com/microsoft/qdk-chemistry/blob/main/python/pyproject.toml)

.devcontainer/devcontainer.json
```diff
@@ -55,16 +55,26 @@
...
+        },
+        "ghcr.io/devcontainers/features/rust:1": {
+            "version": "stable",
+            "profile": "default"
+        }
     },
     "remoteEnv": {
-        "CPATH": "${containerEnv:CPATH}:/usr/include/hdf5/serial:/usr/include/eigen3"
+        "CPATH": "${containerEnv:CPATH}:/usr/include/hdf5/serial:/usr/include/eigen3",
+        "CMAKE_BUILD_TYPE": "Debug"
```
lorisercole: For the sake of clarify, coud we move these to the Dockerfile? It's very confusing to have env variables definitions scattered in different places. Or is there a reason why we should leave them here?

.devcontainer/scripts/post_create.sh
```diff
@@ -0,0 +1,21 @@
...
+# Set a memory-/core-aware CMAKE_BUILD_PARALLEL_LEVEL (unless already set) so the
+# initial build below does not oversubscribe CPU or OOM on constrained machines.
+source /usr/local/share/qdk/parallelism.sh
+
+# Build C++ and install to a user-local prefix.
+# Use the in-tree macis (external/macis) per INSTALL.md; prevents a
+# full rebuild on reconfigure.
+cmake -S cpp -B cpp/build -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="$HOME/.local" \
+    -DCMAKE_DISABLE_FIND_PACKAGE_macis=ON
```
lorisercole, line 17:
```suggestion
    -DCMAKE_DISABLE_FIND_PACKAGE_macis=ON \
    -DQDK_UARCH="native" \
    -DBUILD_TESTING=ON
```

.devcontainer/scripts/post_create.sh
```diff
@@ -0,0 +1,21 @@
...
+cmake --build cpp/build
+cmake --install cpp/build
+
+# Install python library
+cd ./python
+pip install -v .[all]
```
lorisercole: would it make sense to enable persistent build caching with `-C build-dir="build/{wheel_tag}"` to speed up a rebuild, or do we just risk having more issues with stale CMake cache?

lorisercole (changes requested):

.devcontainer/Dockerfile
eimrek: Yes, the MCR image is publicly accessible.

INSTALL.md
eimrek: In principle the `post_create.sh` contains commands/options that don't really need to be called exactly during the dev loop. e.g.
```
source /usr/local/share/qdk/parallelism.sh
...
-DCMAKE_DISABLE_FIND_PACKAGE_macis=ON
```

That's why I tried to word it in a way that users should take inspiration from `post_create.sh` but not necessarily copy exactly line-by-line.

I'm not against calling the script `build.sh`, maybe cleaner, yes.

.devcontainer/Dockerfile
eimrek: I think the currently listed python deps match with `pyproject.toml`, no?

In principle, these dependencies will get installed either way with `pip install` but I guess the idea was that if they're already baked in the image, it will speed up the initial container creation. But could also be removed, to keep the Dockerfile cleaner. let me know what you think.

.devcontainer/devcontainer.json
eimrek: In principle, in some cases it might make sense to be able to set env variables only in the container but not in the image itself.

`CPATH` was one of these, and i kept it, as there probably was a reason (although I think it should be fine baked in the image).

I added the `"CMAKE_BUILD_TYPE": "Debug"` variable here as well, with the loose reasoning being that one might want to use the Docker image independently from the dev container, potentially for a release version.

.devcontainer/scripts/post_create.sh
eimrek: Here i think these 2 lines are not necessary:
- `QDK_UARCH="native"` is already set in the Dockerfile and read automatically by cmake
- `BUILD_TESTING=ON` - already the default

.devcontainer/scripts/post_create.sh
eimrek: Perhaps. but i didn't have time to test if this works robustly (e.g. due to pybind).

.devcontainer/Dockerfile
lorisercole: Yes, but if this is the purpose (and I think it's worth it) then in principle we should be installing almost everything but qdk-chemistry... and I think it's fine :)
I mean, I don't see why we should install `rdkit` but not `pytest`, and other common plugin dependencies like `pyscf`
https://github.com/microsoft/qdk-chemistry/blob/8fa60244a59d76b8f05ec6b37ac7c93535587ea6/python/pyproject.toml#L48-L112
If we want to speed up the installation of python packages, maybe we should avoid plain pip, but use something like conda (that uses mamba) or uv

INSTALL.md
lorisercole: Yes, I understand, but perhaps we could just make the build task simpler to the user: provide a callable script just to rebuild. I always take a lot of time to find the correct list of cmake commands to build correctly 😅
but we can ask for @wavefunction91 's opinion

.devcontainer/devcontainer.json
lorisercole: I see, so in practice this is equivalent to defining these variables in `post_create.sh` then.
I don't have a strong opinion, I just find it confusing to have env variables defined in multiple places. It took me a while to realize if/where they were defined 😆

eimrek: > so in practice this is equivalent to defining these variables in post_create.sh then

Not really, as `post_create.sh` gets only called once when the container is created. The variables you set there (e.g. via `export`) are not retained for user sessions (e.g. the terminal).

INSTALL.md
eimrek: The commands to rebuild also depend on what you changed (e.g. python-only changes; do you need to rerun configure; ...). So at the moment I didn't see a good way to document this.

.devcontainer/Dockerfile
eimrek: Ok so i removed these dependencies from the Dockerfile altogether, just to keep it clean. They will all get built by `post_create.sh` either way, and the order shouldn't practically affect the total build time. Now there is one clean source of truth: just the `pyproject.toml`.

Regarding `uv` or `mamba`, i think these would be great improvements but should probably go in a separate PR.

.devcontainer/devcontainer.json
eimrek: ended up moving `CMAKE_BUILD_TYPE` to the Dockerfile

eimrek:
Ok, I updated and cleaned this up:

- Set Microsoft’s public npm and PyPI package proxies as the image defaults. This allows builds to work by default on Microsoft networks; external users can use the public proxies as well or override the build arguments to select different registries.
- Improved incremental build performance with editable Python installs, persistent CMake state, and `ccache`. Pure Python changes are now visible immediately; in testing, no-op C++ builds took ~0.05s, core C++ rebuilds ~10s, and native Python rebuilds ~50–60s.
- I also fixed this https://github.com/microsoft/qdk-chemistry/pull/634#pullrequestreview-4950546375, as my AI was flagging it. let me know if it's better to fix that separately.

@lorisercole @nabbelbabbel, could you take another look?

.devcontainer/devcontainer-lock.json
lorisercole, line 1: If you could just check if we can use the latest versions of these, it might be better for security reasons

lorisercole (dismissed):
Thanks, looks good to me! :) Just a minor comment if you can check.

.devcontainer/devcontainer-lock.json
eimrek: thanks! updated.

lorisercole (approved):

.devcontainer/scripts/install_cpp_dependencies.sh
```diff
@@ -37,13 +37,30 @@ BUILD_DIR="${BUILD_DIR:-/tmp/qdk_deps_build}"
...
+fi
+
+if command -v parallel_jobs_for_memory >/dev/null 2>&1; then
+    # Use the parallelism helper to determine job counts based on memory
+    JOBS="${JOBS:-$(parallel_jobs_for_memory 1)}"
+    LIBINT_JOBS="${LIBINT_JOBS:-$(parallel_jobs_for_memory 4)}"
 else
-    JOBS=$(sysctl -n hw.logicalcpu) # macOS
+    # Fallback when helper is unavailable
+    if command -v nproc >/dev/null 2>&1; then
```
nabbelbabbel, line 55: Is there a fallback for windows?

eimrek: I think the equivalent Windows installation is handled by this script:
https://github.com/microsoft/qdk-chemistry/blob/main/.pipelines/install-scripts/install-cpp-deps-windows.ps1

This current bash script only supports Linux and macOS, and I would avoid adding a Windows-specific fallback here (as it would fail for windows either way).

@lorisercole, feel free to comment if you disagree.

nabbelbabbel (approved):

.devcontainer/scripts/install_cpp_dependencies.sh
lorisercole: yeah, not sure if devcontainers are/can be supported on Windows anyway, or how hard it would be

PR #492, opened 2026-05-26
Add CodeQL workflow for code analysis

sobolevnrm:


wavefunction91:
## Why `autobuild` fails for `c-cpp`

From the failing job log:

> cpp/autobuilder: At least these two subdirs have build system files in them:
>   ./cpp
>   ./python
> cpp/autobuilder: To avoid ambiguity, build will be attempted from '.'
> ...
> cpp/autobuilder: No supported build system detected.

CodeQL's C++ autobuilder walks the repo looking for a single build root. Because we keep both the C++ library (`cpp/CMakeLists.txt`) and the Python extension build (`python/pyproject.toml` + `python/CMakeLists.txt`) at the repo root, the autobuilder gives up and tries `.`, which has no build system, and bails.

There is no `autobuild` configuration that resolves this — the autobuilder doesn't accept a working-directory hint. The supported workaround is to switch `c-cpp` to `build-mode: manual` and run the build ourselves.

## How the release pipeline gets CodeQL for free

`.pipelines/python-wheels.yaml` extends `v1/1ES.Official.PipelineTemplate.yml` with:

```yaml
sdl:
  codeql:
    compiled:
      enabled: true
    sourceLanguages: python
    binaryLanguages: cpp
```

The 1ES template instruments the wheel build that already runs in that pipeline, so CodeQL observes the real compilation and never invokes autobuild. We want the same shape in CI: drive CodeQL off a build we already run, not off `autobuild`.

## Recommended approach — integrate as a stage of `build-and-test.yaml`

Rather than maintain a parallel C++ build in `codeql.yml`, add a CodeQL job to `build-and-test.yaml` and reuse the install/configure/build steps that the `coverage` job already exercises (system deps install → `install_cpp_dependencies.sh` with the `cgmanifest.json` cache → `cmake -S cpp -B build` → `cmake --build`). Sketch:

```yaml
codeql:
  name: CodeQL Analyze (${{ matrix.language }})
  needs: check-up-to-date
  runs-on: ${{ matrix.runner }}
  permissions:
    security-events: write
    packages: read
    actions: read
    contents: read
  strategy:
    fail-fast: false
    matrix:
      include:
        - language: actions
          build-mode: none
          runner: ubuntu-latest
        - language: python
          build-mode: none
          runner: ubuntu-latest
        - language: c-cpp
          build-mode: manual
          runner: [self-hosted, 1ES.Pool=1es-gh-hb120-pool-wus3, 1ES.ImageOverride=Ubuntu24.04Nightly, 'JobId=qdk_chemistry-codeql-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}']
  steps:
    - uses: actions/checkout@v6
      with:
        submodules: recursive
    - uses: github/codeql-action/init@v4
      with:
        languages: ${{ matrix.language }}
        build-mode: ${{ matrix.build-mode }}
    # For c-cpp only: install system deps, restore dep cache, configure & build
    # (same steps as the `coverage` job, minus the coverage flags and tests).
    - if: matrix.language == 'c-cpp'
      ...same install + cache + cmake configure + cmake --build...
    - uses: github/codeql-action/analyze@v4
      with:
        category: "/language:${{ matrix.language }}"
```

Notes:

1. **Build mode `manual` for `c-cpp` only.** `python` and `actions` stay on `build-mode: none` — they don't need a compiler invocation, just source extraction.
2. **Drop the `swift`/`make bootstrap` boilerplate** from the template; we don't need it.
3. **Don't reuse the `coverage` matrix entries directly.** That matrix builds with `-DQDK_CHEMISTRY_ENABLE_COVERAGE=ON` across three Python versions and macOS — CodeQL only needs one Linux build, with a plain `RelWithDebInfo` (or `Debug`) build. Adding CodeQL to every coverage matrix entry would multiply runtime for no extra coverage value.
4. **Cache reuse.** The new job can read the same `cpp-deps-${{ runner.os }}-${{ matrix.uarch }}-...` cache the `coverage` job populates, so on warm cache it skips the long dependency build.
5. **`schedule:` trigger.** PR #492 keeps the weekly cron — worth keeping so CodeQL still runs on `main` even if no PRs land. We can either keep that on `codeql.yml` (deleting the c-cpp matrix entry there) or add `schedule:` to `build-and-test.yaml`. I'd lean toward the latter so there's one workflow to reason about.
6. **`actions` analysis is new.** That entry will scan the workflow files themselves — useful, basically free, no build needed.

## Alternative shapes (if the above is too much for one workflow)

- **Keep `codeql.yml` as a separate workflow** but switch `c-cpp` to `build-mode: manual` and inline the install/build steps. Same fix, different file. Loses the cache-sharing benefit with `coverage`.
- **Extract the install/configure/build into a composite action** (`.github/actions/build-cpp/`) and call it from both `coverage` and `codeql`. Cleanest long-term, but a bigger refactor than this PR needs.

Happy to push the recommended version to this branch if that direction sounds good.

sobolevnrm:
@wavefunction91 I'm going to close this PR.  The functionality is covered in other workflows and I don't have time to debug this PR or address the issues GHCP raised.

PR #493, opened 2026-05-26
Support qdk 1.29 API migrations

RushiGong:


docs/source/conf.py
```diff
@@ -207,6 +207,7 @@
     (r"py:class", r"qdk::chemistry::algorithms::HamiltonianConstructor"),
     (r"py:class", r"^SumPauliOperatorExpression$"),
     (r"py:class", r"qsharp\..*"),  # qsharp has no intersphinx inventory
+    (r"py:class", r"qdk\..*"),  # qdk has no intersphinx inventory
```
wavefunction91, line 210: Does this also preclude `qdk_chemistry` things?

wavefunction91 (changes requested):
Can you also check that this works with the current minimum required version for qdk?

docs/source/conf.py
RushiGong: No, there is `"\."` restriction to include the literal dot character, ensure it matches `qdk.<something>`, not `qdk_chemistry` or anything else start with `qdk`.

RushiGong:
Local test with the current minimum version of qdk==1.27.0:
- pytest
image

- doc build
image

wavefunction91 (approved):

PR #496, opened 2026-05-27
print available options for factory settings

### LINKED ISSUE #473: print_settings() does not print available string options except default
**Describe the bug**
When running `print_settings` with many methods, many settings do not print allowed strings, for example
`print_settings("orbital_localizer", "pyscf_multi")` tells you the default is `pipek-mezey`, but does not advice `Edmiston-Ruedenberg` or `Foster-Boys`, `print_settings("scf_solver", "qdk")` tells you the default for `method` is `hf`, but does not mention eg. `B3LYP`.  Similar problems exist throughout the software, though not for all setting keys. (Eg . `print_settings("state_prep", "sparse_isometry_gf2x")` is fine.)

**To Reproduce**

1. Install qdk-chemistry via `pip install 'qdk-chemistry[plugins, qiskit-extras, jupyter]'`
2. from qdk_chemistry.algorithms import print_settings
3. print_settings("orbital_localizer", "pyscf_multi")

**Expected behavior**
I expect Value to be the default, and Allowed to show other options, in this example, `Edmiston-Ruedenberg` and `Foster-Boys` which do exist for "orbital_localizer", "pyscf_multi".

**Screenshots**

|Key                       | Value           | Allowed              | Description |
|-------------------|--------------|-------------------|--------------|
|method                    | "pipek-mezey"   | -                    | -|
|occupation_threshold      | 1.00e-10        | -                    | -|
|population_method         | "mulliken"      | -                    | -|

**System (please complete the following information):**

- OS: WSL2 on Win64
- System: x64
- Code Version: qdk-chemitry 1.1.0

rainli323:
As issue #473 says, print_settings(factory, method) does not always print all available options. This PR fixes the issue. In cases where HF is the default, and DFTs are all available, we do not give the complete list of DFTs in the result table, instead, we mention in the comment that the user should refer to the manual. I've also put it a "rule": if a future setting expects a string as value, the setting must have a list of available options or a description for user reference, otherwise a test will fail. 

This PR also removes the scf_type setting from hamiltonian factories (4 center and cholesky). The new design matches with the "auto" setting, which mean if 1) the Orbitals is restricted and 2) active spaces are the same between alpha and beta.

### DIFF
--- a/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
+++ b/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
@@ -53,7 +53,11 @@ Eigen::MatrixXd build_K_from_cholesky(
 class CholeskyHamiltonianSettings : public qdk::chemistry::data::Settings {
  public:
   CholeskyHamiltonianSettings() {
-    set_default("scf_type", "auto");
+    set_default("scf_type", std::string("auto"),
+                "SCF orbital type: 'auto' selects based on spin multiplicity",
+                data::ListConstraint<std::string>{
+                    {std::vector<std::string>{"auto", "restricted",
+                                              "unrestricted"}}});
     set_default("cholesky_tolerance", 1e-8);
     set_default("eri_threshold", 1e-12,
                 "ERI screening threshold for skipping negligible shell "
--- a/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
+++ b/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
@@ -53,10 +53,6 @@ Eigen::MatrixXd build_K_from_cholesky(
 class CholeskyHamiltonianSettings : public qdk::chemistry::data::Settings {
  public:
   CholeskyHamiltonianSettings() {
-    set_default("scf_type", std::string("auto"),
-                "SCF orbital type: 'auto' selects based on spin multiplicity",
-                data::ListConstraint<std::string>{{std::vector<std::string>{
-                    "auto", "restricted", "unrestricted"}}});
     set_default("cholesky_tolerance", 1e-8);
     set_default("eri_threshold", 1e-12,
                 "ERI screening threshold for skipping negligible shell "
--- a/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
+++ b/cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
@@ -55,9 +55,8 @@ class CholeskyHamiltonianSettings : public qdk::chemistry::data::Settings {
   CholeskyHamiltonianSettings() {
     set_default("scf_type", std::string("auto"),
                 "SCF orbital type: 'auto' selects based on spin multiplicity",
-                data::ListConstraint<std::string>{
-                    {std::vector<std::string>{"auto", "restricted",
-                                              "unrestricted"}}});
+                data::ListConstraint<std::string>{{std::vector<std::string>{
+                    "auto", "restricted", "unrestricted"}}});
     set_default("cholesky_tolerance", 1e-8);
     set_default("eri_threshold", 1e-12,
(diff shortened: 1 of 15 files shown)


cpp/src/qdk/chemistry/algorithms/microsoft/localization/vvhv.hpp
```diff
@@ -21,7 +21,10 @@ namespace qdk::chemistry::algorithms::microsoft {
 class VVHVLocalizerSettings : public IterativeOrbitalLocalizationSettings {
  public:
   VVHVLocalizerSettings() {
-    set_default("minimal_basis", std::string("sto-3g"));
+    set_default("minimal_basis", std::string("sto-3g"),
+                "Minimal basis set for VVHV partitioning",
+                data::ListConstraint<std::string>{
+                    {std::vector<std::string>{"sto-3g", "sto-3g*"}}});
```
wavefunction91, line 27: Why is this a list constraint limited to just these two entities?

python/src/qdk_chemistry/plugins/qiskit/circuit_executor.py
```diff
@@ -52,7 +52,23 @@ def __init__(self) -> None:
         Logger.trace_entering()
         super().__init__()
         self._set_default("seed", "int", 42)
-        self._set_default("method", "string", "statevector")
+        self._set_default(
+            "method",
+            "string",
+            "statevector",
+            "Aer simulation method",
+            [
```
RushiGong, line 60: Since qiskit is integrated as plugin, it would be helpful to add tests coverage to verify compatibility across all these supported methods.

cpp/src/qdk/chemistry/algorithms/microsoft/cholesky_hamiltonian.hpp
```diff
@@ -53,7 +53,10 @@ Eigen::MatrixXd build_K_from_cholesky(
 class CholeskyHamiltonianSettings : public qdk::chemistry::data::Settings {
  public:
   CholeskyHamiltonianSettings() {
-    set_default("scf_type", "auto");
+    set_default("scf_type", std::string("auto"),
+                "SCF orbital type: 'auto' selects based on spin multiplicity",
+                data::ListConstraint<std::string>{{std::vector<std::string>{
+                    "auto", "restricted", "unrestricted"}}});
```
wavefunction91: From what I can tell, this is actually a bug - the contract is that the HamiltonianConstructor should produce the representation of a given Orbitals instance - so the nature of the rep should be completely determined by the input Orbitals themselves.

cpp/src/qdk/chemistry/algorithms/microsoft/hamiltonian.hpp
```diff
@@ -6,14 +6,22 @@
...
-    set_default("scf_type", "auto");
+    set_default("eri_method", std::string("direct"),
+                "ERI evaluation method: 'direct' computes integrals "
+                "on-the-fly, 'incore' stores all integrals in memory",
+                data::ListConstraint<std::string>{
+                    {std::vector<std::string>{"direct", "incore"}}});
+    set_default("scf_type", std::string("auto"),
+                "SCF orbital type: 'auto' selects based on spin multiplicity",
+                data::ListConstraint<std::string>{{std::vector<std::string>{
+                    "auto", "restricted", "unrestricted"}}});
```
wavefunction91: See other comment

wavefunction91 (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/localization/vvhv.hpp
rainli323: Because the method somehow requires it. see cpp/src/qdk/chemistry/algorithms/microsoft/localization/vvhv.cpp:340

python/src/qdk_chemistry/plugins/qiskit/circuit_executor.py
rainli323: Added a test that only sees if the create(factory, method).settings().set(key, value) works, but it's not a run time test to see if the plugin has changed, which is going to be difficult.

wavefunction91 (approved):

PR #497, opened 2026-05-30
Add SymmetryBlockedTensor

### LINKED ISSUE #495: HDF5 Hamiltonian generation and loading errors
A Hamiltonian saved to HDF5 on one environment cannot be loaded in another using Hamiltonian.from_hdf5_file. The load fails with:
```
ValueError: Tried to generate invalid Hamiltonian object.
```
### Expected behavior
The Hamiltonian should load successfully across environments when using Hamiltonian.from_hdf5_file.

### To Reproduce

1. Generate hdf5 file in one environment (**envA**) using `qdk_chemistry` version `2.0.0` (branch: `feature/agamshayit/x2c0` ): 
```
from qdk_chemistry.data import Structure
from qdk_chemistry.utils import compute_valence_space_parameters
xyz = """3
Water molecule
O    0.000000    0.000000    0.000000
H    0.758602    0.000000    0.504284
H   -0.758602    0.000000    0.504284
"""
structure = Structure.from_xyz(xyz)
scf = create("scf_solver")
scf.settings().set("scf_type", "restricted")
e_scf, scf_wfn = scf.run(
    structure,
    charge=0,
    spin_multiplicity=1,
    basis_or_guess="def2-svp",
)
num_val_e, num_val_o = compute_valence_space_parameters(scf_wfn, charge=0)
selector = create("active_space_selector", "qdk_valence")
selector.settings().set("num_active_electrons", num_val_e)
selector.settings().set("num_active_orbitals", num_val_o)
active_wfn = selector.run(scf_wfn)
ham_constructor = create("hamiltonian_constructor", "qdk_cholesky")
ham_constructor.settings().set("cholesky_tolerance", 1e-6)
ham = ham_constructor.run(active_wfn.get_orbitals())
ham.to_hdf5_file("water.hamiltonian.h5")
print(f"SCF energy: {e_scf:.10f} Ha")
print(f"Valence active space: {num_val_e}e, {num_val_o}o")
print(f"Saved Hamiltonian: water.hamiltonian.h5")
```
2. Load hdf5 file in new environment (**envB**) where `qdk_chemistry` version `1.1.0` is used:
```
>>> from qdk_chemistry.data import Hamiltonian
>>> h = Hamiltonian.from_hdf5_file("water.hamiltonian.h5")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    h = Hamiltonian.from_hdf5_file("water.hamiltonian.h5")
ValueError: Tried to generate invalid Hamiltonian object.
```
#### Environment A (generation)

| Field          | Value |
|----------------|-------|
| Machine        | Intel Xeon Platinum 8180M |
| OS             | Ubuntu |
| Python         | 3.12.3 |
| qdk_chemistry  | 2.0.0 (`feature/agamshayit/x2c0`) |

#### Environment B (loading)

| Field          | Value |
|----------------|-------|
| Machine        | AMD EPYC 7V12 |
| OS             | Azure Linux 3.0 |
| Python         | 3.13.1 |
| qdk_chemistry  | 1.0.0 |

### Additional Notes: 
- This error is not observed when generating the Hamiltonian via code provided in step 1 using **envB**, then reading on **envA** following the code provided in step 3
- hdf5 files generated in **envA** using `qdk_chemistry` version `2.0.0` can be properly loaded in another virtual environment on **envA** using `qdk_chemistry` version `1.1.0`
 

Perhaps this is a python version discrepancy?

wavefunction91:


cpp/include/qdk/chemistry/data/basis_set.hpp
```diff
@@ -523,6 +550,27 @@ class BasisSet : public DataClass,
...
+   *
+   * @return Shared pointer to the AO @ref Symmetries
+   */
+  std::shared_ptr<const Symmetries> ao_symmetries() const;
+
+  /**
+   * @brief Per-label extents of the atomic-orbital basis
+   *
+   * Maps each admissible @ref SymmetryLabel to the number of atomic orbitals it
+   * carries. With the default restricted spin axis, both @f$\alpha@f$ and
```
wavefunction91: No need to belabor the default behavior here

cpp/include/qdk/chemistry/data/basis_set.hpp
```diff
@@ -842,6 +896,19 @@ class BasisSet : public DataClass,
    */
   bool _is_valid() const;
 
+  /**
```
wavefunction91, line 907: The simpler behavior would be to have the default ctor delegate to the general symmetry-laden ctor with default args, rather than having the population happen in the body of the ctor.

cpp/include/qdk/chemistry/data/errors.hpp
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE.txt in the project root for
+// license information.
+
+#pragma once
+
+#include <stdexcept>
+#include <string>
+
+namespace qdk::chemistry::data {
```
wavefunction91: This should not live in the data directory and should live in an appropriate namespace

cpp/include/qdk/chemistry/data/errors.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * @brief Root of the QDK/Chemistry typed-error hierarchy.
+ *
+ * All structured errors raised by the symmetry vocabulary, the
+ * symmetry-blocked storage primitives, and the single-particle basis
+ * containers derive from this type. It derives from @c std::runtime_error so
+ * existing @c catch(const std::exception&) sites continue to function.
```
wavefunction91: This should not be limited to the symmetry errors.

cpp/include/qdk/chemistry/data/hamiltonian.hpp
```diff
@@ -57,45 +58,43 @@ class HamiltonianContainer {
...
-   * @param one_body_integrals One-electron integrals in MO basis [norb x norb]
-   * @param orbitals Shared pointer to molecular orbital data for the system
-   * @param core_energy Core energy (nuclear repulsion + inactive orbital
-   * energy)
-   * @param inactive_fock_matrix Inactive Fock matrix for the selected active
-   * space
-   * @param type Type of Hamiltonian (Hermitian by default)
-   * @throws std::invalid_argument if orbitals pointer is nullptr
+   * and inactive Fock matrix (restricted).
+   * @deprecated Use the SBT-native constructor instead.
```
wavefunction91: SBT is not a well known acronym. It should be spelled out when in documentation.

cpp/include/qdk/chemistry/data/hamiltonian.hpp
```diff
@@ -57,45 +58,43 @@ class HamiltonianContainer {
...
   HamiltonianContainer(const Eigen::MatrixXd& one_body_integrals_alpha,
                        const Eigen::MatrixXd& one_body_integrals_beta,
                        std::shared_ptr<Orbitals> orbitals, double core_energy,
                        const Eigen::MatrixXd& inactive_fock_matrix_alpha,
                        const Eigen::MatrixXd& inactive_fock_matrix_beta,
                        HamiltonianType type = HamiltonianType::Hermitian);
 
+  /**
+   * @brief SBT-native constructor for active space Hamiltonian.
+   * @param h1 One-body integrals as a rank-2 symmetry-blocked tensor.
```
wavefunction91: Short names like h1 and h2 are not descriptive. Keeping the old names of one/two body integrals is the preferred.

cpp/include/qdk/chemistry/data/hamiltonian_containers/cholesky.hpp
```diff
@@ -210,13 +218,14 @@ class CholeskyHamiltonianContainer : public HamiltonianContainer {
...
-   * Three-center integrals in MO basis, where each channel is stored as a
-   * matrix of dimension [(norb x norb) x naux] where norb x norb is stored in
-   * row major order
-   */
-  const std::pair<std::shared_ptr<Eigen::MatrixXd>,
-                  std::shared_ptr<Eigen::MatrixXd>>
+  /// SBT-canonical three-center integrals (source of truth).
+  /// Row axis: MO spin sym (extent = norb^2). Column axis: no sym (extent =
+  /// naux).
+  std::shared_ptr<const SymmetryBlockedTensor<2>> _three_center_sbt;
```
wavefunction91: SBT needs to be generalized to support rank 3 tensors.

cpp/include/qdk/chemistry/data/orbital_containers/basis_coefficients.hpp
```diff
@@ -0,0 +1,110 @@
...
+ * mo_label) is the matrix expressing the modes of @c mo_label in terms of the
+ * atomic orbitals of @c ao_label.
+ *
+ * Aliasing follows the wrapped tensor: for RHF/ROHF the producer supplies the
+ * @c (alpha, alpha) and @c (beta, beta) blocks as the same @c shared_ptr (so
+ * @ref is_restricted is true); for UHF the blocks are distinct.
+ */
+class BasisCoefficients : public DataClass {
+ public:
+  using Sbt = SymmetryBlockedTensor<2, double>;
```
wavefunction91: Sbt is not an appropriate typedef per coding conventions.

cpp/include/qdk/chemistry/data/orbital_containers/basis_coefficients.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * @brief Construct from a rank-2 symmetry-blocked tensor of coefficients.
+   * @param coefficients Non-null rank-2 tensor with axes [ao, mo]
+   * @throws std::invalid_argument if @p coefficients is null
+   */
+  explicit BasisCoefficients(std::shared_ptr<const Sbt> coefficients);
+
+  /** @brief The underlying rank-2 symmetry-blocked tensor. */
+  const std::shared_ptr<const Sbt>& tensor() const { return _coefficients; }
```
wavefunction91: I would just call this coefficients.

cpp/include/qdk/chemistry/data/orbital_containers/orbital_space_partitioning.hpp
```diff
@@ -0,0 +1,129 @@
...
+ *        @ref OrbitalSpace subspaces.
+ *
+ * Holds one @ref SymmetryBlockedIndexSet per @ref OrbitalSpace. All five index
+ * sets share the same symmetry vocabulary and mode extents. A default (empty)
+ * partitioning, in which every mode is @ref OrbitalSpace::Active, can be built
+ * with @ref all_active.
+ */
+class OrbitalSpacePartitioning : public DataClass {
+ public:
+  static constexpr std::size_t kNumSpaces = 5;
```
wavefunction91: Constants should be snake case per coding conventions.

cpp/include/qdk/chemistry/data/orbitals.hpp
```diff
@@ -435,15 +521,79 @@ class Orbitals : public DataClass,
 
  protected:
   /**
-   * Orbital coefficients [AO x MO] for (alpha, beta) spin channels
+   * Cached single-particle (molecular-orbital) symmetries, lazily built.
    */
-  std::pair<std::shared_ptr<Eigen::MatrixXd>, std::shared_ptr<Eigen::MatrixXd>>
+  mutable std::shared_ptr<const Symmetries> _mo_symmetries_cache = nullptr;
```
wavefunction91: Why is this needed? It should just reference the underlying SBT. The v1 APIs are just (checked) facades to v2 APIs. There's no need to cache.

cpp/include/qdk/chemistry/data/orbitals.hpp
```diff
@@ -658,4 +808,20 @@ static_assert(DataClassCompliant<Orbitals>,
...
+ * For an @ref Orbitals carrying a basis set, this returns the AO symmetries of
+ * that basis set (see @ref BasisSet::ao_symmetries). For a @ref ModelOrbitals
+ * (or any single-particle basis without an underlying AO basis set), it returns
+ * @c nullptr.
+ *
+ * @param basis Single-particle basis to inspect (may be @c nullptr)
+ * @return Shared pointer to the AO @ref Symmetries, or @c nullptr if the basis
+ *         has no underlying AO basis set
+ */
+std::shared_ptr<const Symmetries> ao_symmetries(
```
wavefunction91: I do not see the value in this function.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * @brief Closed enumeration of the single-particle symmetry axes a basis may
+ * be blocked under.
+ *
+ * Only @c Spin is populated in the current release; the remaining axes are
+ * declared for forward compatibility with point-group, time-reversal
+ * (Kramers), bosonic-mode, lattice-site, and lattice-momentum partitions.
+ */
+enum class AxisName { Spin, PointGroup, Kramers, Mode, Site, LatticeMomentum };
```
wavefunction91: The only thing in this PR should be Spin.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+/**
+ * @brief Abstract value carried by a single symmetry axis.
+ *
+ * Concrete subclasses (e.g. @ref SpinValue) represent one label of one axis.
+ * Instances are immutable and are shared via @c shared_ptr<const> so that
+ * orbit-equivalent uses can be interned. Subclasses must register a
+ * deserialization handler via @ref register_symmetry_axis_value so that
+ * @ref symmetry_axis_value_from_json can reconstruct them.
+ */
+class SymmetryAxisValue {
```
wavefunction91, line 45: Why is this not a proper DataClass?

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ public:
+  virtual ~SymmetryAxisValue() = default;
+
+  /** @brief The axis this value belongs to. */
+  virtual AxisName axis() const = 0;
+
+  /** @brief Stable serialization tag identifying the concrete subclass. */
+  virtual std::string kind_name() const = 0;
```
wavefunction91: I do not see why both of these should exist - every concrete subclass will correspond to exactly one axis kind/type, which is codified in axis. Axis already has a string converter.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+  std::size_t hash() const override;
+  nlohmann::json to_json() const override;
+
+  /** @brief Reconstruct a @ref SpinValue from its JSON payload. */
+  static std::shared_ptr<const SymmetryAxisValue> from_json(
+      const nlohmann::json& j);
+};
+
+/**
+ * @brief One named symmetry partition the basis is blocked under.
```
wavefunction91: Why does this refer to a basis here?

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * Holds the axis name, the ordered set of admissible labels, and an
+ * @c equivalent flag indicating whether the orbit partners under this axis
+ * share storage (restricted) or are stored independently (unrestricted).
```
wavefunction91: Do not refer to restricted/unrestricted here. That is purely a spin descriptor in the domain context. This goes for other documentation as well, only refer to these terms if spin is exclusively involved.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ */
+class SymmetryAxis {
+  AxisName _name;
+  std::vector<std::shared_ptr<const SymmetryAxisValue>> _labels;
+  bool _equivalent;
```
wavefunction91, line 177: This indicates global equivalence amongst the labels in the symmetry axis. Are there ever instances where there are equivalences in sub groups but not across all irreps in the axis?

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+ public:
+  SymmetryAxis(AxisName name,
```
wavefunction91, line 191: I'll leave this comment here but it applies everywhere. All public APis need proper doxygen docs.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+  bool operator==(const SymmetryAxis& other) const;
+  bool operator!=(const SymmetryAxis& other) const { return !(*this == other); }
+  std::size_t hash() const;
+
+  nlohmann::json to_json() const;
+  static SymmetryAxis from_json(const nlohmann::json& j);
+};
+
+/**
+ * @brief A symmetry vocabulary: the ordered set of axes a basis is blocked
```
wavefunction91: I'm not convinced that "vocabulary" is the proper term to refer to this. It should be renamed to something with appropriate domain context.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+};
+
+/**
+ * @brief A composite addressing key: one @ref SymmetryAxisValue per axis.
+ *
+ * Used to address a single block of a @ref SymmetryBlockedTensor or a single
+ * label set of a @ref SymmetryBlockedIndexSet. The hash is precomputed at
+ * construction so the label can be used as an unordered-map key cheaply.
+ */
+class SymmetryLabel {
```
wavefunction91, line 552: This should be immutable (a dataclass if it's not too heavy). Therefore "recompute_hash" should be removed.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * @c kind tag through the registry.
+ * @throws SymmetryConditionError if the kind tag is not registered.
+ */
+std::shared_ptr<const SymmetryAxisValue> symmetry_axis_value_from_json(
+    const nlohmann::json& j);
+
+namespace axes {
+
+/**
+ * @brief Build a spin axis carrying the @f$\alpha@f$ and @f$\beta@f$ labels.
```
wavefunction91: Alpha/beta are only spin-1/2 and do not belong in this documentation. This should be very general.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * @param two_s Twice the total spin (unused beyond label construction in the
+ *        Sz-only configuration; reserved for forward compatibility).
```
wavefunction91: The parenthetical is a digression.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+ * @param equivalent Whether orbit partners share storage (restricted).
+ */
+SymmetryAxis spin(int two_s, bool equivalent = true);
+
+/** @brief Interned shared @f$\alpha@f$ spin value (@f$2 M_s = +1@f$). */
+const std::shared_ptr<const SpinValue>& alpha();
+
+/** @brief Interned shared @f$\beta@f$ spin value (@f$2 M_s = -1@f$). */
+const std::shared_ptr<const SpinValue>& beta();
```
wavefunction91: These comments should make the spin-1/2 distinction.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -0,0 +1,454 @@
...
+    const std::vector<SymmetryLabel>& values) {
+  return [&]<std::size_t... I>(std::index_sequence<I...>) {
+    return std::array<SymmetryLabel, Rank>{values[I]...};
+  }(std::make_index_sequence<Rank>{});
+}
+
+}  // namespace detail
+
+/**
+ * @brief Immutable symmetry-addressed container of per-block storage.
```
wavefunction91: This is a global comment: do not over index on immutability in public documentation. This is immutable by convention (DataClass).

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,246 @@
...
+        const nlohmann::json&)>
+        from_json);
+
+/**
+ * @brief Reconstruct a @ref SymmetryAxisValue from JSON by dispatching on its
+ * @c kind tag through the registry.
+ * @throws SymmetryConditionError if the kind tag is not registered.
+ */
+std::shared_ptr<const SymmetryAxisValue> symmetry_axis_value_from_json(
+    const nlohmann::json& j);
```
wavefunction91: These should be internal implementation details. We are not supporting external extensions of symmetries. If this mechanism can be made simpler as backend dispatch on the supported symmetry types, it should.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -0,0 +1,454 @@
...
+struct LabelsHash {
+  std::size_t operator()(
+      const std::array<SymmetryLabel, Rank>& labels) const noexcept {
+    std::size_t seed = 0;
+    for (const auto& label : labels) {
+      seed ^= label.hash() + 0x9e3779b97f4a7c15ULL + (seed << 6) + (seed >> 2);
+    }
+    return seed;
+  }
+};
```
wavefunction91, line 47: Why is this here and not with the other related hash function? Also, it should use the supplied hash_combine functionality for consistency.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -0,0 +1,454 @@
...
+    for (const auto& [labels_a, ptr_a] : _blocks) {
+      for (const auto& [labels_b, ptr_b] : _blocks) {
+        if (!(labels_a == labels_b) && ptr_a.get() == ptr_b.get() &&
+            ptr_a != nullptr) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
wavefunction91, line 307: I do not see the value in determining this here - this is a downstream consumer responsibility, this class is just storage.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -0,0 +1,454 @@
...
+
+  /** @brief Representative (labels, block-pointer) pairs, one per unique data
+   * pointer. */
+  std::vector<std::pair<Labels, BlockPtr>> canonical_blocks() const {
+    std::vector<std::pair<Labels, BlockPtr>> result;
+    for (const auto& group : _group_by_pointer()) {
+      result.emplace_back(group.representative, group.ptr);
+    }
+    return result;
+  }
```
wavefunction91: I do not see the value in these functions. This is a global comment - this PR should be the minimal subset of functionality to achieve the swap from the Sz-symmetry laden storage to the symmetry aware storage with v1 back port shims covering the existing APIs. All API additions should be viewed and judged through this lens.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -0,0 +1,454 @@
...
+    return std::array<SymmetryLabel, Rank>{values[I]...};
+  }(std::make_index_sequence<Rank>{});
+}
+
+}  // namespace detail
+
+/**
+ * @brief Immutable symmetry-addressed container of per-block storage.
+ *
+ * A @ref SymmetryBlocked stores a sparse map from per-slot
```
wavefunction91, line 70: It is not called out in the documentation and it's not clear it's a covered edge case - what happens if I don't define a symmetry on a particular axis. The canonical behavior should be that ever tensor has a "trivial" symmetry blocking (the entire range) that should be populated and trivially constructed when accessing by label. Effort should go into making this a good user experience (I.e. accessing the trivial symmetry in the accessors should be possible through empty brackets, I don't want users to have to specify the trivial symmetry manually)

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @brief Immutable set of symmetry-blocked, sorted-unique integer indices.
+ *
+ * A @ref SymmetryBlockedIndexSet describes, for each admissible
+ * @ref SymmetryLabel of a single @ref Symmetries vocabulary, a sorted set of
+ * unique indices drawn from @c [0, extent) for that label. It is used to carve
+ * out symmetry-respecting subspaces (for example core / active / virtual
+ * orbital partitions) of a @ref SymmetryBlockedTensor.
+ */
+class SymmetryBlockedIndexSet : public DataClass {
```
wavefunction91: This should inherit from the SymmetryBlocked concept.

cpp/include/qdk/chemistry/data/orbital_containers/orbital_space_partitioning.hpp
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
wavefunction91: The functionality in this file is scope creep that was not requested in the spec. We should not fundamentally change how we do space partitioning in this PR - this should be storage only.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_sparse_map.hpp
```diff
@@ -0,0 +1,223 @@
...
+  void to_json_file(const std::string& filename) const override {
+    std::ofstream out(filename);
+    if (!out) {
+      throw std::runtime_error("Failed to open file for writing: " + filename);
+    }
+    out << to_json().dump(2);
+  }
+
+  void to_hdf5(H5::Group& /*group*/) const override {
+    throw std::runtime_error(
```
wavefunction91, line 231: This is a global comment: all data classes should have both json and hdf5 schemas. For instances where the internal storage has changed, backport shims should be added on the back end (involving a bump to the schema version) to populate new data from old serializations. These should be unit tested as well. If the shims are used, deprecation warnings should be issued.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_tensor.hpp
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * Rank-1 and rank-4 blocks are stored as (flat) column vectors; rank-2 blocks
+ * are stored as dense matrices.
+ */
+template <std::size_t Rank, class Scalar>
+struct TensorType;
+
+template <class S>
+struct TensorType<1, S> {
+  using T = Eigen::Matrix<S, Eigen::Dynamic, 1>;
```
wavefunction91: use `type` as the typedef

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_tensor.hpp
```diff
@@ -0,0 +1,317 @@
...
+S scalar_from_json(const nlohmann::json& j) {
+  if constexpr (is_complex_scalar<S>::value) {
+    using Real = typename S::value_type;
+    return S(j.at(0).get<Real>(), j.at(1).get<Real>());
+  } else {
+    return j.get<S>();
+  }
+}
+
+}  // namespace detail
```
wavefunction91, line 67: two comments:
1. I'm sure we have a complex scalar deduction primitive elsewhere in the code. Use that
2. This seems like a generally usesful piece of functionality (serializing complex to JSON) - is this not possible though nlohmann::json? If so, use that, if not, place this in the utilities to be use by other callers.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_tensor.hpp
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * @brief Immutable symmetry-blocked dense tensor.
+ *
+ * A @ref SymmetryBlockedTensor stores the non-zero symmetry sectors of a
+ * rank-@p Rank tensor as a sparse map from per-slot @ref SymmetryLabel arrays
+ * to dense Eigen blocks. Each slot carries its own @ref Symmetries vocabulary
+ * and a per-label extent. Blocks are held via @c shared_ptr<const Tensor> so
+ * that orbit-equivalent sectors can alias the same storage.
+ *
+ * Orbit equivalence is defined on the spin axis: a simultaneous
```
wavefunction91: Global comment: "orbit equivalence" is technically correct, but might get confused as referring to orbitals to domain researchers. Find another way to describe it.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_tensor.hpp
```diff
@@ -0,0 +1,317 @@
...
+ * shares the same @ref Symmetries instance and the spin axis is marked
+ * @c equivalent (restricted), the constructor auto-aliases each orbit's
+ * partner block to the supplied representative. When the slots carry distinct
+ * @ref Symmetries (intertwiner storage such as basis coefficients), no
+ * auto-aliasing is performed.
+ *
+ * Instances are immutable: the full block map is supplied at construction and
+ * there is no mutation API.
+ *
+ * @tparam Rank   Tensor rank (1, 2 or 4 are instantiated).
```
wavefunction91: Per the other comment, we need to be able to support rank3 for Cholesky.

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -537,6 +580,15 @@ class WavefunctionContainer {
   mutable std::shared_ptr<VectorVariant> _two_rdm_spin_dependent_aabb = nullptr;
   mutable std::shared_ptr<VectorVariant> _two_rdm_spin_dependent_bbbb = nullptr;
 
+  // SBT-canonical RDM containers (built eagerly from spin-dependent channels)
+  mutable std::shared_ptr<const SymmetryBlockedTensor<2>> _one_rdm_sbt;
+  mutable std::shared_ptr<const SymmetryBlockedTensor<4>> _two_rdm_sbt;
```
wavefunction91: Global comment - the goal of this PR is to move the internal storage over the SBT - this should *replace* the existing internal storage, and public access via the v1 APIs should be done directly without caching.

cpp/src/qdk/chemistry/data/orbital_containers/basis_coefficients.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <fstream>
+#include <qdk/chemistry/data/orbital_containers/basis_coefficients.hpp>
+#include <sstream>
+#include <stdexcept>
+
+namespace qdk::chemistry::data {
+
+namespace {
+
+constexpr const char* kHdf5JsonDataset = "basis_coefficients_json";
```
wavefunction91: Global comment - constants like this should be snake case.

cpp/src/qdk/chemistry/data/orbital_containers/basis_coefficients.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <H5Cpp.h>
+
+#include <fstream>
+#include <qdk/chemistry/data/orbital_containers/basis_coefficients.hpp>
+#include <sstream>
+#include <stdexcept>
+
+namespace qdk::chemistry::data {
+
+namespace {
```
wavefunction91: global comment: never use anonymous namespaces

cpp/src/qdk/chemistry/data/orbital_containers/basis_coefficients.cpp
```diff
@@ -0,0 +1,121 @@
...
+  dataset.write(payload, str_type);
+}
+
+std::string read_json_string(H5::Group& group) {
+  H5::StrType str_type(H5::PredType::C_S1, H5T_VARIABLE);
+  auto dataset = group.openDataSet(kHdf5JsonDataset);
+  std::string payload;
+  dataset.read(payload, str_type);
+  return payload;
+}
```
wavefunction91: Remove these per the other comment on their usage,

cpp/src/qdk/chemistry/data/orbital_containers/basis_coefficients.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+}  // namespace
+
+BasisCoefficients::BasisCoefficients(std::shared_ptr<const Sbt> coefficients)
```
wavefunction91: Global Comment - using Sbt (or other abbreviations) in these kinds of contexts make the code harder to read.

cpp/src/qdk/chemistry/data/orbital_containers/basis_coefficients.cpp
```diff
@@ -0,0 +1,121 @@
...
+  std::ofstream out(filename);
+  if (!out) {
+    throw std::runtime_error("Failed to open file for writing: " + filename);
+  }
+  out << to_json().dump(2);
+}
+
+void BasisCoefficients::to_hdf5(H5::Group& group) const {
+  write_json_string(group, to_json().dump());
+}
```
wavefunction91: Global comment: HDF5 serialization should be direct and efficient and not go through textual intermediaries. In this instance, it should delegate to SBT functions (presuming if they exist, and if they don't, write them)

cpp/include/qdk/chemistry/data/orbital_containers/basis_coefficients.hpp
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
wavefunction91: Global comment - I see no reason to add the `BasisCoefficients` class, `OrbitalsEnergy` class, and `ORbitalSpacePAritioning` class. `Orbitals` should simply be composed of SB-objects as is has been in v1.

cpp/src/qdk/chemistry/data/symmetry/symmetry.cpp
```diff
@@ -0,0 +1,403 @@
...
+#include <qdk/chemistry/data/errors.hpp>
+#include <unordered_map>
+#include <utility>
+
+namespace qdk::chemistry::data {
+
+namespace {
+
+// Mix a value into a running hash (boost::hash_combine style).
+inline void hash_combine(std::size_t& seed, std::size_t value) {
```
wavefunction91: Use the standard function in the code base for this.

cpp/src/qdk/chemistry/data/symmetry/symmetry.cpp
```diff
@@ -0,0 +1,403 @@
...
+void ensure_builtin_kinds_registered() {
+  static std::once_flag flag;
+  std::call_once(flag, [] {
+    register_symmetry_axis_value("spin", &SpinValue::from_json);
+  });
+}
+
+}  // namespace
+
+void register_symmetry_axis_value(const std::string& kind_name,
```
wavefunction91: Per the other comment - the entire runtime registry of symmetries is not needed.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
wavefunction91, line 45: IT should be attempted to make `SymmetryAxisValue` constexpr/consteval.

cpp/include/qdk/chemistry/data/hamiltonian_containers/canonical_four_center.hpp
```diff
@@ -42,9 +42,9 @@ class CanonicalFourCenterHamiltonianContainer : public HamiltonianContainer {
   /**
    * @brief Constructor for restricted active space Hamiltonian with four center
    * integrals.
-   * @deprecated Use the SBT-native constructor instead.
+   * @deprecated Use the SymmetryBlockedTensor-native constructor instead.
    */
-  [[deprecated("Use the SBT-native constructor instead.")]]
+  [[deprecated("Use the SymmetryBlockedTensor-native constructor instead.")]]
```
wavefunction91: Just call this SymmetryBlockedTensor

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -49,24 +50,23 @@ std::array<SymmetryLabel, Rank> make_labels(
...
  * @f$\alpha \leftrightarrow \beta@f$ swap across all slots. When every slot
  * shares the same @ref Symmetries instance and the spin axis is marked
- * @c equivalent (restricted), the constructor auto-aliases each orbit's
- * partner block to the supplied representative. When the slots carry distinct
- * @ref Symmetries (intertwiner storage such as basis coefficients), no
- * auto-aliasing is performed.
+ * @c equivalent, the constructor auto-aliases each partner block to the
+ * supplied representative. When the slots carry distinct @ref Symmetries
+ * (intertwiner storage such as basis coefficients), no auto-aliasing is
+ * performed.
```
wavefunction91: This is incorrect - basis coefficients for e.g. RHF should alias the same storage for the MO matrix. Check that this is the case in the code currently and amend this comment accordingly.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_tensor.hpp
```diff
@@ -138,18 +113,17 @@ class SymmetryBlockedTensor
...
+        << ", independent=" << this->_group_by_pointer().size() << ")";
     return oss.str();
   }
 
   nlohmann::json to_json() const override {
     nlohmann::json j;
     j["type"] = "SymmetryBlockedTensor";
     j["rank"] = Rank;
-    j["scalar"] = detail::is_complex_scalar<Scalar>::value ? "complex" : "real";
+    j["scalar"] = utils::is_complex_scalar<Scalar>::value ? "complex" : "real";
```
wavefunction91: Change this (and other instances) to `utils::is_complex_scalar_v<Scalar>`

cpp/include/qdk/chemistry/data/hamiltonian.hpp
```diff
@@ -57,45 +58,43 @@ class HamiltonianContainer {
...
+  [[deprecated("Use the SymmetryBlockedTensor-native constructor instead.")]]
   HamiltonianContainer(const Eigen::MatrixXd& one_body_integrals_alpha,
                        const Eigen::MatrixXd& one_body_integrals_beta,
                        std::shared_ptr<Orbitals> orbitals, double core_energy,
                        const Eigen::MatrixXd& inactive_fock_matrix_alpha,
                        const Eigen::MatrixXd& inactive_fock_matrix_beta,
                        HamiltonianType type = HamiltonianType::Hermitian);
 
+  /**
+   * @brief SymmetryBlockedTensor-native constructor for active space Hamiltonian.
```
wavefunction91: As the canonical behavior going forward will be to use the SBT constructors, calling it out as something particular in the docs is not warranted.

cpp/include/qdk/chemistry/utils/scalar_traits.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * two-element JSON array @c [real, imag].
+ */
+template <class S>
+nlohmann::json scalar_to_json(const S& value) {
+  if constexpr (is_complex_scalar<S>::value) {
+    return nlohmann::json::array({value.real(), value.imag()});
+  } else {
+    return value;
+  }
+}
```
wavefunction91: These don't belong in a traits header.

cpp/include/qdk/chemistry/data/errors.hpp
```diff
@@ -4,191 +4,32 @@
```
wavefunction91: It seems that this file was not deleted in the revision, despite it's functionality being moved elsewhere.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
```diff
@@ -0,0 +1,235 @@
...
+ */
+class SymmetryAxisValue {
+ public:
+  virtual ~SymmetryAxisValue() = default;
+
+  /** @brief The axis this value belongs to. */
+  virtual AxisName axis() const = 0;
+
+  /** @brief Value-equality against another axis value. */
+  virtual bool equals(const SymmetryAxisValue& other) const = 0;
```
mmoerchen, line 68: Why is there no `operator==` and `operator!=`?

cpp/include/qdk/chemistry/data/basis_set.hpp
```diff
@@ -523,6 +550,26 @@ class BasisSet : public DataClass,
...
+   * @brief Symmetry definitions the atomic-orbital basis is blocked under
+   *
+   * Defaults to a restricted spin axis (@c axes::spin(1, true)). Override at
+   * construction with the AO-symmetries constructor overload.
+   *
+   * @return Shared pointer to the AO @ref Symmetries
+   */
+  std::shared_ptr<const Symmetries> ao_symmetries() const;
```
mmoerchen: Why is it `ao_symmetries()` and not just `symmetries()`, like in the other classes?

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -423,6 +424,48 @@ class WavefunctionContainer {
...
+   * Blocks: (αα), (ββ). Restricted aliases β to α.
+   * @return Const reference to the 1-RDM SymmetryBlockedTensor.
+   * @throws std::runtime_error if not available.
+   */
+  const SymmetryBlockedTensor<2>& one_rdm() const;
+
+  /**
+   * @brief 1-RDM block for the given row/column symmetry labels.
+   */
+  const Eigen::MatrixXd& one_rdm_block(const SymmetryLabel& row,
```
mmoerchen: Why did we drop support for complex RDMs?

cpp/include/qdk/chemistry/data/wavefunction.hpp
```diff
@@ -423,6 +424,48 @@ class WavefunctionContainer {
...
 
+  // ---- SymmetryBlockedTensor RDM accessors -------------------------------------------
+
+  /**
+   * @brief Spin-dependent 1-RDM as a rank-2 symmetry-blocked tensor.
+   * Blocks: (αα), (ββ). Restricted aliases β to α.
+   * @return Const reference to the 1-RDM SymmetryBlockedTensor.
+   * @throws std::runtime_error if not available.
+   */
+  const SymmetryBlockedTensor<2>& one_rdm() const;
```
mmoerchen: I assume this would return the active RDM, but we should still support "total" RDMs.

cpp/include/qdk/chemistry/data/wavefunction.hpp
mmoerchen: We should deprecate all RDM getter

cpp/src/qdk/chemistry/data/hamiltonian_containers/cholesky.cpp
```diff
@@ -290,6 +314,102 @@ CholeskyHamiltonianContainer::make_restricted_three_center_integrals(
...
+  } else {
+    _three_center_integrals.second = _three_center_integrals.first;
+  }
+}
+
+const SymmetryBlockedTensor<3>& CholeskyHamiltonianContainer::three_center()
+    const {
+  QDK_LOG_TRACE_ENTERING();
+  if (!_three_center_sbt) {
+    throw std::runtime_error("Three-center SBT is not set.");
```
mmoerchen: SBT should be symmetry blocked tensor

cpp/src/qdk/chemistry/data/hamiltonian_containers/sparse.cpp
```diff
@@ -624,4 +630,49 @@ SparseHamiltonianContainer::TwoBodyMap SparseHamiltonianContainer::_to_map(
...
+  blocks[{alpha_label, alpha_label, beta_label, beta_label}] = block;
+
+  _h2_sparse = std::make_shared<const SymmetryBlockedSparseMap<4>>(
+      std::move(symmetries), std::move(extents), std::move(blocks));
+}
+
+const SymmetryBlockedSparseMap<4>& SparseHamiltonianContainer::two_body_integrals_sparse()
+    const {
+  if (!_h2_sparse) {
+    throw std::runtime_error("Sparse two-body SBT (two_body_integrals_sparse) is not set.");
```
mmoerchen: SBT should be symmetry blocked tensor

cpp/src/qdk/chemistry/data/hamiltonian.cpp
```diff
@@ -29,23 +29,17 @@ HamiltonianContainer::HamiltonianContainer(
...
   if (!orbitals) {
     throw std::invalid_argument("Orbitals pointer cannot be nullptr");
   }
-
-  // Validate that orbitals have the necessary data
   if (!orbitals->has_active_space()) {
     throw std::runtime_error(
         "Orbitals must have an active space set for HamiltonianContainer");
   }
+  _set_h1_container(one_body_integrals, nullptr);
```
mmoerchen: `_set_h1_container` should be named `_set_one_body_integrals_container` for consistency

cpp/src/qdk/chemistry/data/hamiltonian.cpp
```diff
@@ -54,25 +48,41 @@ HamiltonianContainer::HamiltonianContainer(
...
   if (!orbitals) {
     throw std::invalid_argument("Orbitals pointer cannot be nullptr");
   }
-
-  // Validate that orbitals have the necessary data
   if (!orbitals->has_active_space()) {
     throw std::runtime_error(
         "Orbitals must have an active space set for HamiltonianContainer");
   }
+  _init_h1_views();
```
mmoerchen: `_set_h1_views` should be named `_set_one_body_integrals_views`

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp
```diff
@@ -0,0 +1,101 @@
...
+ */
+class SymmetryBlockedIndexSet
+    : public SymmetryBlocked<1, std::vector<std::uint32_t>> {
+  using Base = SymmetryBlocked<1, std::vector<std::uint32_t>>;
+
+ public:
+  /**
+   * @brief Construct from symmetry definitions, per-label extents, and
+   * per-label index lists.
+   *
```
nabbelbabbel, line 54: Missing docs of the arguments here, is extent ever not size of the indices in that symmetry label? Does it need to be an argument?

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp
```diff
@@ -0,0 +1,101 @@
...
+   * @throws std::invalid_argument if a label is not admissible under
+   *         @p symmetries, lacks a declared extent, or an index list is not
+   *         strictly increasing.
+   * @throws std::out_of_range if an index is >= its label's extent.
+   */
+  SymmetryBlockedIndexSet(
+      std::shared_ptr<const Symmetries> symmetries,
+      std::unordered_map<SymmetryLabel, std::size_t> extents,
+      std::unordered_map<SymmetryLabel, std::vector<std::uint32_t>> indices);
```
nabbelbabbel, line 76: This seems to be working also in the case where alpha extent and beta extend in the spin label are not of equal size, is there an explicit test that this is in-fact fully supported?

python/src/qdk_chemistry/data/symmetries.py
```diff
@@ -20,13 +26,13 @@
 
     from qdk_chemistry.data import Ansatz, Wavefunction
 
-__all__ = ["Symmetries"]
+__all__ = ["SymmetriesV1"]
 
 
-class Symmetries(DataClass):
+class SymmetriesV1(DataClass):
```
nabbelbabbel: I don't understand why we make this intermediate change. Changing the name breaks all existing code similar to removing it. If we have the new version and are not having a simple deprecation, then I would vote for a clean change rather than this.

Is there some edge case that cannot be resolved with the new version and hence requires the old version to stay in some shape or form?

cpp/tests/data/test_orbital_containers.cpp
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE.txt in the project root for
+// license information.
+
+#include <gtest/gtest.h>
```
nabbelbabbel: Why does this file exist?

cpp/src/qdk/chemistry/data/orbitals.cpp
```diff
@@ -1125,7 +1276,11 @@ std::shared_ptr<Orbitals> Orbitals::from_hdf5(H5::Group& group) {
     H5::Attribute version_attr = group.openAttribute("version");
     std::string version_str;
     version_attr.read(string_type, version_str);
-    validate_serialization_version(SERIALIZATION_VERSION, version_str);
+    if (version_str == "0.1.0") {
+      // Deprecated v0.1.0 format — accepted for backward compatibility.
+    } else {
+      validate_serialization_version(SERIALIZATION_VERSION, version_str);
```
nabbelbabbel: I can't find an increment on the serialization versions.

Is the version actually still compatible, if not this should be incremented so that it cleanly fails/flags the issue, same for other data classes that have been upgraded.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
nabbelbabbel: I'd like a reply to this.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
nabbelbabbel: I'd like this fixed.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
nabbelbabbel: I'd like a reply to this.

python/src/qdk_chemistry/data/symmetries.py
wavefunction91: Took the "clean change rather than this" call. The naming collision was the real bug — two unrelated concepts both called Symmetries — so the half-measure rename + alias was always going to be the wrong tradeoff. Fixed in 015e904e7:

cpp/tests/data/test_orbital_containers.cpp
wavefunction91: Removed

cpp/src/qdk/chemistry/data/orbitals.cpp
wavefunction91: The serialization schema is updated, but still backwards compatible (old files are still interpretable in the new code, not the other way around). I do want to maintain that for as long as we can. I've added a deprecation warning to this effect.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
wavefunction91: SymmetryLabel was not migrated - it's meant only as a facilitator for hash look up in the SBT. I did make it immutable though, and thus removed the `recompute_hash`.

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
wavefunction91: After messing with it - this makes for very nasty nesting in the serializations - these particular values are meant to act as labels, not as formal program state. They never exist on their own, only in a list (which is now a proper DataClass in `SymmetryAxis`).

cpp/include/qdk/chemistry/data/symmetry/symmetry.hpp
wavefunction91: This is fixed.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp
wavefunction91: All three constructor args now have @param blocks at cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp:51-76. They specifically answer your question:

  extents[label] is the upper bound (exclusive) of the index range from which the indices entries for that label are drawn. The extent is not derived from indices[label].size() because the subset may be strictly smaller than (or sparser than, or empty within) its universe, and the universe boundary must survive serialization independently of the chosen subset.

 So extent does need to be a separate argument — it's the universe size (e.g. all 10 α MOs), while indices is a chosen subset within that universe (e.g. active α = {2,3,4}). The class-level docs at symmetry_blocked_index_set.hpp:23-44 walk through that exact example.

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked_index_set.hpp
wavefunction91: Yes — SymmetryBlockedIndexSetTest.AsymmetricAlphaBetaExtents (cpp/tests/symmetry/test_symmetry_blocked_index_set.cpp:107) and SymmetryBlockedIndexSetTest.AsymmetricExtentsEnforcedPerLabel (:162) cover this. They construct α with extent 8 / indices {0,3,5,7} and β with extent 6 /indices {1,2}, then assert: constructor + accessors preserve the asymmetry; JSON and HDF5 round-trips preserve it; per-label bounds are enforced independently (α rejects index 8, β rejects index 6 even though 6 would fit α's extent).

nabbelbabbel (approved):

PR #491, opened 2026-05-25
Fix bug in stability checker

BoqinZhang:
When the rotation vector has only one degree of freedom, the original Davidson eigensolver may return wrong results. The bug is fixed, and warning message is added in the `is_restricted` method of class `Orbitals`.

PR #498, opened 2026-05-31
Add deterministic content hashing and local caching for algorithms

nabbelbabbel:
- C++ HashContext utility and content_hash() on all DataClass types
- Python-side _hashing module with content_hash for DataClass, scalars, arrays
- Algorithm.hash() method for deterministic run identification
- AlgorithmWrapper in registry adds cache kwarg to run()
- Cache backends: FolderCache, TieredCache with job tracking
- Job metadata persistence for cache entries
- Pybind11 hash() bindings for all algorithm types
- Tests for caching, data class hashing, algorithm hashing


cpp/src/qdk/chemistry/data/orbitals.cpp
```diff
@@ -664,24 +664,25 @@ bool Orbitals::is_restricted() const {
...
+  } else {
+    // Also check energies if both are set
+    if (_energies.first && _energies.second) {
+      if (!_energies.first->isApprox(*_energies.second)) {
+        return false;
+      }
     }
   }
-  return _coefficients.first == _coefficients.second;
+  return true;
```
wavefunction91: This semantic was intentionally changed in #491. Restricted checks should be restricted to whether the alpha/beta storage are aliased, not if by some chance there's an approximate numerical equivalency. This cased bugs in down stream digestion and testing. Please revert to the current main behavior and update the caching accordingly.

cpp/include/qdk/chemistry/data/orbitals.hpp
```diff
@@ -228,14 +228,14 @@ class Orbitals : public DataClass,
   virtual std::vector<size_t> get_all_mo_indices() const;
 
   /**
-   * @brief Check if calculation is restricted (RHF/RKS/ROHF/ROKS)
-   * @return True if orbitals are restricted
+   * @brief Check if calculation is restricted (RHF/RKS)
+   * @return True if alpha and beta coefficients are identical
```
wavefunction91: See the comments in the Orbitals source.

cpp/include/qdk/chemistry/utils/hash_context.hpp
```diff
@@ -0,0 +1,141 @@
...
+ *   ctx.update(some_double);
+ *   ctx.update("some_string");
+ *   std::string hash = ctx.hexdigest();
+ * @endcode
+ */
+class HashContext {
+ public:
+  HashContext();
+
+  // -- Primitive types --
```
wavefunction91: How are we intending for users who want to add their own data types (potentially in a plugin we have no access to) to this hashing system? I'm inclined to say this should be an external registration mechanism ala `cereal` or https://github.com/bennybp/BPHash - Since hashing is going to be managed by our context for our types, users would have to add their types manually to our code to allow them to work transparently through the hash. The way cereal does it is pretty clever, not necessarily proposing a specific solution, but it's worth considering how/whether we want to support user-space extension.

cpp/include/qdk/chemistry/utils/hash_context.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include <string>
+#include <tuple>
+#include <variant>
+#include <vector>
+
+namespace qdk::chemistry::utils {
+
+/**
+ * @class HashContext
+ * @brief Streaming SHA-256 hash context for computing content hashes.
```
wavefunction91: We should outline what this is and what this isn't:
1. We're stating determinism, which will (at least with SHA-256) preclude parallel computation by downstream producers. I'm fine precluding this, but it's worth laying it out as a restriction. There are tree-based hashes we could use to lift this restriction, but I think that's a bit in the weeds for the current PR.
2. I don't think that we're claiming hash stability between versions, are we? Does this mean that the caching mechanism will be limited to a particular major.minor? Worth considering how we want to discuss this - i.e. this is not meant for data preservation, it's meant for maintaining single jobs with checkpoint/restart.
3. Do we really want to call out SHA in this high-level documentation? Presumably we're not bound to this for  all future versions?

python/src/pybind11/data/data_class.cpp
```diff
@@ -229,6 +241,27 @@ This is used for file naming conventions and serialization.
...
+Args:
+    truncate_chars (int): Number of hex characters in the result (default 16)
+
+Returns:
+    str: Hex string content hash
+
+Examples:
+    >>> h1 = structure1.content_hash()
+    >>> h2 = structure2.content_hash()
+    >>> assert h1 == h2  # if structures are identical
```
wavefunction91, line 261: sans mutable caches?

python/src/qdk_chemistry/algorithms/base.py
```diff
@@ -109,6 +114,24 @@ def run(self, *args, **kwargs):
         self._settings.lock()
         return self._run_impl(*args, **kwargs)
 
+    def hash(self, *args, **kwargs) -> str:
```
wavefunction91, line 117: Maybe I'm missing something here - but is there a reason we don't also expose this on the C++ side?

python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -40,6 +40,155 @@
 
     from qdk_chemistry.algorithms.base import Algorithm, AlgorithmFactory
 
+
+class _AlgorithmWrapper:
```
wavefunction91, line 45: Is there a reason we're not propagating this to the C++ layer as well - this could very well be embedded into the Algorithm class itself

cpp/src/qdk/chemistry/data/orbitals.cpp
wavefunction91: @nabbelbabbel this looks to be fixed since I originally started the review, but please confirm

nabbelbabbel: Yes, I caught this too and fixed it

cpp/include/qdk/chemistry/utils/hash_context.hpp
nabbelbabbel: Users should not need to go this deep into the code. If they have a new data class, then they need to define its hashing function. If their new class contains something we cannot hash as a primitive with this context, then they can hash whatever way they see fit inside that function and add their bespoke hash to the context as a string or integer primitive. To me this seems like an edge case that is not worth the plugin abstraction and would be solved with a PR to us if the new data type would be so widely used that we would want to support it.

cpp/include/qdk/chemistry/utils/hash_context.hpp
nabbelbabbel: Added documentation here and in the data class.

python/src/pybind11/data/data_class.cpp
nabbelbabbel: Well, if the structures are loaded into the code and they are identical then this shall always be true.
Unless structure1 is somehow an object from a different version of the code than structure2 and that in the same script/terminal/process.

python/src/qdk_chemistry/algorithms/base.py
nabbelbabbel: Added this to C++ too.

python/src/qdk_chemistry/algorithms/registry.py
nabbelbabbel: Added a version of this.

nabbelbabbel: Ill take this back, with us trying to have this as a public endpoint/plugin and hashing of arbitrary data classes this was the classical template vs override issue that was a mess.

I think we are better off keeping this in the Python layer.
I happy to be convinced otherwise in a follow up PR.

cpp/include/qdk/chemistry/algorithms/detail/algorithm_hash.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <type_traits>
+#include <variant>
+#include <vector>
+
+#include "qdk/chemistry/data/data_class.hpp"
+#include "qdk/chemistry/utils/hash_context.hpp"
+
+namespace qdk::chemistry::algorithms::detail {
+
+inline void hash_algorithm_arg(utils::HashContext& ctx,
```
wavefunction91: Public headers functions should have doxygen comments.

wavefunction91: Is there a performance reason to make these inline vs compiling them into the library? Doesn't seem like this would hit a critical path, and will likely slow down compilation.

cpp/include/qdk/chemistry/algorithms/detail/algorithm_hash.hpp
```diff
@@ -0,0 +1,96 @@
...
+  } else if constexpr (std::is_floating_point_v<T>) {
+    ctx.update(uint8_t(7));
+    ctx.update(static_cast<double>(value));
+  } else if constexpr (std::is_signed_v<T>) {
+    ctx.update(uint8_t(8));
+    ctx.update(static_cast<int64_t>(value));
+  } else {
+    ctx.update(uint8_t(9));
+    ctx.update(static_cast<uint64_t>(value));
+  }
```
wavefunction91: I do not understand this block, nor do I see whether it's sufficiently general.

cpp/include/qdk/chemistry/algorithms/detail/algorithm_hash.hpp
```diff
@@ -0,0 +1,96 @@
...
+  ctx.update(value);
+}
+
+inline void hash_algorithm_arg(utils::HashContext& ctx, const char* value) {
+  ctx.update(uint8_t(1));
+  ctx.update(std::string(value));
+}
+
+template <typename T>
+std::enable_if_t<std::is_arithmetic_v<T>, void> hash_algorithm_arg(
```
wavefunction91: This enable_if can be a concept in C++20.

cpp/include/qdk/chemistry/algorithms/algorithm.hpp
```diff
@@ -61,6 +63,27 @@ class Algorithm {
...
+   * Hashes the algorithm type, algorithm name, current settings, and arguments
+   * that would be passed to run(). The result identifies a run for cache and
+   * checkpoint/restart workflows among compatible builds.
+   *
+   * @param args Arguments that would be forwarded to run()
+   * @return 16-character hex content hash
+   */
+  virtual std::string hash(Args... args) const {
+    qdk::chemistry::utils::HashContext ctx;
+    ctx.update(std::string("qdk-chemistry.algorithm-run.v1"));
```
wavefunction91: Is this necessary?

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
```diff
@@ -34,15 +35,16 @@ struct LabelsHash {
...
-  std::size_t operator()(
-      const std::array<SymmetryLabel, Rank>& labels) const noexcept {
-    std::size_t seed = 0;
+  std::size_t operator()(const std::array<SymmetryLabel, Rank>& labels) const {
+    utils::HashContext ctx;
+    ctx.update("symmetry_labels");
+    ctx.update(static_cast<uint64_t>(Rank));
     for (const auto& label : labels) {
-      seed = utils::hash_combine(seed, label.hash());
+      ctx.update(label.to_json().dump());
```
wavefunction91: I do like the reuse of the `HashContext` here, but why should this go through JSON serialization rather than just updating on the values themselves? I see this happens a few times in the updates, so take this as a global comment.

cpp/include/qdk/chemistry/data/orbitals.hpp
```diff
@@ -312,7 +312,7 @@ class Orbitals : public DataClass,
 
   /**
    * @brief Check if calculation is unrestricted (UHF/UKS)
-   * @return True if orbitals are unrestricted
+   * @return True if alpha and beta coefficients are different
```
wavefunction91: The original comment is the correct one.

cpp/include/qdk/chemistry/data/pauli_operator.hpp
```diff
@@ -58,31 +58,43 @@ SparsePauliWord label_to_sparse_pauli_word(const std::string& label);
...
 
 /**
  * @brief Hash function for pairs of SparsePauliWord (used for multiplication
  * caching).
  *
- * Combines two SparsePauliWordHash results using hash_combine.
+ * Hashes both words through HashContext.
  */
 struct SparsePauliWordPairHash {
   std::size_t operator()(
```
wavefunction91, line 82: This is a performance critical operation - it is used in a hot path for F2Q encoding. I'm fine migrating over to the new hasher, but it can't degrade performance. We'll need a benchmark on the F2Q mapping or a reversion of this migration.

cpp/include/qdk/chemistry/utils/hash_context.hpp
wavefunction91: I disagree - yes, *users* will not need to handle this, but it's reasonable to expect that *developers* may want to update this in the future. It's a zero cost abstraction that I would recommend getting covered.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/util/gauxc_util.h
```diff
@@ -175,54 +176,53 @@ struct GAUXCInput {
...
-      const qdk::chemistry::scf::GAUXCInput& input) const noexcept {
-    // 0x9e3779b9 is the golden ratio hash constant from Boost
-    constexpr size_t HASH_CONSTANT = 0x9e3779b9;
-
-    // Lambda to handle hash combining with a unified approach
-    auto hash_combine = [&](size_t& seed, const auto& value) {
+  size_t operator()(const qdk::chemistry::scf::GAUXCInput& input) const {
+    qdk::chemistry::utils::HashContext ctx;
+    ctx.update("gauxc_input");
+    auto update_hash = [&](const auto& value) {
```
wavefunction91: Isn't this covered by the other hashing utilities?

cpp/src/qdk/chemistry/data/hamiltonian_containers/sparse.cpp
```diff
@@ -647,6 +647,19 @@ SparseHamiltonianContainer::TwoBodyMap SparseHamiltonianContainer::_to_map(
   return m;
 }
 
+void SparseHamiltonianContainer::hash_update(
+    qdk::chemistry::utils::HashContext& ctx) const {
+  HamiltonianContainer::hash_update(ctx);
+  ctx.update(get_container_type());
+  ctx.update(_one_body_sparse);
+  if (_two_body_sparse) {
+    ctx.update(uint8_t(1));
```
wavefunction91: I'm seeing this motif appear quite often - while I understand the motivation, it seems a bit fragile. Can we at least document this?

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
```diff
@@ -17,6 +17,27 @@
 
 namespace qdk::chemistry::data {
 
+namespace {
```
wavefunction91: Please no anonymous namespaces.

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
```diff
@@ -17,6 +17,27 @@
...
+  for (const auto& [qubit, op] : word) {
+    ctx.update(static_cast<std::uint64_t>(qubit));
+    ctx.update(op);
+  }
+}
+
+void hash_bilinear_entry(
+    qdk::chemistry::utils::HashContext& ctx,
+    const std::pair<std::complex<double>, SparsePauliWord>& entry) {
+  ctx.update(entry.first.real());
```
wavefunction91: Why do we not just have an overload for `std::complex`?

cpp/src/qdk/chemistry/data/settings.cpp
```diff
@@ -1962,4 +1962,54 @@ std::shared_ptr<Settings> Settings::from_hdf5(H5::Group& group) {
...
+  // settings_ is a std::map which is already sorted by key
+  ctx.update(static_cast<uint64_t>(settings_.size()));
+  for (const auto& [key, value] : settings_) {
+    ctx.update(key);
+    // Visit the variant and hash each type
+    std::visit(
+        [&ctx](const auto& v) {
+          using T = std::decay_t<decltype(v)>;
+          if constexpr (std::is_same_v<T, bool>) {
+            ctx.update(uint8_t(0));
```
wavefunction91: This might related to the other comment, but can this not just by a serialization of RTTI as opposed to specific indexing?

wavefunction91 (changes requested):

cpp/include/qdk/chemistry/utils/hash_context.hpp
nabbelbabbel: Clarification after offline discussion: the issue appears when hashing algorithms, not when hashing data classes.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/util/gauxc_util.h
nabbelbabbel: It wasn't now it is

cpp/src/qdk/chemistry/data/hamiltonian_containers/sparse.cpp
nabbelbabbel: Reworked this to be more obvious (with an enum) and have a central place for docs about it

cpp/include/qdk/chemistry/data/symmetry/symmetry_blocked.hpp
nabbelbabbel: Yes, remove all json round trips

cpp/include/qdk/chemistry/algorithms/detail/algorithm_hash.hpp
nabbelbabbel: Reworked this, the block catches non-standard integers and floating point types.

cpp/include/qdk/chemistry/data/pauli_operator.hpp
nabbelbabbel: Reverted this to keep the faster hash for keying.

wavefunction91 (approved):

PR #500, opened 2026-06-01
Add nuclear derivative calculators

nabbelbabbel:


cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+namespace qdk::chemistry::algorithms {
+
+/**
+ * @brief Initial basis, orbitals, or wavefunction seed for a derivative run.
+ */
+using NuclearDerivativeSeedType =
+    std::variant<std::shared_ptr<data::Orbitals>,
+                 std::shared_ptr<data::BasisSet>,
+                 std::shared_ptr<data::Wavefunction>, std::string>;
```
mmoerchen, line 32: What would a string represent?

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,235 @@
...
+                data::AlgorithmRef("hamiltonian_constructor", "qdk"),
+                "Hamiltonian constructor used when energy_calculator is a "
+                "multi_configuration_calculator. It builds the active-space "
+                "Hamiltonian from the reference orbitals before the MR energy "
+                "calculation.");
+    set_default("compute_hessian", false,
+                "Whether to compute a nuclear Hessian in addition to energy "
+                "and gradients. Currently supported by the finite_difference "
+                "implementation only.");
+    set_default("finite_difference_step", 1.0e-3,
```
mmoerchen: This setting is specific to a finite difference solver and shouldn't live in the general settings

cpp/include/qdk/chemistry/data/nuclear_gradients.hpp
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * @brief Construct nuclear gradients for a structure.
+   *
+   * @param structure Molecular structure used to compute the gradients.
+   * @param values Atom-major gradient vector with length 3 * number of atoms.
+   * @throws std::invalid_argument If the structure is null or the vector size
+   * does not match the structure.
+   */
+  NuclearGradients(std::shared_ptr<Structure> structure,
+                   const Eigen::VectorXd& values);
```
mmoerchen: Maybe rename values to gradient_values

cpp/include/qdk/chemistry/data/nuclear_gradients.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * @brief Return gradients as a num_atoms by 3 matrix.
+   */
+  Eigen::MatrixXd as_matrix() const;
+
+  /**
+   * @brief Get the number of atoms in the associated structure.
+   */
+  size_t get_num_atoms() const;
```
mmoerchen: The number of atoms should be requested through the structure

cpp/include/qdk/chemistry/data/nuclear_hessian.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * @brief Construct a nuclear Hessian for a structure.
+   *
+   * @param structure Molecular structure used to compute the Hessian.
+   * @param matrix Square 3N by 3N Hessian matrix in Hartree/Bohr^2.
+   * @throws std::invalid_argument If the structure is null or the matrix shape
+   * does not match the structure.
+   */
+  NuclearHessian(std::shared_ptr<Structure> structure,
+                 const Eigen::MatrixXd& matrix);
```
mmoerchen: Maybe rename matrix to hessian_matrix

cpp/include/qdk/chemistry/data/nuclear_hessian.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * @brief Get the Hessian matrix in Hartree/Bohr^2.
+   */
+  const Eigen::MatrixXd& get_matrix() const { return matrix_; }
+
+  /**
+   * @brief Get the number of atoms in the associated structure.
+   */
+  size_t get_num_atoms() const;
```
mmoerchen: The number of atoms should be requested through the structure

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,517 @@
...
+#include <qdk/chemistry/utils/logger.hpp>
+#include <stdexcept>
+
+#include "microsoft/scf.hpp"
+
+namespace qdk::chemistry::algorithms {
+
+namespace {
+
+class ScopedLogLevel {
```
mmoerchen: This could be a useful utility for other algorithms as well

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,517 @@
...
+      orbitals->get_coefficients().first, orbitals->get_coefficients().second,
+      energies_a, energies_b, ao_overlap, basis_set,
+      std::make_tuple(
+          std::vector<size_t>(active_a.begin(), active_a.end()),
+          std::vector<size_t>(active_b.begin(), active_b.end()),
+          std::vector<size_t>(inactive_a.begin(), inactive_a.end()),
+          std::vector<size_t>(inactive_b.begin(), inactive_b.end())));
+}
+
+std::string active_determinant_string(size_t n_active_orbitals,
```
mmoerchen: This function just creates the HF det. I would rename it

cpp/tests/test_nuclear_derivative.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+    auto determinant = Configuration(determinant_string(
+        orbitals->get_active_space_indices().first.size(), n_alpha, n_beta));
+    auto container =
+        std::make_unique<SlaterDeterminantContainer>(determinant, orbitals);
+    return {0.0, std::make_shared<Wavefunction>(std::move(container))};
+  }
+};
+
+std::shared_ptr<Structure> create_h2_structure() {
```
mmoerchen: Use the available test utility functions

cpp/tests/test_nuclear_derivative.cpp
```diff
@@ -0,0 +1,306 @@
...
+  EXPECT_TRUE((*hessian)->get_matrix().allFinite());
+  EXPECT_TRUE((*hessian)->get_matrix().isApprox(
+      (*hessian)->get_matrix().transpose(), 1.0e-8));
+
+  ASSERT_TRUE(wavefunction.has_value());
+  EXPECT_NE(*wavefunction, nullptr);
+}
+
+TEST(NuclearDerivativeCalculatorTest,
+     MultiReferenceFiniteDifferenceReusesSeedActiveSpace) {
```
mmoerchen, line 294: No test is testing the localization feature

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
nabbelbabbel: A basis set string

cpp/include/qdk/chemistry/data/nuclear_gradients.hpp
nabbelbabbel: Done

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
nabbelbabbel: Fixed

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+        "orbital_solver", data::AlgorithmRef("scf_solver", "qdk"),
+        "SCF solver used to generate reference orbitals for multi-reference "
+        "energy paths. This setting is ignored for direct SCF energy paths and "
+        "is skipped when the derivative input seed already provides usable "
+        "orbitals for the current geometry.");
+    set_default("hamiltonian_constructor",
+                data::AlgorithmRef("hamiltonian_constructor", "qdk"),
+                "Hamiltonian constructor used when energy_calculator is a "
+                "multi_configuration_calculator. It builds the active-space "
+                "Hamiltonian from the reference orbitals before the MR energy "
```
wavefunction91:
```suggestion
                "Hamiltonian from the reference orbitals before the energy "
```

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+                "calculation.");
+    set_default("compute_hessian", false,
+                "Whether to compute a nuclear Hessian in addition to energy "
+                "and gradients.");
+    set_default(
+        "reuse_seed_active_space", true,
+        "For multi-reference energy paths, reuse active and inactive orbital "
+        "indices from an Orbitals or Wavefunction seed when fresh reference "
+        "orbitals are generated for another geometry. Orbital coefficients are "
+        "not reused for displaced finite-difference geometries.");
```
wavefunction91: This seems like it should be implementation specific - littering the top-level global settings instance seems like a bad idea.

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+        "Number of alpha electrons in the active space for "
+        "multi_configuration_scf and multi_configuration_calculator energy "
+        "paths. Must be positive when those energy paths are selected.",
+        data::BoundConstraint<int64_t>{0, std::numeric_limits<int64_t>::max()});
+    set_default(
+        "n_active_beta_electrons", static_cast<int64_t>(0),
+        "Number of beta electrons in the active space for "
+        "multi_configuration_scf and multi_configuration_calculator energy "
+        "paths. Must be positive when those energy paths are selected.",
+        data::BoundConstraint<int64_t>{0, std::numeric_limits<int64_t>::max()});
```
wavefunction91: If we're matching the SCF and the MC interfaces - these should be inputs, not settings. Do the non-QM-variants not also take an electron count?

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+
+  /**
+   * @brief Register built-in nuclear derivative calculator implementations.
+   */
+  static void register_default_instances();
+
+  /**
+   * @brief Return the default nuclear derivative implementation name.
+   */
+  static std::string default_algorithm_name() { return "finite_difference"; }
```
wavefunction91:
```suggestion
  static std::string default_algorithm_name() { return "qdk_finite_difference"; }
```

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+};
+
+/**
+ * @class FiniteDifferenceNuclearDerivativeCalculator
+ * @brief Numeric nuclear derivative calculator using central finite
+ * differences.
+ */
+class FiniteDifferenceNuclearDerivativeCalculator
```
wavefunction91: To be consistent with the other algorithms - let's put this in its own header and not expose this in the public API - all other internal implementations (e.g. SCF) only expose the factory in the public header and keep the class names, etc internal.

cpp/include/qdk/chemistry/algorithms/nuclear_derivative.hpp
```diff
@@ -0,0 +1,258 @@
...
+      std::shared_ptr<data::Structure> structure, int charge,
+      int spin_multiplicity,
+      NuclearDerivativeSeedType seed_or_basis) const override;
+};
+
+/**
+ * @class QdkNuclearDerivativeCalculator
+ * @brief Internal QDK derivative calculator using analytic SCF gradients.
+ */
+class QdkNuclearDerivativeCalculator : public NuclearDerivativeCalculator {
```
wavefunction91: Also put this in its own header.

cpp/include/qdk/chemistry/data/nuclear_hessian.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * @brief Get the molecular structure associated with this Hessian.
+   */
+  const std::shared_ptr<Structure> get_structure() const;
+
+  /**
+   * @brief Get the Hessian matrix in Hartree/Bohr^2.
+   */
+  const Eigen::MatrixXd& get_matrix() const { return matrix_; }
```
wavefunction91, line 47: I believe the same ultimately goes for the gradients class too - but it would be good to add accessors that snag a particular atom (pair) in addition to returning the whole raw storage (reshaped or not) - the `get_matrix` kind of solves this for gradients, but the extension here would be non-trivial until we move to proper tensor structures.

cpp/include/qdk/chemistry/utils/logger.hpp
```diff
@@ -108,6 +108,20 @@ class Logger {
...
+   * they match, restores restored_level and returns true. Otherwise leaves the
+   * current level unchanged and returns false.
+   *
+   * @param expected_current The level that must still be current
+   * @param restored_level The level to restore when expected_current matches
+   * @return Whether restored_level was applied
+   */
+  static bool restore_global_level_if_unchanged(LogLevel expected_current,
+                                                LogLevel restored_level);
+
```
wavefunction91, line 124: Why is this in this PR?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -395,11 +414,19 @@ class ERI {
...
-              const auto buf_1234 = buf[0];
-              if (buf_1234 == nullptr) continue;
+              const auto* buf_1234 = buf[0];
+              const double* buf_erf_1234 = nullptr;
+              if (need_erf_exchange) {
+                engine_erf->compute2<::libint2::Operator::erf_coulomb,
+                                     ::libint2::BraKet::xx_xx, 0>(
+                    obs_[s1], obs_[s2], obs_[s3], obs_[s4], sp12_data,
+                    sp34_data);
+                buf_erf_1234 = engine_erf->results()[0];
```
wavefunction91: We should have a similar buf alias for ERF.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -477,7 +504,13 @@ class ERI {
                         for (size_t l = 0; l < n4; ++l, ++ijkl) {
                           const size_t bf4 = bf4_st + l;
 
-                          const auto value = buf_1234[ijkl] * s1234_deg;
+                          const auto coulomb_value =
+                              buf_1234 ? buf_1234[ijkl] : 0.0;
+                          const auto erf_value =
+                              buf_erf_1234 ? buf_erf_1234[ijkl] : 0.0;
+                          const auto value = (exchange_scale * coulomb_value -
+                                              beta * erf_value) *
```
wavefunction91: This is heavy for a hot loop - since these always get contracted in the same way - we can remove the branches by recomputing the contraction buffer gated on the data precondition.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -395,11 +414,19 @@ class ERI {
...
+                                     ::libint2::BraKet::xx_xx, 0>(
+                    obs_[s1], obs_[s2], obs_[s3], obs_[s4], sp12_data,
+                    sp34_data);
+                buf_erf_1234 = engine_erf->results()[0];
+              }
+              if (buf_1234 == nullptr && buf_erf_1234 == nullptr) continue;
 
               // Contract shell quartet (J)
-              if (J)
+              if (J && buf_1234)
```
wavefunction91: This seems to use both buffers, why are we gating only on one.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+#ifdef QDK_CHEMISTRY_ENABLE_MPI
+#include <mpi.h>
+#endif
+
+#include "microsoft/scf.hpp"
+
+namespace qdk::chemistry::algorithms {
+
+namespace {
```
wavefunction91: Please do not use anonymous namespaces.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
wavefunction91: @mmoerchen's comment is a good one, please address.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+std::shared_ptr<data::Structure> copy_structure(
+    const std::shared_ptr<data::Structure>& structure) {
+  if (!structure) {
+    throw std::invalid_argument("Structure must not be null");
+  }
+  return std::make_shared<data::Structure>(*structure);
+}
+
+std::optional<Eigen::VectorXd> broadcast_gradient_from_root(
```
wavefunction91: We do not have a path to test the MPI internals at the moment, please remove this and all added MPI internals form this PR.

cpp/tests/test_nuclear_derivative.cpp
```diff
@@ -0,0 +1,397 @@
...
+#include <stdexcept>
+#include <utility>
+#include <vector>
+
+#include "ut_common.hpp"
+
+using namespace qdk::chemistry::algorithms;
+using namespace qdk::chemistry::data;
+
+namespace {
```
wavefunction91: I'll stop commenting this, but please do not use anonymous namespaces.

cpp/tests/test_nuclear_derivative.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+std::vector<std::vector<size_t>> recorded_active_spaces;
+std::vector<std::vector<size_t>> recorded_inactive_spaces;
+std::vector<std::vector<size_t>> recorded_localized_alpha_spaces;
+std::vector<std::vector<size_t>> recorded_localized_beta_spaces;
+constexpr double kDftAnalyticNumericGradientTolerance = 5.0e-5;
```
wavefunction91: Do not use `k`-prefixed snake-case constants

cpp/tests/test_nuclear_derivative.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+std::string determinant_string(size_t n_active_orbitals, unsigned int n_alpha,
```
wavefunction91: We have other utilities for this kind of stuff, please use them.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -599,15 +633,216 @@ class ERI {
...
+                          }
+                        }
+                      }
+                    }
+                  }
+                }
+
+                if (need_J) dJ_out[coord] += 0.5 * dJ_coord;
+                if (need_K) {
+                  dK_out[coord] -= 0.25 * dK_coord;
```
wavefunction91: The dK_coord sum just above runs over `idm < spin_density_factor_`, i.e. per-spin density blocks. For the closed-shell case the 0.25 on the total density looks right to me. Could you sanity-check the prefactor when `spin_density_factor_ == 2`? The energy path uses `F = J - 0.5*K` for RHF but `F = J_tot - K` per spin for UHF/ROHF, and the energy-weighted density picks up the analogous `W *= 0.5` only when `num_density_matrices_ == 1`. Do the open-shell (doublet/triplet) analytic gradients agree with finite differences? I only see closed-shell LiH in the analytic-vs-numeric tests.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -599,15 +633,216 @@ class ERI {
...
+#ifdef _OPENMP
+      const int thread_id = omp_get_thread_num();
+#else
+      const int thread_id = 0;
+#endif
+      auto& engine = engines[thread_id];
+      const auto& buf = engine.results();
+      auto* engine_erf = need_erf_exchange ? &engines_erf[thread_id] : nullptr;
+
+      for (size_t s1 = 0, s1234 = 0; s1 < nsh; ++s1) {
```
wavefunction91: This iterates all ordered quartets `(s1,s2,s3,s4)` over `[0,nsh)` and distributes with `s1234 % nthreads`, so each thread walks the full space. `build_JK` instead loops over significant-pair ranges with permutational symmetry and a degeneracy factor. Was the full quartet loop a deliberate first cut, or worth mirroring `build_JK` here? Related: line 762 screens on `K_schwarz(s1,s2)*K_schwarz(s3,s4)` while `build_JK` also folds in `P1234_nrm` (line 398). Intentional to drop the density-weighted term on the gradient path?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/eri/LIBINT2_DIRECT/libint2_direct.cpp
```diff
@@ -256,8 +263,8 @@ class ERI {
     const size_t mat_size =
         num_density_matrices * num_atomic_orbitals * num_atomic_orbitals;
     const bool is_rsx = std::abs(omega) > 1e-12;
-
-    if (is_rsx) throw std::runtime_error("RSX + LIBINT2_DIRECT NYI");
+    const double exchange_scale = alpha + beta;
+    const bool need_erf_exchange = K && is_rsx && std::abs(beta) > 1e-12;
```
wavefunction91, line 268: If we're going to add this, add unit tests that cover this path. It's not immediately obvious that it's covered by existing tests.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+                              evaluate_energy(*_settings, mm, charge,
+                                              spin_multiplicity, seed, false)
+                                  .energy) /
+                             (4.0 * step * step);
+        hessian_matrix(i, j) = value;
+        hessian_matrix(j, i) = value;
+      }
+    }
+    if (_settings->get<bool>("symmetrize_hessian")) {
+      hessian_matrix = 0.5 * (hessian_matrix + hessian_matrix.transpose());
```
wavefunction91: Lines 505-506 assign the same value to `hessian_matrix(i,j)` and `(j,i)`, and the diagonal is set once, so the matrix seems already symmetric by construction. What does `symmetrize_hessian` change in that case? (Separately: is `H = 0.5*(H + H.transpose())` OK re: Eigen aliasing if H were ever not symmetric? `.transpose()` aliases the in-place assignment.) Wondering whether the setting can be dropped.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+      "derivatives: " +
+      algorithm_type);
+}
+
+}  // namespace
+
+NuclearDerivativeResult FiniteDifferenceNuclearDerivativeCalculator::_run_impl(
+    std::shared_ptr<data::Structure> structure, int charge,
+    int spin_multiplicity, NuclearDerivativeSeedType seed) const {
+  const ScopedLogLevel scoped_log_level(utils::LogLevel::error);
```
wavefunction91: This raises the process-global log level for the duration of the run to quiet the sub-runs. Given the calculator is otherwise stateless, is mutating global logger state a side effect we want here? What happens to SCF non-convergence warnings on displaced geometries we might actually want surfaced? Is there a per-child-solver verbosity knob we could use instead?

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+      orbitals->get_coefficients().first, orbitals->get_coefficients().second,
+      energies_a, energies_b, ao_overlap, basis_set,
+      std::make_tuple(
+          std::vector<size_t>(active_a.begin(), active_a.end()),
+          std::vector<size_t>(active_b.begin(), active_b.end()),
+          std::vector<size_t>(inactive_a.begin(), inactive_a.end()),
+          std::vector<size_t>(inactive_b.begin(), inactive_b.end())));
+}
+
+std::string hartree_fock_determinant_string(size_t n_active_orbitals,
```
wavefunction91: This hand-rolls an aufbau occupation string `('2'/'u'/'d'/'0')`. Is there a reason not to reuse `data::Configuration::canonical_hf_configuration(n_alpha, n_beta, ...)` (configuration.hpp:180, as scf.cpp:474 does)? That would avoid a second implementation of the same HF-occupation logic drifting from the canonical one. I think this shows up in another place as well.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+      --alpha_remaining;
+    } else if (beta_remaining > 0) {
+      occupation = 'd';
+      --beta_remaining;
+    }
+  }
+  return determinant;
+}
+
+std::shared_ptr<data::Wavefunction> wavefunction_from_orbitals(
```
wavefunction91: This builds an aufbau/HF single-determinant wavefunction from orbitals. `detail::new_aufbau_determinant_wavefunction` (localization.hpp:56) already does this and is what every localizer uses; you're also already using `detail::new_wavefunction` just below at line 359. Could this call the existing helper instead of a local reimplementation

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
```diff
@@ -0,0 +1,580 @@
...
+  if (basis->get_name() == data::BasisSet::custom_name) {
+    throw std::invalid_argument(
+        "Custom orbital basis cannot be reused for displaced numeric "
+        "derivatives");
+  }
+  return basis->get_name();
+}
+
+template <typename Factory>
+typename Factory::return_type create_from_ref(const data::AlgorithmRef& ref);
```
wavefunction91: `create_from_ref` and `active_electrons` are forward-declared here and defined ~180 lines later, so the file reads out of order. Is the forward declaration load-bearing, or could these just be defined before first use? While here: several of the anon-namespace helpers below (`seed_to_orbitals, seed_to_wavefunction, copy_active_space_metadata, hartree_fock_determinant_string, wavefunction_from_orbitals`) have a single call site and exist mainly to feed the MR/localization path. If those paths move to a follow-up PR, most of these collapse; for the ones that stay, would inlining at the single call site read more directly?

wavefunction91 (changes requested):

cpp/include/qdk/chemistry/utils/logger.hpp
nabbelbabbel: It is used in the destructor of the scoped log level helper that now moved into the same place as this function. This should make it clearer.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
nabbelbabbel: I did move it to the logger header so we can use it more widely. I did not spread it around.

cpp/src/qdk/chemistry/algorithms/nuclear_derivative.cpp
nabbelbabbel: Can we be sure that this knob always exists even if it is a plugin? I will make this optiona, default on so one can turn it off.

wavefunction91 (dismissed):

wavefunction91 (approved):

PR #505, opened 2026-06-04
Verstraete-Cirac mapping

### LINKED ISSUE #482: Add Verstraete-Cirac Fermion-to-Qubit Encoding
**Is your feature request related to a problem? Please describe.**

 QDK/Chemistry currently ships only Jordan-Wigner, Bravyi-Kitaev, parity, and SCBK fermion-to-qubit encodings. For 2D-local fermionic problems -- Fermi-Hubbard, lattice gauge theories, fermionic tensor-network ansatze, real-space-grid chemistry -- every nearest-neighbor hop maps to a long Pauli string under these encodings, which inflates Trotter circuit depth and measurement cost.

 **Describe the solution you'd like**

 A working **Verstraete-Cirac encoding** exposed as a new `MajoranaMapping` factory and consumable through `QubitMapper`, with the resulting `QubitHamiltonian` correctly representing the original fermionic Hamiltonian on the physical subspace. Mapper consumers (energy estimation, state preparation, Trotterization) should be able to use it without special-casing.

 **Describe alternatives you've considered**

 Other redundant encodings (Derby-Klassen compact, Bravyi-Kitaev superfast, Setia-Whitfield) would address the same locality problem with different qubit-overhead and stabilizer-structure trade-offs. Verstraete-Cirac is requested first asthe historically canonical choice; the others are natural follow-ups.

 Circuit-level Pauli grouping helps measurement cost under JW/BK but does not reduce Trotter depth.

 **Acceptance criteria**

 - A new `MajoranaMapping` factory for Verstraete-Cirac accepts a general 2D lattice of at least sizes 2x2, 2x3, 3x3, and 4x4 (single spin species) and returns a mapping that `QubitMapper` consumes without error.
 - For the Fermi-Hubbard model on a 2x2 lattice with open boundaries, t=1, U=4, at half-filling, the four lowest eigenvalues of the resulting `QubitHamiltonian` restricted to the codespace match the Jordan-Wigner result to within `1e-10` absolute error.
 - For Fermi-Hubbard on open LxL lattices with L in {2, 3, 4}, the maximum Pauli weight over all nearest-neighbor hopping terms produced by the mapper is the same finite integer for all three L (i.e. independent of system size).
 - The constructed mapping survives JSON and HDF5 round-trips, and the round-tripped mapping produces a `QubitHamiltonian` whose terms equal the original's term-by-term (Pauli string and coefficient).
 - A new pytest module under `python/tests/` covers all of the above; the full existing test suite continues to pass; `pre-commit` succeeds on the diff.
 - The encoding is listed in `docs/source/user/comprehensive/algorithms/qubit_mapper.rst` under "Supported encodings" in the same format as the other entries, and citations for the three references below are added to`docs/source/references.bib`.

 **Additional context**

 - F. Verstraete and J. I. Cirac, *Mapping local Hamiltonians of fermions to local Hamiltonians of spins*, J. Stat. Mech. (2005) P09012.
 - J. D. Whitfield, V. Havlicek, and M. Troyer, *Local spin operators for fermion simulations*, Phys. Rev. A **94**, 030301(R) (2016) [arXiv:1605.09789].
 - V. Havlicek, M. Troyer, and J. D. Whitfield, *Operator locality in the quantum simulation of fermionic models*, Phys. Rev. A **95**, 032332 (2017) [arXiv:1701.07072].

nirajvenkat:
# Feature: Verstraete-Cirac Fermion-to-Qubit Encoding

## Description

Closes [#482](https://github.com/microsoft/qdk-chemistry/issues/482).

We implement the **Verstraete-Cirac encoding**, which maps $$N$$ fermionic modes onto $$O(2N)$$ qubits, or 4 Majorana modes per site. The implementation supports an arbitrary `LatticeGraph` topology.

Added a comprehensive test module [test_verstraete_cirac.py](qdk-chemistry/python/tests/test_verstraete_cirac.py) covering all verification paths listed in the criteria below.

---

## Acceptance/Testing Criteria

- [x] **2D Lattice Mapping support**: `MajoranaMapping.verstraete_cirac` correctly constructs mappings for 2D square lattices with more than 3 sites (sizes 2x2, 2x3, 3x3, and 4x4) without errors.
- [x] **Spectral Validation**: Compared the lowest unique eigenvalues of a $2 \times 2$ periodic Fermi-Hubbard model ($t=1$, $U=4$, half-filling) and $3 \times 2$ Hückel model between Jordan-Wigner and the Verstraete-Cirac codespace. The lowest eigenvalues match to within $< 10^{-10}$ absolute error. 
- [x] **Hopping Operator Locality (Constant Pauli Weight)**: Verified that the maximum Pauli weight of mapped nearest-neighbor hopping terms remains constant across these grid dimensions: 2x2, 3x3, 4x4.
- [x] **Serialization**: Verified that the mapping successfully survives JSON and HDF5 serialization round-trips, producing bit-wise identical Pauli strings and coefficients in the resulting `QubitHamiltonian`.

---

## Verification & Test Results
pytest:
```bash
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_factory_and_dimensions PASSED             [  5%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_majorana_raises_error PASSED              [ 11%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_without_tapering PASSED                   [ 16%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_json_serialization PASSED                 [ 22%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_hdf5_serialization PASSED                 [ 27%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[square-2x2] PASSED [ 33%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[square-3x3] PASSED [ 38%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[square-3x4] PASSED [ 44%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[honeycomb-2x2-periodic] PASSED [ 50%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[triangular-2x2] PASSED [ 55%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[kagome-2x2-periodic] PASSED [ 61%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_stabilizers_and_commutation[kagome-3x3] PASSED [ 66%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_pauli_weight_scaling[square] PASSED       [ 72%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_pauli_weight_scaling[triangular] PASSED   [ 77%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_pauli_weight_scaling[honeycomb] PASSED    [ 83%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracMapping::test_pauli_weight_scaling[kagome] PASSED       [ 88%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracSpectral::test_spectral_validation_2x2_hubbard PASSED   [ 94%]
tests/test_verstraete_cirac.py::TestVerstraeteCiracSpectral::test_spectral_validation_3x2_huckel PASSED    [100%]
```

C++ test:
```bash
[ RUN      ] LatticeGraphTest.Permute
[       OK ] LatticeGraphTest.Permute (0 ms)
```

### DIFF
--- a/cpp/include/qdk/chemistry/data/lattice_graph.hpp
+++ b/cpp/include/qdk/chemistry/data/lattice_graph.hpp
@@ -232,7 +232,7 @@ class LatticeGraph : public DataClass {
    * @throws std::invalid_argument If n == 0.
    */
   static LatticeGraph chain(std::uint64_t n, bool periodic = false,
-                            double t = 1.0);
+                            double t = 1.0, bool optimal_ordering = false);
 
   /**
    * @brief Create a two-dimensional square lattice.
@@ -266,7 +266,7 @@ class LatticeGraph : public DataClass {
    */
   static LatticeGraph square(std::uint64_t nx, std::uint64_t ny,
                              bool periodic_x = false, bool periodic_y = false,
-                             double t = 1.0);
+                             double t = 1.0, bool optimal_ordering = false);
 
   /**
    * @brief Create a two-dimensional triangular lattice.
@@ -305,7 +305,8 @@ class LatticeGraph : public DataClass {
   static LatticeGraph triangular(std::uint64_t nx, std::uint64_t ny,
                                  bool periodic_x = false,
                                  bool periodic_y = false, double t = 1.0,
-                                 int coloring_seed = 0);
+                                 int coloring_seed = 0,
+                                 bool optimal_ordering = false);
 
   /**
    * @brief Create a two-dimensional honeycomb lattice.
@@ -344,7 +345,8 @@ class LatticeGraph : public DataClass {
    */
   static LatticeGraph honeycomb(std::uint64_t nx, std::uint64_t ny,
                                 bool periodic_x = false,
-                                bool periodic_y = false, double t = 1.0);
+                                bool periodic_y = false, double t = 1.0,
+                                bool optimal_ordering = false);
 
   /**
    * @brief Create a two-dimensional kagome lattice.
@@ -394,7 +396,8 @@ class LatticeGraph : public DataClass {
    */
   static LatticeGraph kagome(std::uint64_t nx, std::uint64_t ny,
                              bool periodic_x = false, bool periodic_y = false,
-                             double t = 1.0, int coloring_seed = 0);
+                             double t = 1.0, int coloring_seed = 0,
+                             bool optimal_ordering = false);
 
   /**
    * @brief Edge coloring stored at construction time, if any.
@@ -494,6 +497,22 @@ class LatticeGraph : public DataClass {
   /** @brief Check if a sparse matrix is symmetric within a numerical tolerance.
    */
   static bool _check_symmetry(const Eigen::SparseMatrix<double>& mat);
+
+  /**
+   * @brief Permutes the vertices of the lattice graph according to the given
+   * path.
+   *
+   * Reorders the sparse adjacency matrix using Eigen's permutation operations
+   * and updates the edge coloring to align with the new vertex indexing.
+   *
+   * @param graph The source lattice graph to permute.
+   * @param path The sequence of original vertex indices representing the target
+   * permutation.
+   * @return A new LatticeGraph with the permuted adjacency matrix and edge
+   * coloring.
+   */
+  static LatticeGraph permute(const LatticeGraph& graph,
+                              const std::vector<std::uint64_t>& path);
 
   /// Number of sites (vertices) in the lattice
   std::uint64_t _num_sites;
--- a/cpp/include/qdk/chemistry/data/lattice_graph.hpp
+++ b/cpp/include/qdk/chemistry/data/lattice_graph.hpp
@@ -232,7 +232,7 @@ class LatticeGraph : public DataClass {
    * @throws std::invalid_argument If n == 0.
    */
   static LatticeGraph chain(std::uint64_t n, bool periodic = false,
-                            double t = 1.0, bool optimal_ordering = false);
+                            double t = 1.0, bool dfs_ordering = false);
 
   /**
    * @brief Create a two-dimensional square lattice.
@@ -266,7 +266,7 @@ class LatticeGraph : public DataClass {
    */
   static LatticeGraph square(std::uint64_t nx, std::uint64_t ny,
                              bool periodic_x = false, bool periodic_y = false,
-                             double t = 1.0, bool optimal_ordering = false);
+                             double t = 1.0, bool dfs_ordering = false);
 
   /**
    * @brief Create a two-dimensional triangular lattice.
@@ -306,7 +306,7 @@ class LatticeGraph : public DataClass {
                                  bool periodic_x = false,
                                  bool periodic_y = false, double t = 1.0,
                                  int coloring_seed = 0,
-                                 bool optimal_ordering = false);
+                                 bool dfs_ordering = false);
 
   /**
    * @brief Create a two-dimensional honeycomb lattice.
@@ -346,7 +346,7 @@ class LatticeGraph : public DataClass {
   static LatticeGraph honeycomb(std::uint64_t nx, std::uint64_t ny,
                                 bool periodic_x = false,
                                 bool periodic_y = false, double t = 1.0,
-                                bool optimal_ordering = false);
+                                bool dfs_ordering = false);
 
   /**
    * @brief Create a two-dimensional kagome lattice.
@@ -397,7 +397,7 @@ class LatticeGraph : public DataClass {
   static LatticeGraph kagome(std::uint64_t nx, std::uint64_t ny,
                              bool periodic_x = false, bool periodic_y = false,
                              double t = 1.0, int coloring_seed = 0,
-                             bool optimal_ordering = false);
+                             bool dfs_ordering = false);
 
   /**
    * @brief Edge coloring stored at construction time, if any.
@@ -479,6 +479,22 @@ class LatticeGraph : public DataClass {
    */
   static LatticeGraph from_hdf5(H5::Group& group);
 
+  /**
+   * @brief Permutes the vertices of the lattice graph according to the given
+   * path.
+   *
+   * Reorders the sparse adjacency matrix using Eigen's permutation operations
+   * and updates the edge coloring to align with the new vertex indexing.
+   *
+   * @param graph The source lattice graph to permute.
+   * @param path The sequence of original vertex indices representing the target
+   * permutation.
+   * @return A new LatticeGraph with the permuted adjacency matrix and edge
+   * coloring.
+   */
+  static LatticeGraph permute(const LatticeGraph& graph,
+                              const std::vector<std::uint64_t>& path);
+
  private:
   void hash_update(qdk::chemistry::utils::HashContext& ctx) const override;
 
@@ -497,22 +513,6 @@ class LatticeGraph : public DataClass {
   /** @brief Check if a sparse matrix is symmetric within a numerical tolerance.
    */
   static bool _check_symmetry(const Eigen::SparseMatrix<double>& mat);
-
-  /**
-   * @brief Permutes the vertices of the lattice graph according to the given
-   * path.
-   *
-   * Reorders the sparse adjacency matrix using Eigen's permutation operations
-   * and updates the edge coloring to align with the new vertex indexing.
-   *
-   * @param graph The source lattice graph to permute.
-   * @param path The sequence of original vertex indices representing the target
-   * permutation.
-   * @return A new LatticeGraph with the permuted adjacency matrix and edge
-   * coloring.
-   */
-  static LatticeGraph permute(const LatticeGraph& graph,
-                              const std::vector<std::uint64_t>& path);
 
   /// Number of sites (vertices) in the lattice
   std::uint64_t _num_sites;
--- a/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
+++ b/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
@@ -263,12 +263,9 @@ class MajoranaMapping : public DataClass {
    * Maps fermionic modes on a `LatticeGraph` to qubits.
    *
    * @param lattice The `LatticeGraph` connectivity.
-   * @param num_spin_species Number of spin species (usually 2 for spinful, 1
-   *        for spinless). Default: 2.
    * @return MajoranaMapping with name ``"verstraete-cirac"``.
    */
-  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice,
-                                          std::size_t num_spin_species = 2);
+  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice);
 
  private:
   MajoranaMapping(
--- a/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
+++ b/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
@@ -20,6 +20,8 @@ namespace qdk::chemistry::data {
 
 class Hamiltonian;
 
+class LatticeGraph;
+
 /**
  * @brief Data class describing a fermion-to-qubit encoding.
  *
@@ -103,6 +105,12 @@ class MajoranaMapping : public DataClass {
     return tapering_;
   }
 
+  /// Grid X dimension (for Verstraete-Cirac mapping).
+  std::size_t grid_nx() const { return grid_nx_; }
+
+  /// Grid Y dimension (for Verstraete-Cirac mapping).
+  std::size_t grid_ny() const { return grid_ny_; }
+
   /**
    * @brief Return a copy with tapering removed and the base encoding name
    *        restored.
@@ -162,6 +170,19 @@ class MajoranaMapping : public DataClass {
                                    const std::string& type);
 
   // --- Factory methods for standard encodings ---
+
+  /**
+   * @brief Verstraete-Cirac encoding.
+   *
+   * Maps fermionic modes on a 2D square lattice to qubits.
+   *
+   * @param lattice The 2D square lattice connectivity.
+   * @param num_spin_species Number of spin species (usually 2 for spinful, 1
+   * for spinless). Default: 2.
+   * @return MajoranaMapping with name ``"verstraete-cirac"``.
+   */
+  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice,
+                                          std::size_t num_spin_species = 2);
 
   /**
    * @brief Jordan-Wigner encoding.
@@ -252,7 +273,8 @@ class MajoranaMapping : public DataClass {
       std::vector<std::pair<std::complex<double>, SparsePauliWord>> bilinears,
       std::string name, std::size_t num_modes, std::size_t num_qubits,
       std::string base_encoding,
-      std::optional<TaperingSpecification> tapering = std::nullopt);
+      std::optional<TaperingSpecification> tapering = std::nullopt,
+      std::size_t grid_nx = 0, std::size_t grid_ny = 0);
 
   /// Majorana-to-Pauli table (empty for bilinear-only mappings).
   std::vector<SparsePauliWord> table_;
--- a/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
+++ b/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
@@ -260,9 +260,9 @@ class MajoranaMapping : public DataClass {
   /**
    * @brief Verstraete-Cirac encoding.
    *
-   * Maps fermionic modes on a 2D square lattice to qubits.
+   * Maps fermionic modes on a `LatticeGraph` to qubits.
    *
-   * @param lattice The 2D square lattice connectivity.
+   * @param lattice The `LatticeGraph` connectivity.
    * @param num_spin_species Number of spin species (usually 2 for spinful, 1
    *        for spinless). Default: 2.
    * @return MajoranaMapping with name ``"verstraete-cirac"``.
--- a/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
+++ b/cpp/include/qdk/chemistry/data/majorana_mapping.hpp
@@ -105,11 +105,14 @@ class MajoranaMapping : public DataClass {
     return tapering_;
   }
 
-  /// Grid X dimension (for Verstraete-Cirac mapping).
-  std::size_t grid_nx() const { return grid_nx_; }
-
-  /// Grid Y dimension (for Verstraete-Cirac mapping).
-  std::size_t grid_ny() const { return grid_ny_; }
+  /**
+   * @brief Stabilizer terms carried by this mapping.
+   * @return Reference to the vector of stabilizers.
+   */
+  const std::vector<std::pair<std::complex<double>, SparsePauliWord>>&
+  stabilizers() const {
+    return stabilizers_;
+  }
 
   /**
    * @brief Return a copy with tapering removed and the base encoding name
@@ -170,19 +173,6 @@ class MajoranaMapping : public DataClass {
                                    const std::string& type);
 
   // --- Factory methods for standard encodings ---
-
-  /**
-   * @brief Verstraete-Cirac encoding.
-   *
-   * Maps fermionic modes on a 2D square lattice to qubits.
-   *
-   * @param lattice The 2D square lattice connectivity.
-   * @param num_spin_species Number of spin species (usually 2 for spinful, 1
-   * for spinless). Default: 2.
-   * @return MajoranaMapping with name ``"verstraete-cirac"``.
-   */
-  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice,
-                                          std::size_t num_spin_species = 2);
 
   /**
    * @brief Jordan-Wigner encoding.
@@ -267,6 +257,19 @@ class MajoranaMapping : public DataClass {
   static MajoranaMapping symmetry_conserving_bravyi_kitaev(
       std::size_t num_modes, std::size_t n_alpha, std::size_t n_beta);
 
+  /**
+   * @brief Verstraete-Cirac encoding.
+   *
+   * Maps fermionic modes on a 2D square lattice to qubits.
+   *
+   * @param lattice The 2D square lattice connectivity.
+   * @param num_spin_species Number of spin species (usually 2 for spinful, 1
+   *        for spinless). Default: 2.
+   * @return MajoranaMapping with name ``"verstraete-cirac"``.
+   */
+  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice,
+                                          std::size_t num_spin_species = 2);
+
  private:
   MajoranaMapping(
       std::vector<SparsePauliWord> table,
@@ -274,7 +277,8 @@ class MajoranaMapping : public DataClass {
       std::string name, std::size_t num_modes, std::size_t num_qubits,
       std::string base_encoding,
       std::optional<TaperingSpecification> tapering = std::nullopt,
-      std::size_t grid_nx = 0, std::size_t grid_ny = 0);
+      std::vector<std::pair<std::complex<double>, SparsePauliWord>>
+          stabilizers = {});
 
   /// Majorana-to-Pauli table (empty for bilinear-only mappings).
   std::vector<SparsePauliWord> table_;
@@ -302,6 +306,9 @@ class MajoranaMapping : public DataClass {
 
   /// Feed the mapping's identifying data into a content hash.
   void hash_update(qdk::chemistry::utils::HashContext& ctx) const override;
+
+  /// Optional stabilizer terms.
+  std::vector<std::pair<std::complex<double>, SparsePauliWord>> stabilizers_;
 
   /// Upper-triangle index: (j, k) with j < k, M = 2*num_modes.
   std::size_t bilinear_index(std::size_t j, std::size_t k) const {
--- a/cpp/src/qdk/chemistry/data/lattice_graph.cpp
+++ b/cpp/src/qdk/chemistry/data/lattice_graph.cpp
@@ -29,6 +29,19 @@ static void add_edge(std::vector<Triplet>& triplets, int i, int j, double t) {
   triplets.emplace_back(j, i, t);
 }
 
+/**
+ * @brief Depth-first search (DFS) helper to find a Hamiltonian path in a sparse
+ * graph.
+ *
+ * Recursively visits unvisited neighbor vertices to build a path that visits
+ * every vertex in the graph exactly once.
+ *
+ * @param curr    The vertex index currently being visited.
+ * @param adj     The sparse adjacency matrix of the graph.
+ * @param visited Tracks which vertices have already been visited.
+ * @param path    Stores the sequence of vertices in the current path.
+ * @return True if a Hamiltonian path is found, false otherwise.
+ */
 bool find_hamiltonian_path_dfs(std::uint64_t curr,
                                const Eigen::SparseMatrix<double>& adj,
                                std::vector<bool>& visited,
@@ -53,6 +66,16 @@ bool find_hamiltonian_path_dfs(std::uint64_t curr,
   return false;
 }
 
+/**
+ * @brief Search for a Hamiltonian path in the given sparse graph.
+ *
+ * Tries to find a path that visits every vertex exactly once, starting the
+ * search from each possible vertex in the graph.
+ *
+ * @param adj The sparse adjacency matrix representing the graph.
+ * @return A vector of vertex indices in path order, or an empty vector if no
+ * path exists.
+ */
 std::vector<std::uint64_t> find_hamiltonian_path(
     const Eigen::SparseMatrix<double>& adj) {
   std::uint64_t V = adj.rows();
@@ -203,7 +226,7 @@ std::uint64_t LatticeGraph::num_edges() const {
 }
 
 LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t,
-                                 bool optimal_ordering) {
+                                 bool dfs_ordering) {
   if (n == 0) {
     throw std::invalid_argument("chain: n must be > 0.");
   }
@@ -227,7 +250,7 @@ LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t,
   adj.makeCompressed();
   LatticeGraph g(std::move(adj),
                  chain_coloring(static_cast<std::int64_t>(N), periodic));
-  if (optimal_ordering) {
+  if (dfs_ordering) {
     auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
     if (!path.empty()) {
       return permute(g, path);
@@ -241,7 +264,7 @@ LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t,
 
 LatticeGraph LatticeGraph::square(std::uint64_t nx, std::uint64_t ny,
                                   bool periodic_x, bool periodic_y, double t,
-                                  bool optimal_ordering) {
+                                  bool dfs_ordering) {
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("square: nx and ny must be > 0.");
   }
@@ -286,7 +309,7 @@ LatticeGraph LatticeGraph::square(std::uint64_t nx, std::uint64_t ny,
   adj.makeCompressed();
   LatticeGraph g(std::move(adj),
                  square_coloring(Nx, Ny, periodic_x, periodic_y));
-  if (optimal_ordering) {
+  if (dfs_ordering) {
     auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
     if (!path.empty()) {
       return permute(g, path);
@@ -301,7 +324,7 @@ LatticeGraph LatticeGraph::square(std::uint64_t nx, std::uint64_t ny,
 LatticeGraph LatticeGraph::triangular(std::uint64_t nx, std::uint64_t ny,
                                       bool periodic_x, bool periodic_y,
                                       double t, int coloring_seed,
-                                      bool optimal_ordering) {
+                                      bool dfs_ordering) {
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("triangular: nx and ny must be > 0.");
   }
@@ -357,7 +380,7 @@ LatticeGraph LatticeGraph::triangular(std::uint64_t nx, std::uint64_t ny,
   // No known deterministic coloring for triangular lattices with arbitrary
   // periodic boundaries; use greedy with multiple trials instead.
   LatticeGraph g(std::move(adj), greedy_edge_coloring(adj, coloring_seed, 32));
-  if (optimal_ordering) {
+  if (dfs_ordering) {
     auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
     if (!path.empty()) {
       return permute(g, path);
@@ -371,8 +394,8 @@ LatticeGraph LatticeGraph::triangular(std::uint64_t nx, std::uint64_t ny,
 
 LatticeGraph LatticeGraph::honeycomb(std::uint64_t nx, std::uint64_t ny,
                                      bool periodic_x, bool periodic_y, double t,
-                                     bool optimal_ordering) {
-  (void)optimal_ordering;
+                                     bool dfs_ordering) {
+  (void)dfs_ordering;
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("honeycomb: nx and ny must be > 0.");
   }
@@ -425,8 +448,8 @@ LatticeGraph LatticeGraph::honeycomb(std::uint64_t nx, std::uint64_t ny,
 
 LatticeGraph LatticeGraph::kagome(std::uint64_t nx, std::uint64_t ny,
                                   bool periodic_x, bool periodic_y, double t,
-                                  int coloring_seed, bool optimal_ordering) {
-  (void)optimal_ordering;
+                                  int coloring_seed, bool dfs_ordering) {
+  (void)dfs_ordering;
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("kagome: nx and ny must be > 0.");
   }
--- a/cpp/src/qdk/chemistry/data/lattice_graph.cpp
+++ b/cpp/src/qdk/chemistry/data/lattice_graph.cpp
@@ -1055,7 +1055,7 @@ LatticeGraph LatticeGraph::permute(const LatticeGraph& graph,
   for (std::uint64_t i = 0; i < V; ++i) {
     P.indices()[static_cast<Eigen::Index>(i)] = static_cast<int>(path[i]);
   }
-  Eigen::SparseMatrix<double> new_adj = P * adj * P.transpose();
+  Eigen::SparseMatrix<double> new_adj = P.transpose() * adj * P;
   new_adj.makeCompressed();
 
   std::vector<std::uint64_t> inv_p(V);
--- a/cpp/src/qdk/chemistry/data/lattice_graph.cpp
+++ b/cpp/src/qdk/chemistry/data/lattice_graph.cpp
@@ -28,6 +28,44 @@ static void add_edge(std::vector<Triplet>& triplets, int i, int j, double t) {
   triplets.emplace_back(i, j, t);
   triplets.emplace_back(j, i, t);
 }
+
+bool find_hamiltonian_path_dfs(std::uint64_t curr,
+                               const Eigen::SparseMatrix<double>& adj,
+                               std::vector<bool>& visited,
+                               std::vector<std::uint64_t>& path) {
+  path.push_back(curr);
+  if (path.size() == static_cast<std::size_t>(adj.rows())) {
+    return true;
+  }
+  visited[curr] = true;
+
+  // Iterate directly over the sparse matrix columns/rows for neighbors
+  for (Eigen::SparseMatrix<double>::InnerIterator it(adj, curr); it; ++it) {
+    std::uint64_t neighbor = it.row();
+    if (neighbor != curr && !visited[neighbor]) {
+      if (find_hamiltonian_path_dfs(neighbor, adj, visited, path)) {
+        return true;
+      }
+    }
+  }
+  visited[curr] = false;
+  path.pop_back();
+  return false;
+}
+
+std::vector<std::uint64_t> find_hamiltonian_path(
+    const Eigen::SparseMatrix<double>& adj) {
+  std::uint64_t V = adj.rows();
+  std::vector<bool> visited(V, false);
+  std::vector<std::uint64_t> path;
+  for (std::uint64_t start = 0; start < V; ++start) {
+    if (find_hamiltonian_path_dfs(start, adj, visited, path)) {
+      return path;
+    }
+  }
+  return {};
+}
+
 }  // namespace detail
 
 LatticeGraph::LatticeGraph(
@@ -164,7 +202,8 @@ std::uint64_t LatticeGraph::num_edges() const {
   return count;
 }
 
-LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t) {
+LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t,
+                                 bool optimal_ordering) {
   if (n == 0) {
     throw std::invalid_argument("chain: n must be > 0.");
   }
@@ -186,12 +225,23 @@ LatticeGraph LatticeGraph::chain(std::uint64_t n, bool periodic, double t) {
   Eigen::SparseMatrix<double> adj(N, N);
   adj.setFromTriplets(triplets.begin(), triplets.end());
   adj.makeCompressed();
-  return LatticeGraph(std::move(adj),
-                      chain_coloring(static_cast<std::int64_t>(N), periodic));
+  LatticeGraph g(std::move(adj),
+                 chain_coloring(static_cast<std::int64_t>(N), periodic));
+  if (optimal_ordering) {
+    auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
+    if (!path.empty()) {
+      return permute(g, path);
+    } else {
+      throw std::runtime_error(
+          "No Hamiltonian path found in the lattice graph.");
+    }
+  }
+  return g;
 }
 
 LatticeGraph LatticeGraph::square(std::uint64_t nx, std::uint64_t ny,
-                                  bool periodic_x, bool periodic_y, double t) {
+                                  bool periodic_x, bool periodic_y, double t,
+                                  bool optimal_ordering) {
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("square: nx and ny must be > 0.");
   }
@@ -234,13 +284,24 @@ LatticeGraph LatticeGraph::square(std::uint64_t nx, std::uint64_t ny,
   Eigen::SparseMatrix<double> adj(N, N);
   adj.setFromTriplets(triplets.begin(), triplets.end());
   adj.makeCompressed();
-  return LatticeGraph(std::move(adj),
-                      square_coloring(Nx, Ny, periodic_x, periodic_y));
+  LatticeGraph g(std::move(adj),
+                 square_coloring(Nx, Ny, periodic_x, periodic_y));
+  if (optimal_ordering) {
+    auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
+    if (!path.empty()) {
+      return permute(g, path);
+    } else {
+      throw std::runtime_error(
+          "No Hamiltonian path found in the lattice graph.");
+    }
+  }
+  return g;
 }
 
 LatticeGraph LatticeGraph::triangular(std::uint64_t nx, std::uint64_t ny,
                                       bool periodic_x, bool periodic_y,
-                                      double t, int coloring_seed) {
+                                      double t, int coloring_seed,
+                                      bool optimal_ordering) {
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("triangular: nx and ny must be > 0.");
   }
@@ -295,13 +356,23 @@ LatticeGraph LatticeGraph::triangular(std::uint64_t nx, std::uint64_t ny,
   adj.makeCompressed();
   // No known deterministic coloring for triangular lattices with arbitrary
   // periodic boundaries; use greedy with multiple trials instead.
-  return LatticeGraph(std::move(adj),
-                      greedy_edge_coloring(adj, coloring_seed, 32));
+  LatticeGraph g(std::move(adj), greedy_edge_coloring(adj, coloring_seed, 32));
+  if (optimal_ordering) {
+    auto path = detail::find_hamiltonian_path(g.sparse_adjacency_matrix());
+    if (!path.empty()) {
+      return permute(g, path);
+    } else {
+      throw std::runtime_error(
+          "No Hamiltonian path found in the lattice graph.");
+    }
+  }
+  return g;
 }
 
 LatticeGraph LatticeGraph::honeycomb(std::uint64_t nx, std::uint64_t ny,
-                                     bool periodic_x, bool periodic_y,
-                                     double t) {
+                                     bool periodic_x, bool periodic_y, double t,
+                                     bool optimal_ordering) {
+  (void)optimal_ordering;
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("honeycomb: nx and ny must be > 0.");
   }
@@ -354,7 +425,8 @@ LatticeGraph LatticeGraph::honeycomb(std::uint64_t nx, std::uint64_t ny,
 
 LatticeGraph LatticeGraph::kagome(std::uint64_t nx, std::uint64_t ny,
                                   bool periodic_x, bool periodic_y, double t,
-                                  int coloring_seed) {
+                                  int coloring_seed, bool optimal_ordering) {
+  (void)optimal_ordering;
   if (nx == 0 || ny == 0) {
     throw std::invalid_argument("kagome: nx and ny must be > 0.");
   }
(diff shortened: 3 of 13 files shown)


python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Max Pauli weight of nearest-neighbor hops should be constant for grids of dimensions:
+        - 2x2 (L=2)
+        - 2x3 (L=3)
+        - 2x4 (L=4)
+        """
+        mapper = create("qubit_mapper", "qdk")
+
+        max_weights = []
+        for grid_length in [2, 3, 4]:
```
wavefunction91, line 303: Thanks for adding this scaling check. One thing to flag: the Pauli-weight criterion is on L×L lattices ("open L×L lattices with L in {2, 3, 4} … the same finite integer for all three L"), but this iterates 2×L (2×2, 2×3, 2×4). Could you switch it to 2×2, 3×3, 4×4 so it matches the criterion? That's the case the criterion is really after, so we'd need it before merging.

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,195 @@
...
+                if pauli_str == "I" * len(pauli_str):
+                    continue
+                # Weight is the number of non-I characters
+                weight = sum(1 for c in pauli_str if c != "I")
+                weights.append(weight)
+
+            max_weights.append(max(weights))
+
+        # Check that the maximum Pauli weight does not grow with L
+        assert max_weights[0] == max_weights[1] == max_weights[2], (
```
wavefunction91: Just a heads-up that on the L×L grids this needs to hold with the same integer across all three. If it doesn't once you switch, that points at the mapping rather than the test.

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,195 @@
...
+        assert mapping_3x3.grid_ny == 3
+
+        # Invalid spin species count should raise ValueError / invalid_argument
+        with pytest.raises(ValueError, match="spin"):
+            MajoranaMapping.verstraete_cirac(lattice_2x2, num_spin_species=0)
+
+    def test_general_lattices_single_spin(self) -> None:
+        """Verify that QubitMapper consumes VC mapping for general 2D lattices with single spin species."""
+        mapper = create("qubit_mapper", "qdk")
+        for nx, ny in [(2, 2), (2, 3), (3, 3), (4, 4)]:
```
wavefunction91: This is a nice coverage loop. Since the spectral criterion is about matching Jordan-Wigner on the codespace, could we add a spectral comparison (not just num_qubits) on at least the 3×3 here? "Runs without error" can still hide a wrong spectrum, so we'd want that check to be confident the criterion is met.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +335,161 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+  std::uint64_t V = lattice.num_sites();
+  std::uint64_t E = lattice.num_edges();
+
+  // Solve: nx * ny = V, nx + ny = 2V - E
+  double b = double(2 * V - E);
+  double c = double(V);
+  double disc = b * b - 4.0 * c;
+  if (disc < 0) {
+    throw std::invalid_argument(
+        "Lattice graph is not a 2D square lattice with open boundaries");
```
wavefunction91: The criterion asks for a general 2D lattice, and right now triangular/honeycomb lattices (which are also 2D) hit this error path, only row-major rectangular grids get through. Please generalize to accept a general 2D LatticeGraph.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +335,161 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+    const LatticeGraph& lattice, std::size_t num_spin_species) {
+  if (num_spin_species == 0) {
+    throw std::invalid_argument(
+        "verstraete_cirac requires num_spin_species > 0");
+  }
+
+  std::uint64_t V = lattice.num_sites();
+  std::uint64_t E = lattice.num_edges();
+
+  // Solve: nx * ny = V, nx + ny = 2V - E
```
wavefunction91: The challenge with inferring geometry from vertex/edge counts is that it sets aside the connectivity already in the `LatticeGraph` and only works for one specific labeling, which is what leads to the square-only restriction above. Could the encoding be built from the graph's edges directly instead? That would also make the general-2D-lattice criterion fall out naturally.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +335,161 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+  std::uint64_t nx = 0, ny = 0;
+  if (check_nx(nx1)) {
+    nx = nx1;
+    ny = V / nx1;
+  } else if (check_nx(nx2)) {
+    nx = nx2;
+    ny = V / nx2;
+  } else {
+    throw std::invalid_argument(
+        "Lattice graph is not a 2D square lattice with open boundaries");
```
wavefunction91: Could you confirm what happens with a periodic lattice like `LatticeGraph.square(3, 3, periodic_x=True)`? Periodic Fermi-Hubbard is one of the motivating cases in the issue, so it'd be good to know it's covered

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class TestVerstraeteCiracPauliWeightScaling:
+    """Verify that nearest-neighbor hop Pauli weight remains constant for grid lengths L in {2, 3, 4}."""
+
+    def test_pauli_weight_scaling(self) -> None:
+        """Check nearest-neighbor hopping terms in the mapped qubit Hamiltonian.
+
+        Max Pauli weight of nearest-neighbor hops should be constant for grids of dimensions:
+        - 2x2 (L=2)
```
wavefunction91: Small thing: the docstring lists 2×L; worth updating to L×L so it matches the criterion once the test changes.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +335,161 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+  std::uint64_t col;
+  if (row % 2 == 0) {
+    col = index % nx;
+  } else {
+    col = nx - 1 - (index % nx);
+  }
+  return {col, row};
+}
+
+bool is_vertical_edge(std::uint64_t i, std::uint64_t j, std::uint64_t nx,
```
wavefunction91: I think there's an indexing mismatch worth checking here:` is_vertical_edge` works in `snake_coordinates`, but `LatticeGraph::square` lays sites out row-major `(idx = y*nx + x)`. Tracing it through, the two conventions don't line up: on 3×3 only the center-column vertical edges get flagged, and on even `nx` (2×N and 4×4) none do, so the stabilizer dressing doesn't actually run on those grids. That's also why the 2×L test passes so easily. Could you add an assertion that the dressed-edge count equals the lattice's vertical-edge count on 3×3 and 4×4, and reconcile the two conventions?

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,195 @@
...
+        """Verify that Verstraete-Cirac mapping survives JSON and HDF5 serialization round-trips."""
+        lattice = LatticeGraph.square(2, 2)
+        mapping = MajoranaMapping.verstraete_cirac(lattice)
+
+        # JSON Round-trip
+        json_data = mapping.to_json()
+        loaded_json = MajoranaMapping.from_json(json_data)
+        assert loaded_json.name == mapping.name
+        assert loaded_json.grid_nx == mapping.grid_nx
+        assert loaded_json.grid_ny == mapping.grid_ny
```
wavefunction91: The serialization criterion is "the round-tripped mapping produces a QubitHamiltonian whose terms equal the original's term-by-term (Pauli string and coefficient)." Could you assert that directly: map before and after, compare Pauli strings + coefficients rather than checking grid fields? That's the actual requirement, and it won't depend on grid metadata (which I think should come off the data class; see the header comment).

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
```diff
@@ -101,6 +103,12 @@ class MajoranaMapping : public DataClass {
     return tapering_;
   }
 
+  /// Grid X dimension (for Verstraete-Cirac mapping).
+  std::size_t grid_nx() const { return grid_nx_; }
+
+  /// Grid Y dimension (for Verstraete-Cirac mapping).
+  std::size_t grid_ny() const { return grid_ny_; }
+
```
wavefunction91: `MajoranaMapping` is the general encoding type, and these fields bake one encoding's lattice geometry into it. Please remove these from the data class.

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -449,38 +457,159 @@ MajoranaMapResult majorana_map_impl(
...
+                                          mode_alpha(p), mode_alpha(q), eri);
             }
           }
         }
       }
     }
+  }
 
-    // αβ + βα cross-spin channels, related by Coulomb symmetry (pq|rs)=(rs|pq)
+  if (mapping.base_encoding() == "verstraete-cirac" && mapping.grid_nx() > 0 &&
```
wavefunction91: This is where the coupling starts: the engine matches on the encoding name and reads grid_nx/grid_ny off the mapping to rebuild the plaquette penalty , which is what forces grid metadata onto MajoranaMapping and hardcodes the square structure. The stabilizer/penalty terms should be carried by the mapping in a geometry-free form so the engine adds them generically, without special-casing "verstraete-cirac" or knowing about grids. That removes the leak and generalizes the construction at once.

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -449,38 +457,159 @@ MajoranaMapResult majorana_map_impl(
...
+        p = 2 * bot_aux_mode;
+        q = 2 * top_aux_mode + 1;
+      }
+
+      auto [coeff_stab, word_stab] = jw_base.bilinear(p, q);
+      return {coeff_stab, sparse_to_packed<NW>(word_stab)};
+    };
+
+    // Calculate number of plaquettes per spin sector: (nx - 1) * (ny - 1)
+    std::size_t num_plaquettes = (nx - 1) * (ny - 1);
```
wavefunction91: Plaquettes are hardcoded here because the penalty is rebuilt from grid dims (line 496). Same fix (deriving stabilizers from the graph) covers this; flagging so it's not missed.

python/src/pybind11/data/majorana_mapping.cpp
```diff
@@ -145,6 +146,8 @@ bilinear(j, k) is available on both forms.
       .def_property_readonly("num_qubits", &MajoranaMapping::num_qubits)
       .def_property_readonly("name", &MajoranaMapping::name)
       .def_property_readonly("base_encoding", &MajoranaMapping::base_encoding)
+      .def_property_readonly("grid_nx", &MajoranaMapping::grid_nx)
+      .def_property_readonly("grid_ny", &MajoranaMapping::grid_ny)
```
wavefunction91: These expose the grid fields to Python; they should come off once the members are removed from the data class (see header comment).

wavefunction91 (changes requested):

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
```diff
@@ -276,6 +298,10 @@ class MajoranaMapping : public DataClass {
   /// Optional tapering metadata for post-mapping qubit reduction.
   std::optional<TaperingSpecification> tapering_;
 
+  /// Optional grid dimensions for 2D mappings.
+  std::size_t grid_nx_ = 0;
```
agamshayit-ms: See above regarding general lattice geometries

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
nirajvenkat: Removed this and replaced it with a stabilizer object that is populated in the mapper. We are now topology agnostic, not relying on mapping onto a 2D grid. In theory, we should support the other `LatticeGraph` types, but we are not testing for that yet.

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
nirajvenkat: Plaquettes are now dynamically determined in the mapper, not here.

nirajvenkat: Derived from the stabilizers, which should be independent of the dimensions.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
nirajvenkat: We have improved our approach, now we find a Hamiltonian path dynamically directly from the graph adjacency.

python/tests/test_verstraete_cirac.py
nirajvenkat: Added a 3x3 Hückel (spinless) spectral comparison test.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
nirajvenkat: Yes, we have covered this in our Fermi-Hubbard test case now. Its sending in `periodic_x=True` when constructing the lattice.

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
```diff
@@ -244,13 +255,28 @@ class MajoranaMapping : public DataClass {
   static MajoranaMapping symmetry_conserving_bravyi_kitaev(
       std::size_t num_modes, std::size_t n_alpha, std::size_t n_beta);
 
+  /**
+   * @brief Verstraete-Cirac encoding.
+   *
+   * Maps fermionic modes on a 2D square lattice to qubits.
+   *
+   * @param lattice The 2D square lattice connectivity.
```
agamshayit-ms: The current implementation takes in a general LatticeGraph, so the "square lattice" comment no longer applies

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
```diff
@@ -206,6 +220,16 @@ MajoranaMapping MajoranaMapping::from_json(const nlohmann::json& data) {
   if (data.contains("tapering") && !data.at("tapering").is_null()) {
     tapering = TaperingSpecification::from_json(data.at("tapering"));
   }
+  std::vector<std::pair<std::complex<double>, SparsePauliWord>> stabilizers;
```
agamshayit-ms, line 263: It would be good to wrap this in a function (similar to TaperingSpecification::from_json())

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
```diff
@@ -173,6 +179,14 @@ nlohmann::json MajoranaMapping::to_json() const {
   if (tapering_) {
     data["tapering"] = tapering_->to_json();
   }
+  if (!stabilizers_.empty()) {
```
agamshayit-ms, line 227: It would be good it wrap this in a function (similar to tapering_->to_json())

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
```diff
@@ -219,21 +243,17 @@ MajoranaMapping MajoranaMapping::from_json(const nlohmann::json& data) {
     }
     auto mapping = MajoranaMapping::from_bilinears(
         num_modes, std::move(bilinears), base_encoding);
-    if (name == base_encoding && !tapering) {
```
agamshayit-ms, line 238: Why was this case removed?

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
                          std::move(tapering));
 }
 
+// ── Factory: Verstraete-Cirac ──────────────────────────────────────────
+
+MajoranaMapping MajoranaMapping::verstraete_cirac(
+    const LatticeGraph& lattice, std::size_t num_spin_species) {
+  using namespace detail;
+  const std::string name = "verstraete-cirac";
+  if (num_spin_species == 0) {
```
agamshayit-ms: in line 200 of [‎cpp/src/qdk/chemistry/data/majorana_map_engine.cpp], you only let `num_spin_species` be 0 or 1. You should also guard against `num_spin_species > 2` for consistency

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+  }
+  std::uint64_t V = lattice.num_sites();
+  if (V < 3) {
+    throw std::invalid_argument(
+        name + " requires a lattice graph with at least 3 sites");
+  }
+
+  // Find all edges in the lattice (upper triangle of adjacency matrix)
+  std::vector<std::vector<std::uint64_t>> adj_list(V);
+  const auto& adj = lattice.sparse_adjacency_matrix();
```
agamshayit-ms, line 353: Why can't `adj` be used directly without converting to `adj_list`?

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+      std::uint64_t u = it.row();
+      std::uint64_t v = it.col();
+      if (u != v) {
+        adj_list[u].push_back(v);
+      }
+    }
+  }
+
+  // Find Hamiltonian path dynamically directly from the graph adjacency
+  std::vector<std::uint64_t> snake_path = find_hamiltonian_path(adj_list);
```
agamshayit-ms: Since the current implementation works for general 2D lattices, the name `snake_path` doesn't make sense anymore

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+  if (snake_path.empty()) {
+    throw std::invalid_argument(
```
agamshayit-ms: `invalid_argument` is not the correct exception to throw here

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+        name +
+        ": No Hamiltonian path found in the lattice graph. A Hamiltonian path "
+        "is required to construct commuting stabilizers.");
+  }
+
+  // Maps row-major site index to its position in the snake path (system mode
+  // order)
+  std::vector<std::uint64_t> rm_to_snake(V);
```
agamshayit-ms: All variables names containing "snake" should be renamed to reflect applicability to general 2D lattices

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+        }
+      }
+    }
+  }
+
+  // Ensure that no vertex has more than 2 non-path edges (auxiliary space
+  // limit)
+  for (std::uint64_t u = 0; u < V; ++u) {
+    if (non_path_incident[u].size() > 2) {
+      throw std::invalid_argument(name + ": Vertex has " +
```
agamshayit-ms: Should be runtime error

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -450,37 +459,67 @@ MajoranaMapResult majorana_map_impl(
...
+        if (num_spin_species == 2) {
+          if (!spin_symmetric) {
+            h1_sum += std::abs(h1_beta[p * n_spatial + q]);
+          } else {
+            h1_sum += std::abs(h1_alpha[p * n_spatial + q]);
           }
         }
       }
     }
+    double aux_ham_coefficient = 1.0 + h1_sum;
```
agamshayit-ms, line 732: I would be good to comment on this choice of `aux_ham_coefficient`

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -450,37 +459,67 @@ MajoranaMapResult majorana_map_impl(
...
+
+    PackedPauliWord<NW> identity{};
+    acc.accumulate(
+        identity, std::complex<double>(
+                      aux_ham_coefficient * mapping.stabilizers().size(), 0.0));
+
+    for (const auto& [coeff, word] : mapping.stabilizers()) {
```
agamshayit-ms, line 743: This implements a penalty term for each stabilizer explicitly used in the mapping. While correct, it does not pin down the sector of the remaining ("unpaired") auxiliary Majorana operators, leading to a degenerate codespace. https://arxiv.org/pdf/cond-mat/0508353 fixes this by including "gauge conditions" in the auxiliary Hamiltonian, which fix the values of these unpaired auxiliary modes.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +369,182 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+                                non_path_incident[u].end(), v);
+          auto it_v = std::find(non_path_incident[v].begin(),
+                                non_path_incident[v].end(), u);
+          std::size_t a = std::distance(non_path_incident[u].begin(), it_u);
+          std::size_t b = std::distance(non_path_incident[v].begin(), it_v);
+
+          std::size_t p = get_aux_majorana(u, spin, a);
+          std::size_t q = get_aux_majorana(v, spin, b);
+
+          auto [coeff, word] = get_bilinear(p, q);
```
agamshayit-ms, line 437: Look at Fig. 2 of [https://arxiv.org/pdf/cond-mat/0508353]. The stabilizer graph defining the `P_{ij}` should form loop structures. The auxiliary Hamiltonian is then formed out of Plaquettes of `P_{ij}`. This is missing from the current implementation

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
nirajvenkat: Added a comment which is a short version of the following:

If stabilizers are included in the mapping, our total Hamiltonian becomes $H_{\rm total} = H_{\rm physical} + H_{\rm aux}$. We add $H_{\rm aux} = λ · Σ_i (I − S_i)$ to penalize unphysical codespace sectors. In the physical sector all $S_i = +1$, so $H_{\rm aux} = 0$ and the spectrum is unchanged. In any unphysical sector at least one $S_i = −1$, contributing an energy penalty of $2λ$ that pushes those states above the physical ground state. For this to be guaranteed, $λ$ must exceed the spectral range of $H_{\rm physical}$ across sectors. We use $λ = 1 + ‖h_1‖_1$ as a practical lower bound; the true sufficient condition would require the full 1-norm of the mapped Hamiltonian, which is unavailable at this stage.

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
nirajvenkat: We now identify and pair up any remaining unused aux Majorana modes to lift boundary/corner codespace degeneracy.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
nirajvenkat: Done, its now clear we are not using snake coords.

cpp/src/qdk/chemistry/data/majorana_mapping.cpp
nirajvenkat: I forgot to add this back after tinkering around with the code. Thanks for flagging this, added an extra condition here.

nirajvenkat:
I will leave comments unresolved so that code reviewers can mark resolved. Not sure what your convention is.

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
```diff
@@ -244,13 +255,28 @@ class MajoranaMapping : public DataClass {
...
+   *
+   * Maps fermionic modes on a `LatticeGraph` to qubits.
+   *
+   * @param lattice The `LatticeGraph` connectivity.
+   * @param num_spin_species Number of spin species (usually 2 for spinful, 1
+   *        for spinless). Default: 2.
+   * @return MajoranaMapping with name ``"verstraete-cirac"``.
+   */
+  static MajoranaMapping verstraete_cirac(const LatticeGraph& lattice,
+                                          std::size_t num_spin_species = 2);
```
wavefunction91: The mapping engine is shared infrastructure for every encoding, and I'd like it to keep one strict, global contract: a mapping is spinful with `2·n_spatial modes`. Inferring `num_spin_species` here from `num_modes < 2·n_spatial` breaks that in two ways: it teaches the general engine a spin model, and it silently reinterprets any undersized (i.e. incorrectly sized) mapping as "spinless" instead of raising, which removes a useful invariant. We're spin-1/2-only here. Please drop this inference and the downstream spinless branches so the engine requires the full `2·n_spatial` mode count. With that, the VC factory's `num_spin_species` parameter and the spinless (`num_spin_species=1`) tests fall away too, since they can no longer round-trip through the engine

When making this change, please take this as a global comment - any logic that was modified to accommodate non spin-1/2 fermions should be reverted.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
```diff
@@ -333,4 +372,203 @@ MajoranaMapping MajoranaMapping::symmetry_conserving_bravyi_kitaev(
...
+          non_path_incident[u].push_back(v);
+          non_path_incident[v].push_back(u);
+        }
+      }
+    }
+  }
+
+  // Ensure that we have no more than 2 aux Majorana modes per site
+  for (std::uint64_t u = 0; u < V; ++u) {
+    if (non_path_incident[u].size() > 2) {
```
wavefunction91: This cap is on off-path edges per site, which is a property of the Hamiltonian path you happen to find, not of the graph. I verified this: the 3×3 grid admits 40 Hamiltonian paths, and the max off-path degree ranges [2..3] across them — e.g. the valid path [0,1,2,5,8,7,6,3,4] ends at the degree-4 center and yields 3 off-path edges, which throws. So the same lattice is accepted or rejected based purely on DFS iteration order; the current tests pass only because the ascending-order DFS happens to land on a corner endpoint (off-path = 2).
 
 The cap itself = textbook VC's one auxiliary fermion per site (hence exactly 2 aux Majoranas). The faithful VC extension to arbitrary lattices is to let the per-site auxiliary count follow that site's off-backbone degree (d−2 for an interior path site). That is still genuinely VC:  JW backbone + site auxiliaries cancelling off-backbone strings, the Whitfield–Havlíček–Troyer generalization — not superfast (edge ancillas) or Derby–Klassen (vertex+face), so it stays within the encoding family we want here. It removes the cap and, as a bonus, makes acceptance path-independent, which dissolves the DFS-order bug above.
 
 I understand this touches the hardcoded base_modes = 2·num_modes (fixed 1 aux/site) assumption, so it's bigger than a one-liner. But I don't think the fixed cap should ship as-is: the path-dependence is a correctness/robustness issue now, not just a scope limit. At minimum acceptance must be a deterministic property of the graph; the principled fix is the degree-scaled auxiliary count.

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,280 @@
...
+
+        qh_hdf5 = mapper.run(ham, loaded_hdf5)
+        assert qh_hdf5.num_qubits == qh_orig.num_qubits
+        assert len(qh_hdf5.pauli_strings) == len(qh_orig.pauli_strings)
+        terms_hdf5 = sorted(zip(qh_hdf5.pauli_strings, qh_hdf5.coefficients, strict=False))
+        for (p_orig, c_orig), (p_h5, c_h5) in zip(terms_orig, terms_hdf5, strict=False):
+            assert p_orig == p_h5
+            assert np.isclose(c_orig, c_h5, atol=1e-10)
+
+    def test_stabilizers_and_commutation(self) -> None:
```
wavefunction91: Every test input is `LatticeGraph.square`, so graph-generality is asserted but never exercised. Two asks, tied to the cap discussion above:
 
 (1) Now, within current limits: add a non-grid lattice that still satisfies the ≤2 off-backbone-degree cap: `LatticeGraph.honeycomb(...)` (degree-3) is a good fit. This proves the code paths aren't secretly coupled to a square grid, independent of the cap fix.
 
 (2) After the cap is lifted (degree-scaled auxiliaries): add a genuinely higher-connectivity lattice: `LatticeGraph.triangular(...)` or `kagome(...)` (degree up to 6). These currently throw on the L460 cap, so they're exactly the test that demonstrates the degree-scaled-aux generalization works and that acceptance no longer depends on the path. I'd treat this as the acceptance test for "VC on arbitrary lattices."

docs/source/user/comprehensive/algorithms/qubit_mapper.rst
```diff
@@ -52,6 +52,10 @@ Symmetry-conserving Bravyi-Kitaev :cite:`Bravyi2017tapering`
 Bravyi-Kitaev tree :cite:`Havlicek2017`
    A tree-based variant of the Bravyi-Kitaev transformation that uses a different qubit indexing strategy.
 
+.. _encoding-verstraete-cirac:
+
+Verstraete-Cirac :cite:`Verstraete2005`
+   An auxiliary-qubit encoding designed for 2D lattices. It eliminates non-local Z-strings by introducing flag qubits that locally track parity. Max Pauli weight of nearest-neighbor hops remains constant regardless of lattice size. Use :meth:`~qdk_chemistry.data.MajoranaMapping.verstraete_cirac` with a :class:`~qdk_chemistry.data.LatticeGraph`.
```
wavefunction91: Describe VC generically: it operates on a general LatticeGraph, not specifically "2D lattices." Once the auxiliary count is degree-scaled, the only precondition is a connected graph admitting a suitable backbone; phrase it that way, with 2D grids as the motivating example rather than the scope. Also, the "constant Pauli weight regardless of lattice size" claim should be backed by the L×L Pauli-weight test as written: please keep them in sync so the doc doesn't assert a property the tests don't demonstrate.

wavefunction91 (changes requested):
Thanks for the continued iteration

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
nirajvenkat: I have reverted, agreed it feels forcibly retrofitted into the engine and it requires too much testing coverage. But theres a reason this was implemented. Theres no other way to support 3x3 spectral tests, we will run into a RuntimeError during SciPy sparse matrix creation ($2^{32} \times 2^{32}$ dims which exceeds 32-bit size). So we must accept the compromise that the 3x3 test is now 3x2.

cpp/src/qdk/chemistry/data/majorana_mapping_factories.cpp
nirajvenkat: Thanks for testing and explaining what's wrong. We have new indexing to accommodate each site having $[0, d-2]$ aux modes.

python/tests/test_verstraete_cirac.py
nirajvenkat: `LatticeGraph.honeycomb` and `LatticeGraph.kagome` will only work if its a torus (`periodic_x=True, periodic_y=True`). Tested without those boundary conditions and no Hamiltonian path is found.

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
agamshayit-ms: you can modify the test to use qdk/chemistry's sparse matrices, which do not have this 32-bit size issue

python/tests/test_verstraete_cirac.py
nirajvenkat: There might be some graph theory reasons behind it as well.

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
nirajvenkat: @v-agamshayit We are indeed using sparse matrices throughout `test_spectral_validation_3x3_huckel`, please elaborate if I misunderstood. When creating the mask in `pauli_matrix.py` we can change the mask dtype to `np.int64` to avoid a RuntimeError. But since the mask shares its dtype with the downstream CSR matrix creation, we run out of memory during the 3x3 test and it shows "Killed" on my system.

python/tests/test_verstraete_cirac.py
wavefunction91: This is an interesting point, and I'm glad you did some experimentation here and came back to us with results. This actually gets at a deeper issue that I hadn't noticed before, because when I've looked at VC in the past, I've always thought about it in terms of Hamiltonian paths (as you did here).

In fact, VC has no requirement on index ordering: Hamiltonian paths only minimize the number of non-local terms, and thus the number of stabilizers (and ancilla) needed to cancel long parity strings. I really appreciate that you added the Hamiltonian path code for optimality, but it's not strictly needed for correctness.

What I suggest is the following:
1. Verify that your code works with an arbitrary graph ordering - this will generate more stabilizers and ancilla, but it should still be correct. In other words, it should just work with the default ordering the standard lattice factories already give you. Now that the ancilla allocation is dynamic I think this mostly falls out for free - drop the mandatory path search, run off the default ordering, and throw a lattice with no Hamiltonian path (kagome 3x3 is a good one) into the tests so we don't regress.
2. Add an optional argument on the lattice factories to produce a Hamiltonian-path ordering - I wouldn't even be opposed to leaving that option off for kagome and honeycomb. It's a nice optimization (fewer non-local terms, so fewer stabilizers and ancilla), but it's an ordering concern, so I'd rather it live with the lattices than inside VC.

One practical note so you don't end up fighting the test suite: keep using the Hamiltonian-path ordering in the spectral and Pauli-weight tests. Those are really exercising the optimization - the spectral ones need to stay small enough to diagonalize, and constant Pauli weight is a property of a good ordering in the first place - and since correctness doesn't depend on the ordering you don't lose anything by leaving them on the Hamiltonian path. Just use the default ordering for the generality/commutation tests, and add a no-Hamiltonian-path lattice like kagome 3x3 there to confirm it builds and the stabilizers still commute without throwing. Heads up that a 3x2 spectral case on the default ordering jumps from roughly 2^20 to 2^24 in matrix dimension, so it'll probably hit the same RuntimeError you saw on 3x3 - another reason to keep those particular tests on the Hamiltonian path.


If you *want* to add a more optimal, but always present ordering, something like Reverse Cuthill McGee would fit here, but I won't gate the PR on this.

nirajvenkat: Interesting to hear that the contiguity of the path is not really a necessary condition for VC. But as you rightly pointed out it is the optimal condition.

Went ahead and added kagome 3x3 test and also simplified the testing module.

The DFS traversal for Hamiltonian path construction is moved to the lattices now, and this has the side effect of simplifying the VC code as well since whatever path we get from the lattice is ordered now.

cpp/include/qdk/chemistry/data/lattice_graph.hpp
```diff
@@ -232,7 +232,7 @@ class LatticeGraph : public DataClass {
    * @throws std::invalid_argument If n == 0.
    */
   static LatticeGraph chain(std::uint64_t n, bool periodic = false,
-                            double t = 1.0);
+                            double t = 1.0, bool optimal_ordering = false);
```
wavefunction91: Let's not call this `optimal_ordering` - optimal is definitive, and may be different depending on the downstream usage.

cpp/src/qdk/chemistry/data/lattice_graph.cpp
```diff
@@ -28,6 +28,44 @@ static void add_edge(std::vector<Triplet>& triplets, int i, int j, double t) {
   triplets.emplace_back(i, j, t);
   triplets.emplace_back(j, i, t);
 }
+
+bool find_hamiltonian_path_dfs(std::uint64_t curr,
```
wavefunction91, line 45: Even though this is not public context, it would be good to get proper documentation for this.

nirajvenkat:
@wavefunction91 Any idea why the builds are failing? I am guessing this is not a testing issue, something related to permissions?

wavefunction91:
@nirajvenkat yes, that's what it seems. We're looking into it.

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,275 @@
...
+from qdk_chemistry.algorithms import create
+from qdk_chemistry.data import LatticeGraph, MajoranaMapping, QubitHamiltonian
+from qdk_chemistry.utils.model_hamiltonians import create_hubbard_hamiltonian, create_huckel_hamiltonian
+
+
+class TestVerstraeteCiracMapping:
+    """Tests covering the Verstraete-Cirac mapping factory, dimensions, and properties."""
+
+    def test_factory_and_dimensions(self) -> None:
+        """Test that VC factory accepts correct grid dimensions and rejects invalid ones."""
```
wavefunction91: What are you intending to reject here?

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,275 @@
...
+        qh_json = mapper.run(ham, loaded_json)
+        assert qh_json.num_qubits == qh_orig.num_qubits
+        assert len(qh_json.pauli_strings) == len(qh_orig.pauli_strings)
+        terms_orig = sorted(zip(qh_orig.pauli_strings, qh_orig.coefficients, strict=False))
+        terms_json = sorted(zip(qh_json.pauli_strings, qh_json.coefficients, strict=False))
+        for (p_orig, c_orig), (p_json, c_json) in zip(terms_orig, terms_json, strict=False):
+            assert p_orig == p_json
+            assert np.isclose(c_orig, c_json, atol=1e-10)
+
+        # HDF5 Round-trip
```
wavefunction91, line 104: Can this be a separate test?

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,275 @@
...
+        def commute(p1: str, p2: str) -> bool:
+            assert len(p1) == len(p2)
+            anti_commutes = 0
+            for c1, c2 in zip(p1, p2, strict=False):
+                if c1 != "I" and c2 not in {"I", c1}:
+                    anti_commutes += 1
+            return (anti_commutes % 2) == 0
+
+        mapper = create("qubit_mapper", "qdk")
+        lattices = [
```
wavefunction91: One might consider to make this a parameterized pytest.

wavefunction91 (changes requested):
Mostly cleanup, this is looking pretty close to merging. Ping @v-agamshayit to resolve comments or comment further.

python/src/qdk_chemistry/algorithms/qubit_mapper/qdk_qubit_mapper.py
```diff
@@ -143,10 +143,14 @@ def _run_impl(
         n_spin_orbitals = 2 * n_spatial
         spin_symmetric = hamiltonian.get_orbitals().is_restricted()
 
-        if base_mapping.num_modes != n_spin_orbitals:
+        allowed_modes: tuple[int, ...] = (
+            (n_spin_orbitals,) if hamiltonian.has_two_body_integrals() else (n_spatial, n_spin_orbitals)
+        )
```
wavefunction91: This relaxes the mode check to admit n_spatial (spinless) modes for one-body-only Hamiltonians, but the C++ engine in this same PR went the other way and is now strict — majorana_map_impl throws if mapping.num_modes() != 2 * n_spatial (majorana_map_engine.cpp:200). So a one-body Hamiltonian with an n_spatial-mode mapping passes this Python guard and then throws in C++ — the two checks contradict each other. VC always produces 2V = n_spin_orbitals modes, so the n_spatial branch isn't exercised by anything here. Please revert this to the strict n_spin_orbitals-only check so the Python and C++ contracts agree.

docs/source/user/comprehensive/algorithms/qubit_mapper.rst
```diff
@@ -52,6 +52,10 @@ Symmetry-conserving Bravyi-Kitaev :cite:`Bravyi2017tapering`
 Bravyi-Kitaev tree :cite:`Havlicek2017`
    A tree-based variant of the Bravyi-Kitaev transformation that uses a different qubit indexing strategy.
 
+.. _encoding-verstraete-cirac:
+
+Verstraete-Cirac :cite:`Verstraete2005`
+   An auxiliary-qubit encoding that operates on a general connected :class:`~qdk_chemistry.data.LatticeGraph` that admits a Hamiltonian path. It eliminates non-local Z-strings by introducing auxiliary qubits that locally track parity. For 2D square grids, the maximum Pauli weight of nearest-neighbor hopping terms remains constant regardless of the grid size. Use :meth:`~qdk_chemistry.data.MajoranaMapping.verstraete_cirac` with a :class:`~qdk_chemistry.data.LatticeGraph`.
```
wavefunction91: "that admits a Hamiltonian path" is no longer accurate — VC now runs on the lattice's default ordering, which works for any connected graph (kagome 3×3, which has no Hamiltonian path, is in the test suite for exactly this reason). The Hamiltonian path is only needed for the optional optimal_ordering optimization. Suggest something like: "operates on any connected LatticeGraph; an optional Hamiltonian-path reordering (optimal_ordering=True on the lattice factories) minimizes the number of auxiliary qubits and stabilizers."

cpp/src/qdk/chemistry/data/lattice_graph.cpp
```diff
@@ -949,4 +1021,38 @@ void LatticeGraph::hash_update(qdk::chemistry::utils::HashContext& ctx) const {
   hash_value(ctx, _is_symmetric);
 }
 
+LatticeGraph LatticeGraph::permute(const LatticeGraph& graph,
```
wavefunction91, line 1047: permute is new and has no unit test, and it remaps the adjacency with P*adj*Pᵀ (built from path) but the edge coloring with inv_p — two different directions of the same permutation. If those conventions don't agree, the coloring silently misaligns with the adjacency after a permute. VC won't catch this (it doesn't use the coloring), but downstream consumers of the edge coloring (e.g. Trotter edge ordering) would get wrong colors with no error. Could you add a small permute unit test that (a) asserts are_connected(i,j) after permute matches the remapped coloring keys, and (b) confirms optimal_ordering=True actually yields path-consecutive adjacency? That locks down the convention on both halves.

python/tests/test_verstraete_cirac.py
```diff
@@ -0,0 +1,275 @@
...
+            vc_mapping = MajoranaMapping.verstraete_cirac(lattice)
+            qh_vc = mapper.run(ham, vc_mapping)
+
+            weights = []
+            for pauli_str, coeff in zip(qh_vc.pauli_strings, qh_vc.coefficients, strict=False):
+                if pauli_str == "I" * len(pauli_str):
+                    continue
+                # Skip stabilizer penalty terms (which have coefficient >= 9.0,
+                # whereas nearest-neighbor hops have coefficient <= 1.0)
+                if np.abs(coeff) > 2.0:
```
agamshayit-ms: Can you explain this `> 2.0` and the values in the comments?

cpp/src/qdk/chemistry/data/lattice_graph.cpp
nirajvenkat: Added a new C++ test `LatticeGraphTest.Permute`

python/tests/test_verstraete_cirac.py
nirajvenkat: Split the serialization into two separate tests.

python/tests/test_verstraete_cirac.py
nirajvenkat: Added a rejection test that requires the exception in majorana_mapping_factories.cpp:346 to be thrown.

cpp/tests/test_lattice_graph.cpp
```diff
@@ -629,3 +629,59 @@ TEST_F(LatticeGraphTest, KagomeColoringSeed) {
...
+  // (0,3), (1,4), (2,5) [vertical]
+  auto sq = LatticeGraph::square(3, 2, false, false);
+  ASSERT_TRUE(sq.edge_coloring().has_value());
+
+  // Define a permutation path: e.g., [0, 3, 4, 1, 2, 5]
+  std::vector<std::uint64_t> path = {0, 3, 4, 1, 2, 5};
+  auto permuted = LatticeGraph::permute(sq, path);
+
+  EXPECT_EQ(permuted.num_sites(), sq.num_sites());
+  EXPECT_EQ(permuted.num_edges(), sq.num_edges());
```
wavefunction91, line 650: This test can't actually catch a `permute` direction bug, which is the one thing it exists to verify. The explicit-path block below only checks that the permuted adjacency and permuted coloring are self-consistent, and a wrong-direction permute passes that, since both get permuted the same way. And the `dfs_ordering` check at the bottom is vacuous here too: the 3×3 Hamiltonian path is `[0,1,2,5,4,3,6,7,8]`, whose inverse equals itself (an involution), so wrong-direction and correct are identical for that input. I confirmed a direction-inverted `permute` passes both checks.

Please add an independent assertion against the source graph,  new vertex `i` must correspond to old vertex `path[i]`:
 
```cpp
for (std::uint64_t i = 0; i < 6; ++i) {
  for (std::uint64_t j = i + 1; j < 6; ++j) {
    EXPECT_EQ(permuted.are_connected(i, j), sq.are_connected(path[i], path[j]));
  }
}
```
 
The implementation itself is correct (`new_adj[i,j] == old_adj[path[i],path[j]]`), so this is purely hardening the test so it can't silently pass on a future direction regression.

wavefunction91 (changes requested):

python/tests/test_verstraete_cirac.py
nirajvenkat: There was actually an error in the calculation, fixed in the new comment.

cpp/include/qdk/chemistry/data/majorana_mapping.hpp
agamshayit-ms: This could in principle be circumvented by utilizing a sparse eigensolver (like Davidson) but we won't require this here

python/tests/test_verstraete_cirac.py
agamshayit-ms: Thanks for the clarification and correction. The comment is still a bit vague (why 17.0?). It would be good to symbolically define this threshold variable using $\lambda$ and $t$ for clarity rather than use a hardcoded numerical value. Also, if physical terms have coefficients $t/2=1/2$ and stabilizer terms have $\lambda=1 + \text{positive}$. wouldn't 1.0 be a sufficient threshold?

wavefunction91: @nirajvenkat this comment is the last outstanding.

agamshayit-ms: Another note: The naive penalty Hamiltonian is indeed nonlocal. in the [paper](https://arxiv.org/pdf/cond-mat/0508353), they remedy this by modifying the penalty Hamiltonian to be a sum of products of stabilizers $P \mapsto P\times P$ (yielding plaquettes of auxiliary Majoranas). Because stabilizers commute, this preserves the ground state of the auxiliary sector. You can generalize this construction to arbitrary lattices to achieve auxiliary-sector locality

nirajvenkat: Right we have fully resolved this by going further into the VC paper. No more hardcoded thresholds.

We find the better way to implement the aux Hamiltonian. We still use the L1 norm as the weighting factor, but are now using boundary link stabilizers to lift the $\mathbb{Z}_2$ gauge degeneracy and products of adjacent stabilizers to keep the overall term local.

Everything is tested on multiple lattice topologies.

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -502,18 +502,68 @@ MajoranaMapResult majorana_map_impl(
...
+        const auto& s2 = packed_stabs[j];
+        auto [phase_idx, word] = multiply_packed(s1.word, s2.word);
+        std::complex<double> prod_coeff =
+            apply_phase(phase_idx, s1.coeff * s2.coeff);
+        acc.accumulate(word, -aux_ham_coefficient * prod_coeff);
+      };
+      // S[0] * S[0..n/2] for α spin Majoranas
+      acc.accumulate(packed_stabs[0].word,
+                     -aux_ham_coefficient * packed_stabs[0].coeff);
+      for (std::size_t i = 0; i < half_stabs - 1; ++i) {
```
agamshayit-ms: This multiplies stabilizers that are consecutive in array index, but that does not guarantee the geometric nesting relationship needed for Z-string cancellation. Only nested pairs - where one edge's path interval $[u_1,v_1]$ fully contains the other's $[u_2,v_2]$ - give bounded-weight products.

**Counterexample (3×3 square lattice, DFS ordering):**
The factory emits edge stabilizers in order of ascending site $u$
Index | Edge | Path interval | Weight
0        |  (0,5) |       [0, 5]       | 10
1        |  (1,4) |       [1, 4]       | 6
2        |  (3,8) |       [3, 8]       | 10
3        |  (4,7) |       [4, 7]       | 6


S[0] · S[1] = $S_{0,5} S_{1,4}$: nested ([1,4] ⊂ [0,5]), Z-strings cancel fully, weight=6
S[1] · S[2] = $S_{1,4} S_{3,8}$: crossing ([1,4] ∩ [3,8] = [3,4], only 1 qubit overlap), weight > 6

Fix: Instead of consecutive-index products, for each stabilizer $S_i$, find the stabilizer $S_j$ such that $S_i\times S_j$ has the minimum weight (can be easily computed via multiply_packed -> popcount on the result.)

nirajvenkat: I believe we could use a minimum spanning tree for this?

python/tests/test_verstraete_cirac.py
```diff
@@ -264,6 +265,39 @@ def check_local_plaquette(
                         f"Local plaquette stabilizer {p_str} has weight {w} > {max_weight}, violating local scaling"
                     )
 
+        def get_mst_edges(stabs_list: list[str], start_idx: int, count: int) -> list[tuple[int, int]]:
+            if count <= 1:
+                return []
+
+            in_mst = [False] * count
+            min_weight = [1000000000] * count
```
agamshayit-ms: Is this "infinity" for Prin's algorithm? can use `float('inf')` for clarity

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
```diff
@@ -768,19 +776,61 @@ MajoranaMapResult majorana_map_impl(const MajoranaMapping& mapping,
...
 
-      // S[n/2] * S[n/2..] for β spin Majoranas
-      acc.accumulate(packed_stabs[half_stabs].word,
-                     -aux_ham_coefficient * packed_stabs[half_stabs].coeff);
-      for (std::size_t i = half_stabs; i < num_stabs - 1; ++i) {
-        accumulate_product(i, i + 1);
-      }
+      auto build_mst_and_accumulate = [&](std::size_t start_idx,
+                                          std::size_t count) {
+        // Accumulate the root boundary link stabilizer
```
agamshayit-ms, line 795: The root is always `packed_stabs[start_idx]` (first in array order = the longest stabilize). Consider picking the stabilizer with minimum individual Pauli weight as the root instead, that way the boundary term is weight-minimized. 
Replace 
```
// Accumulate the root boundary link stabilizer
acc.accumulate(packed_stabs[start_idx].word,
               -aux_ham_coefficient * packed_stabs[start_idx].coeff);
```
with something like
```
// Find the lightest stabilizer as root
std::size_t root = 0;
int root_weight = get_packed_weight(packed_stabs[start_idx].word);
for (std::size_t i = 1; i < count; ++i) {
    int w = get_packed_weight(packed_stabs[start_idx + i].word);
    if (w < root_weight) { root = i; root_weight = w; }
}
```

cpp/src/qdk/chemistry/data/majorana_map_engine.cpp
agamshayit-ms: Great solution!

nirajvenkat:
@wavefunction91 Looks like all checks are passing! Are we good on changes requested?

wavefunction91 (approved):
lgtm, thanks for all the iteration on this, it's really great solution!

Issue #483, opened 2026-05-18
Add Zassenhaus Recursion for Hamiltonian Simulation


wavefunction91:
**Is your feature request related to a problem? Please describe.**

 QDK/Chemistry's `HamiltonianUnitaryBuilder` currently offers only Trotter-Suzuki product formulas for constructing exp(-iHt). For a target accuracy epsilon, Trotter-Suzuki of order p requires a step count whose leading error term scales as O(t^(p+1)/N^p); the prefactor is governed by nested commutators of the Hamiltonian summands but is not made explicit in the formula -- those commutators are absorbed into the step count rather than cancelled directly. For Hamiltonians whose dominant commutators are sparse or analytically tractable (lattice models with limited geometric coupling, chemistry Hamiltonians under term partitions with small inter-group commutators), this wastes steps.

 **Describe the solution you'd like**

 A new `HamiltonianUnitaryBuilder` implementation that uses a **Zassenhaus expansion** -- the dual of the Baker-Campbell-Hausdorff formula -- to construct exp(-iHt) as a product of exponentials in which low-order commutator corrections appear as explicit factors, rather than being controlled implicitly by the step count.

 The output must remain a `UnitaryRepresentation` consumable by existing downstream algorithms (e.g. `PhaseEstimation`) without changes to their interfaces.

 **Describe alternatives you've considered**

 - **Higher-order Trotter-Suzuki.** Already available; the recursive Suzuki construction has an exponentially growing step cost in the order, and its error prefactor is implicit.
 - **qDRIFT / randomized compilation.** Solves a different problem (scaling with the L1 norm of coefficients rather than with commutator structure); useful for very long Hamiltonians but does not exploit commutator sparsity.

 **Acceptance criteria**

 - A new factory registered under `HamiltonianUnitaryBuilder` (discoverable via the standard registry mechanism) accepts any `QubitHamiltonian` that the existing `"trotter"` factory accepts and returns a `UnitaryRepresentation` of the same shape (a `PauliProductFormulaContainer` with a repetition count).
[...]

PR #508, opened 2026-06-04
Zassehause evolution

neogyk:


wavefunction91 (changes requested):
The goal of #483 is to add a QDK-native implementation of Zassenhaus recursion. The current implementation serves as a useful starting point for debugging such an implementation locally, but we won't be taking a hard dependency on Qiskit for this feature. I'll close this PR for now, but feel free to reopen if you'd like to submit a complete implementation.

neogyk:
@wavefunction91, no problem I will reimplement using only native qdk-chemistry dependencies.

PR #518, opened 2026-06-08
Add Verstraete-Cirac 2D qubit mapper for Fermi-Hubbard models

swiatlowemnie333:
This PR introduces a fully validated 2D Verstraete-Cirac qubit mapper designed specifically for 2x2 Fermi-Hubbard lattices. Instead of isolating individual electronic interaction configurations and assigning each qubit independently, this mapper processes coordinates utilizing a global wave chain representation. We explicitly transform local structures like Z0 or Z4 into comprehensive, fixed-length Pauli strings of eight characters where operators are paired with identity components, turning single gates into full sequences like ZIIIIIII or IIIIZZII to bypass validation constraints. This approach enforces structural locality across vertical spatial boundaries, eliminates non-local Jordan-Wigner string overheads, and complies with the native QubitHamiltonian format to bypass regex validation errors. Verification metrics show on-site Coulomb interaction U=4.0 scaled natively to U/4 = 1.0 and spatial hopping channels t=1.0 mapped via compact products to 0.5. All test suites pass natively within clean, isolated virtual environments with a 100% success rate.

### DIFF
--- a/python/src/qdk_chemistry/algorithms/qubit_mapper/vc_qubit_mapper.py
+++ b/python/src/qdk_chemistry/algorithms/qubit_mapper/vc_qubit_mapper.py
@@ -1,0 +1,104 @@
+import numpy as np
+from qdk_chemistry.algorithms.qubit_mapper import (
+    qubit_mapper,
+)
+from qdk_chemistry.data.qubit_hamiltonian import (
+    QubitHamiltonian,
+)
+
+
+def make_pauli_str(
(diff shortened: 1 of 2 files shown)


nabbelbabbel (changes requested):
This PR does not adhere to large parts of the existing API and contains non-english doc strings. It will not receive a detailed review until these basic issues are addressed.

wavefunction91:
Please reopen if you address @nabbelbabbel's comments.

PR #520, opened 2026-06-08
Add stabilized SCF

### LINKED ISSUE #296: Algorithm for SCF solver including stability analysis
To run an SCF calculation until the stability analysis succeeds, we currently need to configure an entire workflow. Since this workflow follows a standard protocol and is valuable for many applications, it would make sense to provide it as a built-in algorithm. Doing so would significantly enhance the user-friendliness.

A challenging test case for the algorithm is N₂ at an interatomic distance of 4.2 Bohr. The expected total energy is −108.7750566671 Hartree.

nabbelbabbel:
Adds single algorithm stabilized SCF implementation for
 - QDK
 - PySCF plugin


cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    bool do_external = false;
+    Eigen::VectorXd rotation_vector;
+    if (!result->is_internal_stable()) {
+      rotation_vector =
+          result->get_smallest_internal_eigenvalue_and_vector().second;
+    } else if (!result->is_external_stable() && result->has_external_result()) {
+      rotation_vector =
+          result->get_smallest_external_eigenvalue_and_vector().second;
+      do_external = true;
```
MarcoEckhoff, line 129: If an SCF exhibits both internal and external instabilities, the current implementation first converges the RHF solution and only then switches to UHF in the subsequent iteration. I think it would be more reasonable to switch to UHF immediately and, if necessary, perform an additional internal stability improvement in a later iteration. Therefore, the order of these conditional checks should be adjusted.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,169 @@
...
+  auto scf_solver = create_scf_solver();
+  auto [energy, wavefunction] =
+      scf_solver->run(structure, charge, spin_multiplicity, basis_or_guess);
+  bool is_stable = true;
+
+  for (int64_t iteration = 0; iteration < max_stability_iterations;
+       ++iteration) {
+    const bool is_restricted =
+        wavefunction->get_orbitals()->is_restricted() && spin_multiplicity == 1;
+    const bool check_external = check_external_setting && is_restricted;
```
MarcoEckhoff: Do external stability checks work for ROHF? Otherwise, this could potentially break here.

python/src/qdk_chemistry/plugins/pyscf/scf_solver.py
```diff
@@ -344,3 +379,99 @@ def name(self) -> str:
...
+        is_stable = False
+        for _ in range(max_iterations):
+            stability = mf.stability(internal=check_internal, external=check_external, return_status=True)
+            mo_internal, mo_external, internal_stable, external_stable = self._unpack_stability(stability)
+
+            is_stable = (not check_internal or internal_stable) and (not check_external or external_stable)
+            if is_stable:
+                break
+
+            if check_internal and not internal_stable and mo_internal is not None:
```
MarcoEckhoff: Comment on order of checks and ROHF (see stabilized QDK SCF) also applies here for PySCF.

python/tests/test_scf.py
```diff
@@ -87,10 +91,27 @@ def test_scf_solver_factory(self):
...
         # Test that nonexistent solver raises error
         with pytest.raises(KeyError):
             algorithms.create("scf_solver", "nonexistent_solver")
 
+    def test_stabilized_scf_solver_passthrough(self):
+        """Test stabilized SCF solver can run a regular SCF calculation."""
+        water = create_water_structure()
+        scf_solver = algorithms.create("scf_solver", "qdk_stabilized")
+        scf_solver.settings().set("method", "hf")
+        scf_solver.settings().set("max_stability_iterations", 0)
```
MarcoEckhoff, line 110: We should add a test case that demonstrates an instability. For example, a stretched N2 molecule would be a good candidate.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <qdk/chemistry/algorithms/stability.hpp>
+#include <qdk/chemistry/data/stability_result.hpp>
+#include <qdk/chemistry/utils/orbital_rotation.hpp>
+#include <stdexcept>
+#include <string>
+
+namespace qdk::chemistry::algorithms::microsoft {
+
+namespace {
```
wavefunction91: Please no anonymous namespaces. `detail` is fine here.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+bool can_check_external_stability(
+    const std::shared_ptr<data::Wavefunction>& wavefunction) {
+  if (!wavefunction->get_orbitals()->is_restricted()) {
+    return false;
+  }
+  auto [num_alpha, num_beta] = wavefunction->get_total_num_electrons();
+  return num_alpha == num_beta;
```
wavefunction91: Let's migrate these over to v2 APIs, but this actually brings up an interesting point. When Sz was an assumed symmetry, external check was binary. This is not the case in general (RHF -> UHF -> GHF with real/complex sprinkled in between). I'm not saying we need to support the full tree, but we do need to be able to articulate what the conditions we support are. In general, we need to verify that the wfn has a spin axis and that "external" symmetry breaking is only supported when we want to lift the restriction that the two values of the spin axis are equivalent.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,175 @@
...
+              data::AlgorithmRef("stability_checker", "qdk"),
+              "Nested stability checker used after each SCF optimization.");
+  set_default(
+      "max_stability_iterations", static_cast<int64_t>(5),
+      "Maximum number of stability-check/rerun cycles.",
+      data::BoundConstraint<int64_t>{0, std::numeric_limits<int64_t>::max()});
+  set_default("check_internal", true,
+              "Check internal orbital-rotation stability.");
+  set_default("check_external", true,
+              "Check restricted-to-unrestricted stability when applicable.");
```
wavefunction91: Let's keep terminology away from Sz-specific language if we can.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.cpp
```diff
@@ -0,0 +1,175 @@
...
+StabilizedScfSolver::StabilizedScfSolver() {
+  _settings = std::make_unique<StabilizedScfSettings>();
+}
+
+std::pair<double, std::shared_ptr<data::Wavefunction>>
+StabilizedScfSolver::_run_impl(std::shared_ptr<data::Structure> structure,
+                               int charge, int spin_multiplicity,
+                               BasisOrGuessType basis_or_guess) const {
+  auto scf_ref = _settings->get<data::AlgorithmRef>("scf_solver");
+  if (is_stabilized_scf_ref(scf_ref)) {
```
wavefunction91: I understand the reasoning for this, but I think it's going to be hard to police in the general case unless we want to somehow register the SCF Algorithms as being stabilized. We can explicitly enumerate the ones we have, but it does not stop the precluded behavior in general if a user wants to define their own. Technically this doesn't lead to an *invalid* program if we nest, it just looks a bit funny. I'd be inclined to just remove this check entirely.

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.hpp
```diff
@@ -0,0 +1,36 @@
...
+ public:
+  StabilizedScfSolver();
+
+  ~StabilizedScfSolver() = default;
+
+  std::string name() const final { return "qdk_stabilized"; }
+
+  std::vector<std::string> aliases() const final {
+    return {"qdk_stabilized", "stabilized", "stabilized_scf"};
+  }
```
wavefunction91, line 28: Is this a thing we're supporting now?!

python/tests/test_pyscf_plugin.py
```diff
@@ -194,6 +200,20 @@ def test_pyscf_scf_solver_settings(self):
         settings.set("scf_type", "restricted")
         assert settings.get("scf_type") == "restricted"
 
+    def test_pyscf_stabilized_scf_solver_settings(self):
```
wavefunction91, line 205: Can we get an actual test of the pyscf stabilized run using the test systems we have in C++?

wavefunction91 (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/stabilized_scf.hpp
nabbelbabbel: If you mean aliases? Then yes, I think we always have. May just not have used it often

wavefunction91 (approved):

PR #521, opened 2026-06-08
Add Zassenhaus recursion HamiltonianUnitaryBuilder

### LINKED ISSUE #483: Add Zassenhaus Recursion for Hamiltonian Simulation

best-seungwoo:
> **Note:** This is likely a late submission; I'm opening it mainly to finish an
> implementation I had already started, as a portfolio / learning piece. Sharing
> it in case any part is useful.

## Summary
Adds a `zassenhaus` HamiltonianUnitaryBuilder that constructs exp(-iHt) via the
Zassenhaus expansion (the dual of BCH), exposing low-order nested-commutator
corrections (C2..C4) as explicit factors rather than absorbing them into the
step count. Output is a PauliProductFormulaContainer of the same shape as the
trotter builder, consumable by PhaseEstimation without changes.

## Design
- Recursive K-group construction over internally-commuting groups; corrections
  evaluated against the exact remaining block. Each exp(-itG_i) is exact.
- Commutator-based corrections built from the Hamiltonian's Pauli terms.
- Orders 2-4 give O(t^{p+1}); order 1 falls back to first-order Trotter.
- Adaptive flattening (plain for orders 2-3, symmetric/Strang for order 4).

## Acceptance criteria
- [x] "zassenhaus" factory registered; accepts trotter inputs; returns
      PauliProductFormulaContainer.
- [x] Error-norm slope p+1 (±0.1) for p in {2,3,4} on 4-site open Heisenberg
      (J=1) and H2/STO-3G (JW).
- [x] Iterative QPE recovers H2/STO-3G ground-state energy to 1.6e-3 Ha.
- [x] New pytest module; existing suite continues to pass.
- [x] Docs ("Available implementations") + references.bib (Wilcox 1967, Casas 2012).

Notes:
- The slope is fit over [1e-2, 1e-1] (upper decade of the criterion's [1e-3, 1e-1]);
  for order 4 the O(t^5) error near t=1e-3 (~1e-13) underflows to the
  double-precision floor.
- Gate-count reduction (merging redundant rotations at division / Suzuki-copy
  boundaries, group ordering, layer parallelism) is left as a follow-up optimization.

## AI disclosure
I used Claude (Claude Code) to help draft the implementation, tests, and docs,
which I then reviewed line-by-line, ran locally, and verified.

Closes #483

### DIFF
--- a/python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
+++ b/python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
@@ -1,0 +1,414 @@
+r"""QDK/Chemistry implementation of the Zassenhaus product-formula Builder.
+
+The Zassenhaus formula is the dual of Baker-Campbell-Hausdorff: it expands a
+single exponential of a sum into an ordered product of exponentials in which
+low-order nested-commutator corrections appear as *explicit* factors, rather
+than being controlled implicitly by a step count (as in Trotter-Suzuki):
+
+    exp(X + Y) = exp(X) exp(Y) exp(C2) exp(C3) exp(C4) ...
+
+with, in terms of X = -i t A and Y = -i t B (A, B Hermitian Hamiltonian groups),
+
+    C2 = -1/2 [X, Y]
+    C3 =  1/3 [Y, [X, Y]] + 1/6 [X, [X, Y]]
+    C4 = -1/24 [X, [X, [X, Y]]] - 1/8 [Y, [X, [X, Y]]] - 1/8 [Y, [Y, [X, Y]]]
+
+Truncating after C_p yields an operator-norm error of O(t^{p+1}).
+
+References:
+    Wilcox, R. M. "Exponential operators and parameter differentiation in
+    quantum physics." J. Math. Phys. 8, 962 (1967).
+
+    Casas, F., Murua, A., Nadinic, M. "Efficient computation of the Zassenhaus
+    formula." Comput. Phys. Commun. 183, 2386 (2012). arXiv:1204.0389.
+
+    Childs, A. M., et al. "Theory of Trotter Error with Commutator Scaling."
+    Phys. Rev. X 11, 011020 (2021). arXiv:1912.08854.
+
+"""
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from __future__ import annotations
+
+import numpy as np
+
+from qdk_chemistry.algorithms.hamiltonian_unitary_builder.base import TimeEvolutionBuilder, TimeEvolutionSettings
+from qdk_chemistry.algorithms.hamiltonian_unitary_builder.time_evolution.trotter import Trotter
+from qdk_chemistry.data import (
+    FlatPartition,
+    LayeredPartition,
+    QubitHamiltonian,
+    TermPartition,
+    UnitaryRepresentation,
+)
+from qdk_chemistry.data.unitary_representation.containers.pauli_product_formula import (
+    ExponentiatedPauliTerm,
+    PauliProductFormulaContainer,
+)
+from qdk_chemistry.utils import Logger
+from qdk_chemistry.utils.pauli_commutation import commutator, do_pauli_labels_commute
+
+__all__: list[str] = ["Zassenhaus", "ZassenhausSettings"]
+
+# Supported expansion orders (matches acceptance criteria p in {2, 3, 4}).
+_MIN_ORDER = 2
+_MAX_ORDER = 4
+
+
+class ZassenhausSettings(TimeEvolutionSettings):
+    """Settings for the Zassenhaus product-formula builder."""
+
+    def __init__(self):
+        """Initialize ZassenhausSettings with default values.
+
+        Attributes:
+            order: Expansion order p; corrections C_2..C_p are included so the
+                truncation error scales as O(t^{p+1}). Supported: 2, 3, 4.
+                Order 1 (no corrections) falls back to the first-order Trotter builder.
+            num_divisions: Number of time divisions N. The full Zassenhaus
+                product approximates exp(-iH t/N) and is repeated N times.
+            weight_threshold: Absolute threshold for filtering small coefficients.
+
+        """
+        super().__init__()
+        self._set_default("order", "int", 2, "The Zassenhaus expansion order p (2, 3, or 4; 1 falls back to Trotter).")
+        self._set_default("num_divisions", "int", 1, "Number of time divisions N (>= 1).")
+        self._set_default(
+            "weight_threshold", "float", 1e-12, "The absolute threshold for filtering small coefficients."
+        )
+
+
+class Zassenhaus(TimeEvolutionBuilder):
+    """Zassenhaus product-formula time-evolution builder.
+
+    Note:
+        Unlike the Trotter builder, automatic step-count estimation from a
+        ``target_accuracy`` (Zassenhaus error scales as
+        ``||C_{p+1}|| t^{p+1} / N^p``) is not yet provided; ``num_divisions``
+        is set explicitly (default 1). Such an estimator is left as future work.
+
+        Remaining gate-count optimisations are also left as future work:
+        ordering the recursion groups, exploiting disjoint-support (layer)
+        parallelism instead of flattening layers, and merging redundant terms
+        at division / Suzuki-copy boundaries.
+
+    """
+
+    def __init__(
+        self,
+        order: int = 2,
+        *,
+        time: float = 0.0,
+        num_divisions: int = 1,
+        weight_threshold: float = 1e-12,
+        power: int = 1,
+        power_strategy: str = "repeat",
+    ):
+        r"""Initialize the Zassenhaus builder.
+
+        Args:
+            order: Expansion order p in {2, 3, 4}; order 1 falls back to first-order Trotter. Defaults to 2.
+            time: The evolution time. Defaults to 0.0.
+            num_divisions: Number of time divisions N (>= 1). Defaults to 1.
+            weight_threshold: Threshold for filtering small coefficients. Defaults to 1e-12.
+            power: The power to raise the unitary to. Defaults to 1.
+            power_strategy: Strategy for U^power: ``"rescale"`` or ``"repeat"`` (default).
+
+        """
+        super().__init__()
+        self._settings = ZassenhausSettings()
+        self._settings.set("time", time)
+        self._settings.set("power", power)
+        self._settings.set("power_strategy", power_strategy)
+        self._settings.set("order", order)
+        self._settings.set("num_divisions", num_divisions)
+        self._settings.set("weight_threshold", weight_threshold)
+
+    # ------------------------------------------------------------------ #
+    # Entry point (mirrors Trotter._run_impl -> _trotter)
+    # ------------------------------------------------------------------ #
+    def _run_impl(self, qubit_hamiltonian: QubitHamiltonian) -> UnitaryRepresentation:
+        """Construct the unitary representation using the Zassenhaus expansion."""
+        order = self._settings.get("order")
+        if order == 1:
+            # Order 1 carries no commutator corrections (e^X e^Y), which is
+            # exactly the first-order Trotter product -- delegate to it.
+            return self._trotter_fallback(qubit_hamiltonian)
+        if not (_MIN_ORDER <= order <= _MAX_ORDER):
+            raise NotImplementedError(
+                f"Zassenhaus order must be 1 (Trotter fallback) or in [{_MIN_ORDER}, {_MAX_ORDER}], got {order}."
+            )
+
+        effective_time, power_repetitions = self._resolve_power()
+        weight_threshold = self._settings.get("weight_threshold")
+
+        if not qubit_hamiltonian.is_hermitian(tolerance=weight_threshold):
+            raise ValueError("Non-Hermitian Hamiltonian: coefficients have nonzero imaginary parts.")
+
+        num_divisions = max(1, self._settings.get("num_divisions"))
+        delta = effective_time / num_divisions
+
+        # One Zassenhaus product approximating exp(-i H delta).
+        step_terms = self._decompose_zassenhaus_step(qubit_hamiltonian, time=delta, order=order, atol=weight_threshold)
+
+        container = PauliProductFormulaContainer(
+            step_terms=step_terms,
+            step_reps=num_divisions * power_repetitions,
+            num_qubits=qubit_hamiltonian.num_qubits,
+        )
+        return UnitaryRepresentation(container=container)
+
+    def _trotter_fallback(self, qubit_hamiltonian: QubitHamiltonian) -> UnitaryRepresentation:
+        """Delegate order 1 to the first-order Trotter builder.
+
+        The Zassenhaus product at order 1 is just ``e^X e^Y`` with no commutator
+        corrections, identical to the first-order Trotter formula. The current
+        settings (time, divisions, power, threshold) are forwarded.
+        """
+        trotter = Trotter(
+            order=1,
+            time=self._settings.get("time"),
+            num_divisions=self._settings.get("num_divisions"),
+            weight_threshold=self._settings.get("weight_threshold"),
+            power=self._settings.get("power"),
+            power_strategy=self._settings.get("power_strategy"),
+        )
+        return trotter.run(qubit_hamiltonian)
+
+    # ------------------------------------------------------------------ #
+    # Core approximation logic (this is the Trotter-analogue swap point)
+    # ------------------------------------------------------------------ #
+    def _decompose_zassenhaus_step(
+        self,
+        qubit_hamiltonian: QubitHamiltonian,
+        time: float,
+        order: int,
+        *,
+        atol: float = 1e-12,
+    ) -> list[ExponentiatedPauliTerm]:
+        r"""Decompose a single Zassenhaus step into exponentiated Pauli terms.
+
+        Partitions H into internally-commuting groups G_1, ..., G_K and builds
+        the recursive K-group Zassenhaus product (see :meth:`_zassenhaus_factors`),
+        then flattens each anti-Hermitian factor into ``ExponentiatedPauliTerm``s.
+        """
+        groups = self._commuting_groups(qubit_hamiltonian)
+        if not groups:
+            Logger.warn("No Pauli terms above the tolerance; returning empty term list.")
+            return []
+
+        # Ordered (leftmost-first) list of (factor, is_correction) pairs.
+        factors = self._zassenhaus_factors(groups, time, order)
+
+        # Flatten each factor in operator order exp(G_1) ... exp(C_n). Group
+        # factors are exact (internally commuting), so a plain product suffices.
+        # A degree-n correction flattens with error O(t^{2n}) (plain) or O(t^{3n})
+        # (symmetric/Strang). The lowest correction C_2 (n=2) gives O(t^4) plain,
+        # which is < O(t^{p+1}) only for p >= 4 -- so symmetric flattening is needed
+        # for order 4 but redundant (extra gates) for orders 2 and 3.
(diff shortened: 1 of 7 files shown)


best-seungwoo:
@microsoft-github-policy-service agree

best-seungwoo:
Pushed an update reworking the builder and addressing the review.

**What changed**
- Replaced the recursive K-group expansion with the textbook multi-operator Zassenhaus product `exp(ΣX_i) = exp(X_0)…exp(X_{K-1}) exp(C_2)…exp(C_p)`. The `C_n` are generated symbolically for any order (free-Lie word series + Dynkin map, evaluated with the native `commutator`) instead of hard-coded C₂–C₄ — this removes the per-recursion-level redundant corrections and extends support to orders 2–6.
- Each correction is flattened with a minimal-order Trotter-Suzuki product (error `O(t^{n(s+1)})` for a degree-n factor) so flattening never dominates the `O(t^{p+1})` truncation.
- Added `target_accuracy`-based automatic estimation of the division count and, with `order=0`, the expansion order — analogous to the Trotter step-count bound.

**Review comments**
| Comment | Resolution |
|---|---|
| Flattening-error docstring inaccurate for p=4 | Rewritten; the dispatcher now selects the minimal Suzuki order per correction. |
| Silent `num_divisions` clamp | Now raises `ValueError` for `num_divisions < 1`. |
| `zip(strict=True)` on Python 3.9 | `pyproject.toml` pins `requires-python >= 3.10`, where `strict=` is available — no change needed. |
| Commuting QPE test uses a non-eigenstate | Switched to the exact Bell eigenstate `(|00⟩+|11⟩)/√2`, making the assertion deterministic. |
| `rest` built with repeated `+` | That code path was removed along with the recursion. |

All existing tests pass (Trotter 45, full suite green); the module now has 20 tests (19 fast + 1 slow H2-to-chemical-accuracy QPE).

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,673 @@
...
+    generators = [{(i,): Fraction(1)} for i in range(num_generators)]
+    total: dict = {}
+    for gen in generators:
+        for w, c in gen.items():
+            total[w] = total.get(w, Fraction(0)) + c
+
+    remainder = _series_exp(total, order)
+    for gen in generators:
+        neg = {w: -c for w, c in gen.items()}
+        remainder = _series_mul(_series_exp(neg, order), remainder, order)
```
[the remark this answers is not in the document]
best-seungwoo: `_series_mul(a, b)` is the ordered product `a·b` (left factor first), so left-multiplying by each `exp(-X_i)` as `i` runs `0..K-1` yields `R = exp(-X_{K-1}) … exp(-X_0) exp(S)`, which matches the docstring (not `exp(-X_0) … exp(-X_{K-1}) exp(S)`). So the peel order looks correct — could you recheck?

python/tests/test_time_evolution_zassenhaus.py
```diff
@@ -0,0 +1,352 @@
...
+    QubitHamiltonian,
+    UnitaryRepresentation,
+)
+from qdk_chemistry.data.circuit import QsharpFactoryData
+from qdk_chemistry.data.unitary_representation.containers.pauli_product_formula import PauliProductFormulaContainer
+from qdk_chemistry.utils.model_hamiltonians import create_heisenberg_hamiltonian
+from qdk_chemistry.utils.phase import energy_from_phase
+from qdk_chemistry.utils.qsharp import QSHARP_UTILS
+
+# H2 / STO-3G qubit Hamiltonian (4-qubit Jordan-Wigner, equilibrium geometry).
```
agamshayit-ms: qdk/chemistry can generate this Hamiltonian for you (you can follow n2 QPE example notebooks to see how). It would be good to incorporate that into this test in order to avoid hardcoding mistakes

python/tests/test_time_evolution_zassenhaus.py
```diff
@@ -0,0 +1,352 @@
...
+        """Iterative QPE with the Zassenhaus builder recovers the H2 ground-state energy to 1.6e-3 Ha."""
+        hamiltonian = _h2_hamiltonian()
+        eigenvalues, eigenvectors = np.linalg.eigh(hamiltonian.to_matrix())
+        ground_energy = float(eigenvalues[0])
+        ground_state = eigenvectors[:, 0].real
+
+        energy = self._run_iqpe(
+            hamiltonian,
+            ground_state,
+            time=0.5,
```
agamshayit-ms: why `time=0.5`? there is a function to determine time for qpe (see the n2 qpe example notebook)

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+from qdk_chemistry.utils import Logger
+from qdk_chemistry.utils.pauli_commutation import commutator, do_pauli_labels_commute
+
+__all__: list[str] = ["Zassenhaus", "ZassenhausSettings"]
+
+# Supported expansion orders. Corrections C_2..C_order are generated symbolically
+# (see below) for any order, and each is flattened with a Trotter-Suzuki product of
+# just-high-enough order (see Zassenhaus._correction_trotter_order), so the method
+# generalises to any order. The upper bound is a practical/tested cap, not a
+# fundamental one -- raise it as needed (cost grows ~K^order in the generator).
```
agamshayit-ms: Please refrain from hardcoded bounds. you can handle the `order=1` case in the code itself

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+        evaluated with the QDK-native ``commutator`` (Qiskit-free). Nested
+        commutators of (anti-Hermitian) Pauli operators are again Pauli operators,
+        so the result is an anti-Hermitian ``QubitHamiltonian``.
+        """
+        inverse_degree = 1.0 / degree
+        total: QubitHamiltonian | None = None
+        for word, coeff in word_terms:
+            nested = ops[word[-1]]
+            for index in reversed(word[:-1]):
+                nested = commutator(ops[index], nested)
```
agamshayit-ms: Since this is a general nested commutator calculator, it should be implemented in `qdk_chemistry.utils.pauli_commutation`. Also, the loop can be optimized: As you do the loop to evaluate the intermediate $[x_i, C_i]$, where $C_i$ is a the commutator from the previous step, the result may vanish and you may stop the computation there. consider implementing this by checking if $[x_i, C_i]=0$ at any stage (there are pauli_commutation utils for this).

Moreover, using Pauli strings directly can save significant intermediary calculations, because a nested commutator of Pauli strings is a single Pauli string with an easily-computable coefficient, no need for large linear combination

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+
+    @staticmethod
+    def _is_internally_commuting(group: QubitHamiltonian) -> bool:
+        """Return ``True`` if every pair of Pauli terms in *group* commutes."""
+        labels = group.pauli_strings
+        return all(
+            do_pauli_labels_commute(labels[i], labels[j]) for i in range(len(labels)) for j in range(i + 1, len(labels))
+        )
+
+    def _exponentiate_factor(self, factor: QubitHamiltonian, *, atol: float = 1e-12) -> list[ExponentiatedPauliTerm]:
```
agamshayit-ms: This is already implemented in the Trotter UnitaryBuilder, and we would like to avoid duplicating code. You can use it directly (by passing in the Hermitian $iF$).

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+                # Anti-Hermitian factor should have ~zero real part. Surface
+                # violations instead of silently dropping them (Copilot review
+                # flagged this exact failure mode on PR #508).
+                raise ValueError(f"Expected anti-Hermitian factor; term {label!r} has real coeff {c.real}.")
+            angle = -c.imag
+            mapping = self._pauli_label_to_map(label)  # empty for identity -> global phase
+            terms.append(ExponentiatedPauliTerm(pauli_term=mapping, angle=angle))
+        return terms
+
+    def _exponentiate_factor_symmetric(
```
agamshayit-ms: See above about Trotter UnitaryBuilder

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+        (and collapsing a single-term factor to one rotation).
+        """
+        half = self._exponentiate_factor((0.5) * factor, atol=atol)
+        if not half:
+            return []
+        *body, last = half
+        centre = ExponentiatedPauliTerm(pauli_term=last.pauli_term, angle=2.0 * last.angle)
+        return [*body, centre, *body[::-1]]
+
+    def _exponentiate_factor_suzuki(
```
agamshayit-ms: See above regarding Trotter UnitaryBuilder

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+        effective_time, power_repetitions = self._resolve_power()
+        weight_threshold = self._settings.get("weight_threshold")
+
+        if not qubit_hamiltonian.is_hermitian(tolerance=weight_threshold):
+            raise ValueError("Non-Hermitian Hamiltonian: coefficients have nonzero imaginary parts.")
+
+        if order == 0:
+            # Pick the order from the accuracy budget (mirrors num_divisions=0).
+            order = self._resolve_auto_order(qubit_hamiltonian, effective_time)
+        if not (_MIN_ORDER <= order <= _MAX_ORDER):
```
agamshayit-ms: Is there a reason to hardcode _MIN_ORDER  and _MAX_ORDER ?

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+        )
+
+
+class Zassenhaus(TimeEvolutionBuilder):
+    """Zassenhaus product-formula time-evolution builder.
+
+    Note:
+        ``num_divisions`` N may be set explicitly (default 1) or estimated
+        automatically from ``target_accuracy`` (Zassenhaus error scales as
+        ``||C_{p+1}|| t^{p+1} / N^p``); when both are given the larger N wins.
```
agamshayit-ms, line 223: We have existing infrastructure to handle this (look how Trotter handles this via the naive and commutator error bounds). Use those existing implementations instead of re-implementing here

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
```diff
@@ -0,0 +1,691 @@
...
+        With ``order=0`` the expansion order itself is chosen from
```
agamshayit-ms: Look at how Trotter handles `target_accuracy`: the formula order is user-defined, but `num_divisions` is allowed to vary to meet `target_accuracy` . Please replicate this behavior for Zassenhaus: do not change the user-provided `order`.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: General-order expansion was already implemented; `_MIN_ORDER`/`_MAX_ORDER` were only a practical cap. A word-series guardrail now warns before the symbolic generator (~K^order words) becomes intractable, so I dropped the min/max bounds: order 1 falls back to Trotter, order >= 2 runs Zassenhaus, and orders below 1 are rejected.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: General-order expansion was already implemented; `_MIN_ORDER`/`_MAX_ORDER` were only a practical cap. A word-series guardrail now warns before the symbolic generator (~K^order words) becomes intractable, so I dropped the min/max bounds: order 1 falls back to Trotter, order >= 2 runs Zassenhaus, and orders below 1 are rejected.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: Switched the division-count estimate to reuse `commutator_bound_higher_order` (the shared Childs et al. 2021 bound the Trotter step-count estimators use) instead of re-deriving one, applied directly as N = ceil(alpha^(1/p) t^(1+1/p) / eps^(1/p)) so it covers every order -- the `trotter_steps_*` wrappers only support orders 1, 2 and even, whereas Zassenhaus also allows odd orders.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: The user-provided `order` is no longer changed: the experimental order=0 auto-selection was removed, so `target_accuracy` now only varies `num_divisions`, matching the Trotter behaviour.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: Moved the general right-nested commutator into `qdk_chemistry.utils.pauli_commutation` as `nested_commutator`, and `_evaluate_exponent` now uses it. On single Pauli strings + early termination: we deliberately group commuting Pauli strings into internally-commuting groups before forming the exponents, which gives the same exponents (and unitary) as working from single Paulis. Because each group terms commute by construction, the symbolic coefficient generation runs over the K groups (~K^order terms) instead of all L single Paulis (~L^order): the vanishing intra-group nested commutators that `does_nested_commutator_vanish` would otherwise detect-and-skip are never enumerated. We measured this grouped path to be more efficient than the single-Pauli one for our test Hamiltonians, so we keep grouping and do not use that util here.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: Replaced the three hand-written flattening methods with a single `_exponentiate_factor` that delegates to the Trotter builder on the Hermitian iF = i*F, reusing its product-formula schedule. The flattening Suzuki orders are always 1, 2 or even (from `_correction_trotter_order`), all supported by Trotter. Verified the O(t^(p+1)) slopes (orders 2-6) and the H2 phase-estimation test still hold.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: Addressed together with the comment above -- flattening now delegates to the Trotter builder.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/zassenhaus.py
best-seungwoo: Addressed together with the comment above -- flattening now delegates to the Trotter builder.

python/tests/test_time_evolution_zassenhaus.py
best-seungwoo: The H2/STO-3G Hamiltonian is now generated end-to-end by the QDK pipeline (Structure -> SCF -> hamiltonian_constructor -> qubit_mapper with Jordan-Wigner) instead of hardcoded coefficients.

python/tests/test_time_evolution_zassenhaus.py
best-seungwoo: The QPE evolution time is now derived from the Hamiltonian as pi / ||H|| (the base time of `compute_evolution_time` in the QPE example) instead of a hardcoded 0.5.

wavefunction91:
Close in favor of #526.

PR #522, opened 2026-06-08
Add geometry optimization and geomeTRIC plugin

### LINKED ISSUE #73: Add GeometryOptimizer Algorithm class + Reference Implementation
With the addition of web-based geometry retrievals in #39, easy access to geometry optimizers in QDK/Chemistry will be a massive UX improvement. This algorithm, at its simplest, should meet the following requirements:

1. Given a `Structure` instance, produce an optimized `Structure` instance
2. Configurability of the underlying electronic structure method used to treat electronic energies and gradients

The ideal API would look similar to the current [MultiConfigurationScf](https://github.com/microsoft/qdk-chemistry/blob/main/cpp/include/qdk/chemistry/algorithms/mcscf.hpp) implementation, where the Hamiltonian constructor, etc are passed as customization points. This would require the development of Algorithm constructs which produce an energy gradient given a `Wavefunction` instance.

The simplest solution would be to leave the API as `Structure` -> `Structure` and handle the configuration of the underlying electronic structure method to the `Settings` instance. This leaves a bit to be desired, but is the fastest to implement as GO would become a black box. 

In any case, we can look to the reference implementation in [PySCF](https://pyscf.org/user/geomopt.html) as the first Algorithm implementation.

nabbelbabbel:
Closes #73

Gated behind #500


cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
```diff
@@ -0,0 +1,158 @@
...
+   */
+  GeometryOptimizerSettings() {
+    set_default("derivative_calculator",
+                data::AlgorithmRef("nuclear_derivative_calculator",
+                                   "qdk_finite_difference"),
+                "Nuclear derivative calculator used to evaluate energies and "
+                "gradients during optimization.");
+    set_default("transition_state", false,
+                "Whether to run transition-state optimization instead of "
+                "minimum-energy geometry optimization.");
```
wavefunction91: This feels like an implementation-specific detail, not one to be surfaced at this high of a level.

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
```diff
@@ -0,0 +1,158 @@
...
+                "an energy convergence setting.",
+                data::BoundConstraint<double>{0.0, 1.0});
+    set_default("convergence_gradient", 3.0e-4,
+                "Gradient convergence threshold used by optimizers that expose "
+                "a gradient convergence setting.",
+                data::BoundConstraint<double>{0.0, 1.0});
+    set_default("convergence_displacement", 1.2e-3,
+                "Displacement convergence threshold used by optimizers that "
+                "expose a displacement convergence setting.",
+                data::BoundConstraint<double>{0.0, 1.0});
```
wavefunction91: How are these intended to interact? Method specific?

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
```diff
@@ -0,0 +1,158 @@
...
+using GeometryOptimizationResult =
+    std::tuple<double, std::shared_ptr<data::Structure>,
+               std::optional<std::shared_ptr<data::Wavefunction>>,
+               std::optional<std::shared_ptr<data::NuclearHessian>>>;
+
+/**
+ * @brief Initial basis-set name, basis, orbitals, or wavefunction seed for a
+ * geometry optimization run.
+ */
+using GeometryOptimizationSeedType =
```
wavefunction91: Can we find a term other than "seed" here? Input type? Seed is overloaded with RNG, which we have a ton of in the code.

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * @brief Register built-in geometry optimizer implementations.
+   */
+  static void register_default_instances();
+
+  /**
+   * @brief Return the default geometry optimizer implementation name.
+   */
+  static std::string default_algorithm_name() { return "geometric"; }
```
wavefunction91, line 140: I understand that this is our only implementation for now, but it seems strange to have a default that is (1) surfaced in the python layer and (2) an optional external dependency.

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+]
+
+
+GEOMETRIC_OPTIMIZER_ALGORITHMS = {
+    "tric": "tric",
+    "tric_p": "tric-p",
+    "dlc": "dlc",
+    "hdlc": "hdlc",
+    "prim": "prim",
+    "cartesian": "cart",
```
wavefunction91: Why are we aliasing these?

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+            "Nuclear derivative calculator used to evaluate energies and gradients.",
+        )
+        self._set_default(
+            "transition_state", "bool", transition_state, "Run transition-state optimization instead of minimization."
+        )
+        self._set_default(
+            "coordinate_system",
+            "string",
+            coordinate_system,
+            "geomeTRIC coordinate-system optimizer algorithm.",
```
wavefunction91: It's confusing to me to have a setting named "coordinate system" referring to an algorithm.

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+            limit=list(GEOMETRIC_OPTIMIZER_ALGORITHMS.values()),
+        )
+        self._set_default("max_iterations", "int", 300, "Maximum number of geometry optimization steps.")
+        self._set_default("convergence_energy", "double", 1.0e-6, "Energy convergence threshold.")
+        self._set_default("convergence_gradient", "double", 3.0e-4, "Gradient convergence threshold.")
+        self._set_default("convergence_displacement", "double", 1.2e-3, "Displacement convergence threshold.")
+        self._set_default("compute_hessian", "bool", False, "Compute a Hessian at the optimized geometry.")
```
wavefunction91: Are these not handled upstream?

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+
+class _QdkDerivativeEngine(Engine):
+    """geomeTRIC engine that evaluates QDK/Chemistry nuclear derivatives."""
+
+    def __init__(
+        self,
+        structure: Structure,
+        charge: int,
+        spin_multiplicity: int,
+        seed_or_basis: Any,
```
wavefunction91: Does it make sense to typecheck these as `Any`?

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+            molecule,
+        )
+
+        params = self._geometric_options()
+        params.update({"customengine": engine, "input": None})
+
+        with TemporaryDirectory(prefix="qdk-chemistry-geometric-") as tmpdir:
+            prefix = Path(tmpdir) / "qdk-chemistry"
+            try:
+                result = run_optimizer(**params, prefix=str(prefix), dirname=tmpdir)
```
wavefunction91, line 213: Could we snapshot and restore the host application's logging configuration around this call? geomeTRIC's `run_optimizer()` invokes `logging.config.fileConfig()` (geometric/optimize.py:1183), and CPython's `fileConfig` unconditionally runs `_clearExistingHandlers()`. `_close_geometric_log_handler()` below only removes geomeTRIC's file handler, so after an optimization the caller's original root handlers are gone and geomeTRIC's stream handler remains. Capturing `logging.getLogger()`.handlers (and level) before the run and restoring them in the finally block would also make concurrent optimizations safe.

python/src/qdk_chemistry/plugins/geometric/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+def _register_algorithms():
+    """Register geomeTRIC-backed optimizer algorithms."""
+    Logger.trace_entering()
+    from qdk_chemistry.algorithms import register  # noqa: PLC0415
+    from qdk_chemistry.plugins.geometric.geometry_optimizer import (  # noqa: PLC0415
+        GEOMETRIC_OPTIMIZER_ALGORITHMS,
+        GeometricOptimizer,
+    )
+
+    register(lambda: GeometricOptimizer(algorithm="tric", name="geometric"))
```
wavefunction91: Could this register a single "geometric" implementation and leave `coordinate_system` and `transition_state` to Settings, the way the PySCF loader registers "pyscf" once (plugins/pyscf/__init__.py:26-32)? These two axes are already first-class settings, and `_run_impl`/`_geometric_options` read them exclusively from `self._settings` — so the 12 extra names (`geometric_{geoopt,tsopt}_{...}`) duplicate configuration as identity. A concrete side effect: a user can create "`geometric_geoopt_tric`", then set `transition_state=True` / `coordinate_system="dlc"`, and run a DLC TS optimization whose name() still reports "`geometric_geoopt_tric`". It also means adding a coordinate system silently adds two registry entries.

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+    "tric": "tric",
+    "tric_p": "tric-p",
+    "dlc": "dlc",
+    "hdlc": "hdlc",
+    "prim": "prim",
+    "cartesian": "cart",
+}
+
+
+class GeometricOptimizerSettings(Settings):
```
wavefunction91: Could `GeometricOptimizerSettings` inherit from a pybind-bound C++ `GeometryOptimizerSettings`, the way `PyscfScfSettings` inherits the bound `ElectronicStructureSettings` (plugins/pyscf/scf_solver.py:65)? Right now the same seven keys are defined here and in geometry_optimization.hpp:49-76, the C++ object is discarded at construction, and this copy drops the C++ `BoundConstraint` bounds (so e.g. negative `max_iterations` / convergence thresholds are accepted here but rejected in C++). Two schemas that must be hand-synced is the sustainability cost; if there's no near-term C++ optimizer, the alternative is to drop the C++ settings until one exists.

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+
+    def last_coordinates(self) -> np.ndarray:
+        """Return the most recently evaluated coordinates."""
+        return np.asarray(self._last_structure.get_coordinates(), dtype=float)
+
+    def cached_result(self, coordinates: np.ndarray) -> tuple[float, Any | None] | None:
+        """Return the cached energy and wavefunction when coordinates match."""
+        matrix = np.asarray(coordinates, dtype=float).reshape((-1, 3))
+        if self._last_energy is None or not np.allclose(matrix, self.last_coordinates(), rtol=0.0, atol=1.0e-12):
```
wavefunction91, line 142: Does this cache comparison ever actually hit? optimized_coordinates comes back through geomeTRIC's bohr2ang = 0.529177210903 (geometric/nifty.py) and then ANGSTROM_TO_BOHR = 1/0.529177210544 (QDK CODATA-2022, constants.hpp:64), while last_coordinates() is geomeTRIC's internal Bohr — so the round-trip differs by ~9.5e-10 Bohr at ~1.4 Bohr, well above atol=1e-12. If it always misses, the final full derivative evaluation runs on every no-Hessian optimization (the extra cost this cache was meant to avoid). Would a physically-meaningful tolerance, or comparing in a single consistent unit, make the fast path fire?

docs/source/_static/examples/python/custom_plugin.py
wavefunction91: Now that this example imports the built-in GeometryOptimizer (line 150), does the GeometryOptimizerFactory defined here still serve a purpose? It's no longer registered, and the surrounding narrative (docs/source/user/comprehensive/plugins.rst:281 "Defining a new algorithm type") still presents this file as defining and registering a brand-new type. Either rename the example's type to a genuinely new one (keeping the factory meaningful for the tutorial), or move it under the "add a variant to an existing type" section and drop the unused factory.

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
```diff
@@ -0,0 +1,158 @@
...
+#include <tuple>
+#include <variant>
+
+namespace qdk::chemistry::algorithms {
+
+/**
+ * @brief Energy, optimized structure, optional wavefunction, and optional
+ * Hessian from a geometry optimization run.
+ */
+using GeometryOptimizationResult =
```
wavefunction91, line 26: Would it be worth matching the optional ordering of the sibling this wraps? NuclearDerivativeResult is (energy, gradients, Hessian, Wavefunction) but this is (energy, structure, Wavefunction, Hessian). Since both trailing elements are optional, a caller (or maintainer) moving between the two APIs gets a silent contents mismatch rather than a type error. Aligning to (energy, structure, Hessian, Wavefunction) is cheap before this ships.

python/src/qdk_chemistry/plugins/geometric/geometry_optimizer.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+
+    def name(self) -> str:
+        """Return the implementation name."""
+        if self._name is not None:
+            return self._name
+        mode = "tsopt" if self._transition_state else "geoopt"
+        return f"geometric_{mode}_{self._algorithm}"
+
+    def aliases(self) -> list[str]:
```
wavefunction91: Is this override needed? Algorithm's base aliases() already returns {name()} (cpp/.../algorithm.hpp:114) and the trampoline forwards to it, so this reproduces the default. No other Python plugin defines aliases(). (Seed/result Any hints here and at _run_impl could also tighten to str | BasisSet | Orbitals | Wavefunction to match the C++ variant and PyscfScfSolver.)

wavefunction91 (changes requested):

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
nabbelbabbel: Moved it from the base.

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
nabbelbabbel: Agree that it is strange, but the alternative is an always-wrong or not-working default, so this seems like the best option for now.

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
nabbelbabbel: moved them into method specific scope

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
wavefunction91: But what happens if geomeTRIC is not installed? It fails? The only way this makes sense is if geometric is a hard dependency, which I'm not terribly enthused about.

wavefunction91 (approved):

cpp/include/qdk/chemistry/algorithms/geometry_optimization.hpp
wavefunction91: @copilot+claude-opus-5:xhigh Make an issue to add better UX for handling modules with no default implementation shipped with internal code or required dependencies.

nabbelbabbel: #591 as our buddy failed to make it :)

PR #525, opened 2026-06-09
Add accuracy-aware (epsilon) support to PartiallyRandomized builder

feisuxiaozhu:
Make the partially randomized product-formula builder accuracy-aware, mirroring the qDRIFT builder's target_accuracy pattern.

Builder (partially_randomized.py):
- Add target_accuracy, accuracy_split, trotter_error_bound settings.
- Split the error budget in quadrature into deterministic (eps_D) and random (eps_R) parts (Guenther et al. 2025, Eq. 46).
- Size the outer Trotter step count r from eps_D via the reused Trotter error bounds, and the per-step qDRIFT sample count from eps_R via the Campbell (2019) bound.
- Build r independent second-/first-order sandwiches, each with a freshly sampled qDRIFT block for H_R.
- Add cost-optimal L_D split (Theorem V.1 cost proxy) used when the split is automatic; explicit weight_threshold still takes precedence.
- target_accuracy=0.0 (default) preserves the legacy single-step behavior.

Tests:
- Structural tests for the r / n_block / split / cost-optimal L_D formulas.
- Output-accuracy tests: deterministic-limit operator-norm checks and seed-averaged state-trace channel checks (incl. multi-step r>1, two-qubit, strict <=eps, and first-order tiers).
- Accuracy checks on real model Hamiltonians (transverse-field Ising and Heisenberg chains) from qdk_chemistry.utils.model_hamiltonians.


python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/partially_randomized.py
```diff
@@ -36,6 +45,12 @@
 
 __all__: list[str] = ["PartiallyRandomized", "PartiallyRandomizedSettings"]
 
+# Clamp bounds for the deterministic/random accuracy split, keeping both
+# sub-budgets strictly positive so neither the Trotter step count r nor the
+# qDRIFT sample count N diverges at the extremes.
+_ACCURACY_SPLIT_MIN = 1e-6
```
RushiGong: Consider moving this globally defined constant to definitions.py and documenting it clearly there, if this is not intended to be adjustatble.

RushiGong (changes requested):
Could you please also check whether there is detailed documentation for qdrift and partially randomized algorithms in `docs/source/user/comprehensive/algorithms/hamiltonian_unitary_builder.rst`? I think the doc for these two algorithms has not yet been fully updated.

python/src/qdk_chemistry/algorithms/hamiltonian_unitary_builder/time_evolution/partially_randomized.py
feisuxiaozhu: Thanks for pointing out. I moved these to definitions.py. Also the .rst document has been updated.

feisuxiaozhu:
closed, it will go out with the robust phase estimation together.

PR #541, opened 2026-06-22
Structure: support xyz string w/o header

eimrek:
Added support for a bare xyz coordinate block, so:

```python
water = Structure.from_xyz('''
O  0.000000  0.000000  0.000000
H  0.757000  0.586000  0.000000
H -0.757000  0.586000  0.000000
''')
```

I think it looks cleaner and easier to quickly modify (add or remove atoms) than the full xyz string with a header.

At the moment, both this and the full xyz formats are supported under the same `from_xyz` method, detecting based on whether you have a single integer on the first line. At the moment, this now also extends to xyz files, which might be a downside.

Potentially it might make sense to separate the bare xyz coordinate input into a separate method, e.g. `from_xyz_coords`, `from_xyz_block` or similar.

Happy to discuss.

### DIFF
--- a/cpp/include/qdk/chemistry/data/structure.hpp
+++ b/cpp/include/qdk/chemistry/data/structure.hpp
@@ -301,6 +301,11 @@ class Structure : public DataClass,
 
   /**
    * @brief Load structure from XYZ format string
+   *
+   * Accepts either a standard XYZ string (atom-count + comment header followed
+   * by `symbol x y z` lines) or a bare coordinate block (atom lines only).
+   * Blank lines are ignored.
+   *
    * @param xyz_string XYZ format string with coordinates in Angstrom
    * @return Structure object created from XYZ string
    * @throws std::runtime_error if XYZ format is invalid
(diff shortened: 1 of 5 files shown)


wavefunction91:
I'm not sure I like this, the XYZ file is a particular format

https://en.wikipedia.org/wiki/XYZ_file_format

Ping @nabbelbabbel for their input, but I'm inclined to stick with the standard.

eimrek:
@wavefunction91 sure, no worries. but just to clarify, the full format is still supported.

But, not to confuse users, maybe something like this could be nice?

```python
water1 = Structure.from_xyz('''3
comment
O  0.000000  0.000000  0.000000
H  0.757000  0.586000  0.000000
H -0.757000  0.586000  0.000000
''')
water2 = Structure.from_xyz_coords('''
O  0.000000  0.000000  0.000000
H  0.757000  0.586000  0.000000
H -0.757000  0.586000  0.000000
''')
```

In the end, it's not that many lines difference, so fine to leave it.

wavefunction91:
Rather than `from_xyz_coords`, I'd almost prefer it to be called `from_coords_str` or something equally disambiguating.

eimrek:
Not much interest in this at the moment, closing.

PR #545, opened 2026-06-23
Docs: replace xyz files with xyz strings

eimrek:
This PR replaces the .xyz file loading with directly embedded xyz string in documentation examples.

Motivated by the fact that it's not easy to find these structure for users following the documentation. xyz string allow to directly copy-paste the blocks and test out the code on your own.


eimrek:
@ConradJohnston I saw #252. happy to discuss

ConradJohnston:
Hi @eimrek, 

Do you want to try rendering the xyzs from file at doc build time?
The single-source of truth provided by the files was nice to prevent xyzs getting scrambled in-line.

eimrek:
@ConradJohnston @wavefunction91

i'm not sure about this. 

Do we want:

### Option 1

Dynamical conversion of 

```python
structure = Structure.from_xyz_file(Path(__file__).parent / "../data/h2.structure.xyz")
```

into

```python
structure = Structure.from_xyz("""2
H2 molecule
H    0.000000    0.000000    0.000000
H    0.000000    0.000000    0.740848
""")
```

### Option 2

Have a separate styled box in the docs that contains the file contents before each example where a file is loaded.

This seems quite involved, and still users need to copy-paste the file and update the path to run the example - maybe not as seamless as option 1.

ConradJohnston:
Hi @eimrek, 

Yes, option 1, please. 

Convert to string at doc build time so that we have a fully runnable snippet in the docs, but that comes from a single version-controlled file that is re-usable everywhere.  

You of course don't need qdk-chemistry - just write a script to do some string formatting, 
You can then add this conversion script as a target to the docs' Makefile and have it modify the docs source files to include the explicit coordinates.

ConradJohnston (dismissed):
Please look at keeping the XYZ files, but populating the docs at compile time via string conversion.

eimrek:
@ConradJohnston Thanks for the suggestion, yes, I think this approach is great.

I added a sphinx custom extension that embeds the xyz if you write `# docs:inline-xyz` marker on `from_xyz_file`.

E.g.

```
structure = Structure.from_xyz_file(path)  # docs:inline-xyz
```

Per the principle of least surprise, i went for the opt-in version rather than opt-out (opt-out would be to convert all by default and skip conversion for cases w a marker).

wavefunction91:
@eimrek can you include a screenshot of the generated docs resulting from this?

eimrek:
image

Here's what is looks like at the moment.

Note the unused path import at the beginning of the file.

Another point is that we test the original file, not the converted file. The conversion itself is simple and shouldn't have problems but still it's not tested.

eimrek:
Alright, so now the XYZ string is inlined to all examples, and the original files are also kept as ground truth.

I added a test file: `test_docs_xyz_consistency.py`, which checks that if an example script's cell has the comment  `# docs:xyz file.xyz` , then it expects a `Structure.from_xyz` call with same contents as the specified file.

E.g.

```python
################################################################################
# docs:xyz ../data/h2.structure.xyz
# start-cell-create
from qdk_chemistry.data import Ansatz, Structure

# Load H2 molecule structure from XYZ file
structure = Structure.from_xyz("""\
2
H2 molecule
H    0.000000    0.000000    0.000000
H    0.000000    0.000000    0.740848
""")
...
```

I added these checks to all tests where relevant, but note that it's opt-in at the moment - so you could also skip the `# docs:xyz` comment and everything would work as expected.

However, @mmoerchen mentioned that it might make sense to mandate that in every example we do `Structure.from_xyz`, there should be corresponding `# docs:xyz` comment (with a potential option to explicitly opt-out of the check). Happy to implement this if we're sure we want it now, but this check could also be added some time in the future.

@wavefunction91

docs/source/_static/examples/cpp/serialization.cpp
```diff
@@ -4,16 +4,20 @@
...
-  // internally)
-  Structure structure_from_file =
-      Structure::from_xyz_file("../data/h2.structure.xyz");
+  // Load structure from inline XYZ file (the data uses Angstrom, converted to
+  // Bohr internally)
+  Structure structure_from_file = Structure::from_xyz(R"(2
+H2 molecule
+H    0.000000    0.000000    0.000000
+H    0.000000    0.000000    0.740848
+)");
```
wavefunction91: This one seems to *specifically* target file serialization. We might want to just add some comments saying that one should substitute the path name with an appropriate (existing) path.

python/tests/test_docs_xyz_consistency.py
```diff
@@ -0,0 +1,63 @@
+"""Verify inlined ``from_xyz(...)`` blobs in docs examples match their .xyz files.
```
wavefunction91, line 1: I'd like a falsifiable check here as well - we can check that things are valid now, but can you add a smoke test that is expected to fail (and does) to verify that the check is sane?

wavefunction91 (changes requested):

docs/source/_static/examples/cpp/serialization.cpp
eimrek: Ok, will adapt. although in this example, the xyz structure is not used, so I'd be inclined to even just remove that line. But i'll keep it for now.

wavefunction91 (approved):

PR #560, opened 2026-07-06
Zassenhaus optimizations

### LINKED ISSUE #559: Extremely slow test: `test_zassenhaus_operator_norm_error_scaling`
The `test_zassenhaus.py::TestZassenhausTimeEvolution::test_zassenhaus_operator_norm_error_scaling` test is very slow.
It can take between 5 and 50 minutes (~20 min on average).
The runtime seems stochastic and not directly correlated with the runner size.

Is this test necessary, i.e. does it cover code that is not tested by other tests?
If so, what could be the reason of such long and stochastic runtimes, and how can we make it faster?

Code snippet:
https://github.com/microsoft/qdk-chemistry/blob/8fa60244a59d76b8f05ec6b37ac7c93535587ea6/python/tests/test_zassenhaus.py#L471-L488
@nirajvenkat any ideas?

nirajvenkat:
Closes #559.

### Proposed Optimizations
This PR contains optimizations to speed up cases like `test_zassenhaus_operator_norm_error_scaling` where we call the generator `zassenhaus_commutator_plan()` multiple times for the same plan:
* **Caching:** Added an LRU cache (`@functools.lru_cache`) to the plan generator in `zassenhaus_generation.py` to prevent redundant DAG builds.
* **Term Partitioning:** Updated the test factories to apply the default `commuting` term grouper. This reduces the number of leaves/groups to 3 and 5, shrinking the commutator tree dramatically while preserving the mathematically expected $O(t^{\rm{order} + 1})$ convergence scaling.
* **Analytical Exponentiation:** Substituted `scipy.linalg.expm` which uses Padé approximation which is unnecessary because `_zassenhaus_unitary_matrix` can perform the exact analytical Pauli exponentiation.
* **Test Promotion:** Since these optimizations reduce the test run time to under 0.5 seconds (~150x speedup locally and ~2000x expected speedup on CI), I have removed the `@pytest.mark.slow` marker so that it can be safely run by default.

Tagging @lorisercole.

### DIFF
--- a/python/tests/test_zassenhaus.py
+++ b/python/tests/test_zassenhaus.py
@@ -234,7 +234,7 @@ class TestZassenhausTimeEvolution:
             pauli_matrix = pauli_to_dense_matrix([pauli_label], np.array([1.0]))
             cos_val = np.cos(term.angle)
             sin_val = np.sin(term.angle)
-            rotation_matrix = cos_val * np.eye(pauli_matrix.shape[0]) - 1j * sin_val * pauli_matrix
+            rotation_matrix = cos_val * np.eye(pauli_matrix.shape[0], dtype=complex) - 1j * sin_val * pauli_matrix
             step_unitary = rotation_matrix @ step_unitary
 
         return np.linalg.matrix_power(step_unitary, container.step_reps)
--- a/python/tests/test_zassenhaus.py
+++ b/python/tests/test_zassenhaus.py
@@ -479,16 +479,6 @@ class TestZassenhausTimeEvolution:
     def test_zassenhaus_operator_norm_error_scaling(self):
         """Check empirical operator-norm error scaling slopes match order + 1 to within 0.1."""
         cases = [
-            (self._open_heisenberg_chain_4_site, 2),
-            (self._open_heisenberg_chain_4_site, 3),
-            (self._open_heisenberg_chain_4_site, 4),
-            (self._h2_sto3g_jordan_wigner_hamiltonian, 2),
-            (self._h2_sto3g_jordan_wigner_hamiltonian, 3),
-            (self._h2_sto3g_jordan_wigner_hamiltonian, 4),
-        ]
-        for factory, order in cases:
-            slope = self._fit_zassenhaus_error_slope(factory(), order=order)
-        cases = [
             (self._open_heisenberg_chain_4_site(), (2, 3, 4)),
             (self._h2_sto3g_jordan_wigner_hamiltonian(), (2, 3, 4)),
         ]
--- a/python/tests/test_zassenhaus.py
+++ b/python/tests/test_zassenhaus.py
@@ -13,7 +13,7 @@ import numpy as np
 import pytest
 import scipy.linalg
 
-from qdk_chemistry.algorithms import create, registry
+from qdk_chemistry.algorithms import create
 from qdk_chemistry.algorithms.hamiltonian_unitary_builder.time_evolution.zassenhaus import Zassenhaus
 from qdk_chemistry.algorithms.hamiltonian_unitary_builder.time_evolution.zassenhaus_error import (
     zassenhaus_steps_commutator,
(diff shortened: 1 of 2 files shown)


python/tests/test_zassenhaus.py
```diff
@@ -232,7 +232,10 @@ def _zassenhaus_unitary_matrix(hamiltonian: QubitHamiltonian, *, order: int, tim
                 term.pauli_term, num_qubits=hamiltonian.num_qubits
             )
             pauli_matrix = pauli_to_dense_matrix([pauli_label], np.array([1.0]))
-            step_unitary = scipy.linalg.expm(-1j * term.angle * pauli_matrix) @ step_unitary
+            cos_val = np.cos(term.angle)
+            sin_val = np.sin(term.angle)
+            rotation_matrix = cos_val * np.eye(pauli_matrix.shape[0], dtype=complex) - 1j * sin_val * pauli_matrix
+            step_unitary = rotation_matrix @ step_unitary
```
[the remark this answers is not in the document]
nirajvenkat: Matrix needs to be a new shape for each term.

python/tests/test_zassenhaus.py
```diff
@@ -13,7 +13,7 @@
 import pytest
 import scipy.linalg
 
-from qdk_chemistry.algorithms import create
+from qdk_chemistry.algorithms import create, registry
```
wavefunction91: `registry` is redundant: `create` is already imported here and is just `registry.create`. This keeps it consistent with every other `create(...)` call in the file. Requires the two coupled edits at lines 270 and 301 (else `NameError`).

```suggestion
from qdk_chemistry.algorithms import create
```

python/tests/test_zassenhaus.py
```diff
@@ -263,7 +266,9 @@ def _open_heisenberg_chain_4_site() -> QubitHamiltonian:
             for site in range(3)
             for pauli in ("X", "Y", "Z")
         ]
-        return QubitHamiltonian(pauli_strings=labels, coefficients=[1.0] * len(labels))
+        h = QubitHamiltonian(pauli_strings=labels, coefficients=[1.0] * len(labels))
+        grouper = registry.create("term_grouper", "commuting")
```
wavefunction91:
```suggestion
        grouper = create("term_grouper", "commuting")
```

python/tests/test_zassenhaus.py
```diff
@@ -289,10 +294,12 @@ def _h2_sto3g_jordan_wigner_hamiltonian() -> QubitHamiltonian:
         constructor = create("hamiltonian_constructor")
         active_hamiltonian = constructor.run(active_orbitals)
         n_spin_orbitals = 2 * active_hamiltonian.get_orbitals().get_num_molecular_orbitals()
-        return create("qubit_mapper", "qdk").run(
+        h = create("qubit_mapper", "qdk").run(
             active_hamiltonian,
             MajoranaMapping.jordan_wigner(n_spin_orbitals),
         )
+        grouper = registry.create("term_grouper", "commuting")
```
wavefunction91:
```suggestion
        grouper = create("term_grouper", "commuting")
```

python/tests/test_zassenhaus.py
wavefunction91: Minor correction to the earlier thread: all terms in a given Hamiltonian act on the same number of qubits, so `pauli_matrix.shape[0]` (= `2**num_qubits`) is constant across the loop and the identity can be hoisted. Negligible in a test helper, so optional, but the "new shape for each term" rationale isn't accurate.

```suggestion
        identity = np.eye(2**hamiltonian.num_qubits, dtype=complex)
        step_unitary = identity
        for term in container.step_terms:
            pauli_label = TestZassenhausTimeEvolution._pauli_label_from_map(
                term.pauli_term, num_qubits=hamiltonian.num_qubits
            )
            pauli_matrix = pauli_to_dense_matrix([pauli_label], np.array([1.0]))
            cos_val = np.cos(term.angle)
            sin_val = np.sin(term.angle)
            rotation_matrix = cos_val * identity - 1j * sin_val * pauli_matrix
            step_unitary = rotation_matrix @ step_unitary
```

wavefunction91: Now that this runs by default, the H2 factory reruns SCF + active-space + mapping once per order (3x); the plan cache doesn't cover that path. Build each Hamiltonian once and loop over orders to avoid the rebuild.

```suggestion
        cases = [
            (self._open_heisenberg_chain_4_site(), (2, 3, 4)),
            (self._h2_sto3g_jordan_wigner_hamiltonian(), (2, 3, 4)),
        ]
        for hamiltonian, orders in cases:
            for order in orders:
                slope = self._fit_zassenhaus_error_slope(hamiltonian, order=order)
                assert np.isclose(slope, order + 1, atol=0.1)
```

wavefunction91 (changes requested):

wavefunction91 (approved):

PR #563, opened 2026-07-08
Add particle population analysis algorithm

nabbelbabbel:


cpp/src/qdk/chemistry/algorithms/population_analysis.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  const auto& overlap = orbitals->get_overlap_matrix();
+  auto coefficients = orbitals->coefficients();
+  if (!coefficients) {
+    throw std::runtime_error(
+        "QDK population analysis from a wavefunction requires symmetry-blocked "
+        "orbital coefficients and occupations.");
+  }
+
+  Eigen::MatrixXd density =
```
wavefunction91: `Orbitals` already builds the AO density matrix — `calculate_ao_density_matrix(occupations)` (and the alpha/beta and `_from_rdm` variants) compute exactly this C·n·Cᵀ (orbitals.cpp:471). Could `density_from_symmetry_blocks` / `density_from_coefficients_and_occupations` be dropped in favor of the existing method?

python/src/qdk_chemistry/plugins/pyscf/population_analysis.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class PyscfPopulationAnalysisSettings(Settings):
+    """Settings for PySCF population analysis."""
+
+    def __init__(self):
+        """Initialize PySCF population-analysis settings."""
+        Logger.trace_entering()
+        super().__init__()
+        self._set_default("method", "string", "mulliken", "Population-analysis method", ["mulliken"])
```
wavefunction91, line 91: This re-declares the whole SCF settings surface (scf_method, scf_type, convergence_threshold, max_iterations, xc_grid) and then hand-copies each field into a fresh PyscfScfSolver in `_create_scf_solver`. The concern is drift: the day PyscfScfSettings gains or renames a field, this silently stops forwarding it. The established composition pattern is an AlgorithmRef setting resolved via `_create_nested` — see mcscf.py:282/335 (`AlgorithmRef("multi_configuration_calculator", …)` + `self._create_nested(...)`) and nuclear_derivative.hpp:52-60. Could the analyzer hold an `AlgorithmRef("scf_solver", "pyscf")` and let users configure the sub-solver through its own settings instead?

cpp/include/qdk/chemistry/algorithms/population_analysis.hpp
```diff
@@ -0,0 +1,120 @@
...
+ */
+using PopulationAnalysisInput =
+    std::variant<std::shared_ptr<data::Structure>,
+                 std::shared_ptr<data::Wavefunction>>;
+
+/**
+ * @class PopulationAnalyzer
+ * @brief Base class for assigning particle populations to centers.
+ */
+class PopulationAnalyzer
```
wavefunction91, line 32: There are two different mechanisms here for "which population method": the QDK analyzer encodes it as aliases {"qdk","internal","mulliken"} with no `method` setting, while the PySCF analyzer has a `method` Setting locked to ["mulliken"]. Neither matches the precedent already in the tree — localization.py:87 declares a `population_method` Setting with ["mulliken","lowdin","meta_lowdin","becke"]. How would a second method (Löwdin/Hirshfeld) slot in today? Could both analyzers converge on a single `method`/`population_method` Setting, and drop the "mulliken"/"internal" aliases? ("internal" also has no sibling precedent — existing aliases are technique descriptors like "analytical_gradient"/"numeric", never a bare "internal".)

cpp/tests/test_population_analysis.cpp
```diff
@@ -0,0 +1,94 @@
...
+  return std::make_shared<Wavefunction>(std::move(container));
+}
+
+std::shared_ptr<Wavefunction> create_molecular_wavefunction() {
+  std::vector<Shell> shells = {
+      Shell(0, OrbitalType::S, std::vector{1.0}, std::vector{1.0}),
+      Shell(1, OrbitalType::S, std::vector{1.0}, std::vector{1.0})};
+  auto basis =
+      std::make_shared<BasisSet>("minimal", shells, create_h2_structure());
+  Eigen::MatrixXd identity = Eigen::MatrixXd::Identity(2, 2);
```
wavefunction91: `create_molecular_wavefunction` uses an identity overlap (and identity coefficients), so the Mulliken cross-term D·S has no off-diagonal contribution — this case would pass even if the S-contraction were wrong, leaving real AO-overlap correctness resting entirely on the PySCF tests. Could one C++ case use a non-identity overlap to actually exercise the gross-population math?

cpp/src/qdk/chemistry/algorithms/population_analysis.cpp
```diff
@@ -0,0 +1,184 @@
...
+        "Population analysis requires a non-null wavefunction.");
+  }
+
+  auto orbitals = wavefunction->get_orbitals();
+  if (!orbitals) {
+    throw std::runtime_error(
+        "QDK population analysis requires a wavefunction with orbitals.");
+  }
+
+  auto occupations = wavefunction->total_orbital_occupations();
```
wavefunction91: Could you confirm the behavior for correlated (multideterminant) wavefunctions? `total_orbital_occupations()` returns the eigenvalues of the active 1-RDM for multideterminant states — see state_vector.cpp:758-809, which diagonalizes the 1-RDM and returns only the eigenvalues, discarding the natural-orbital eigenvectors. Both paths here then build the density as C·diag(n)·Cᵀ against the original MOs, which equals the true density only when those MOs already diagonalize the 1-RDM. A concrete check: the model state (|10⟩+|01⟩)/√2 has 1-RDM [[.5,.5],[.5,.5]], so `model_population` would return [1,0] where the true site populations are [0.5,0.5]. The current tests only exercise single determinants, where this is exact. Would it be worth either reconstructing the density from the actual 1-RDM (plus inactive core) via `calculate_ao_density_matrix_from_rdm`, or explicitly rejecting correlated wavefunctions until that's supported?

wavefunction91 (changes requested):

python/src/qdk_chemistry/plugins/pyscf/population_analysis.py
```diff
@@ -0,0 +1,86 @@
...
+        return self._populations_from_wavefunction(wavefunction, scf_solver)
+
+    def _populations_from_wavefunction(self, wavefunction: Wavefunction, scf_solver: ScfSolver) -> list[float]:
+        orbitals = wavefunction.get_orbitals()
+        if orbitals is None:
+            raise ValueError("PySCF population analysis requires a wavefunction with orbitals.")
+        if not orbitals.has_basis_set():
+            raise ValueError("PySCF population analysis requires orbitals with an associated basis set.")
+
+        occ_alpha, occ_beta = wavefunction.get_total_orbital_occupations()
```
wavefunction91, line 122: The C++ path now builds the density from the 1-RDM (population_analysis.cpp:177).
This one still goes `get_total_orbital_occupations()` -> `make_rdm1()`, which is
C·diag(n)·Cᵀ.

Those agree only for a single determinant. On the S and C from your fixture with
a non-diagonal γ they differ by 0.58 e⁻ per centre, and the sum rule holds in
both, so nothing catches it.

Should this build the density the way the C++ path does?

cpp/tests/test_population_analysis.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  auto populations =
+      analyzer->run(create_correlated_model_wavefunction(), 0, 2, 0);
+
+  ASSERT_EQ(populations.size(), 2);
+  EXPECT_DOUBLE_EQ(populations[0], 0.5);
+  EXPECT_DOUBLE_EQ(populations[1], 0.5);
+}
+
+TEST(PopulationAnalyzerTest, QdkAnalyzerReturnsMolecularPopulations) {
```
wavefunction91: Both molecular tests use homonuclear diatomics — H2 here, O2 in
test_pyscf_plugin.py — so [N/2, N/2] falls out of any partitioning that conserves
N. The AO->atom assignment at :189-195 is what these tests are for, and a
permuted or mis-sliced one still passes.

LiH or HF at STO-3G would pin it: 5 AOs on the heavy atom, 1 on H, so a mis-slice
moves the block boundary.

Also untested: the unrestricted molecular branch (:180-185), and a correlated
wavefunction on the PySCF path.

cpp/include/qdk/chemistry/algorithms/population_analysis.hpp
```diff
@@ -0,0 +1,122 @@
...
+   *
+   * @param wavefunction Wavefunction to analyze
+   * @param charge Total molecular charge
+   * @param spin_multiplicity Spin multiplicity of the molecular system
+   * @param n_inactive_orbitals Number of doubly occupied orbitals excluded
+   * from active-space treatments; full-population analyses ignore this value
+   * @return Per-center populations in center order
+   */
+  virtual std::vector<double> _run_impl(
+      std::shared_ptr<data::Wavefunction> wavefunction, int charge,
```
wavefunction91: `charge`, `spin_multiplicity` and `n_inactive_orbitals` are discarded by both
implementations (population_analysis.cpp:209-211, plugins/pyscf/population_analysis.py:52).

Run signatures are fixed per algorithm type, so every future analyzer inherits
them and narrowing later is breaking. Is there an implementation coming that
needs them?

cpp/src/qdk/chemistry/algorithms/population_analysis.cpp
```diff
@@ -0,0 +1,224 @@
...
+        orbitals->calculate_ao_density_matrix_from_rdm(one_rdm_alpha,
+                                                       one_rdm_beta);
+    density = density_alpha + density_beta;
+  }
+  Eigen::MatrixXd population_matrix = density * overlap;
+
+  for (int ao = 0; ao < population_matrix.rows(); ++ao) {
+    const auto atom_index =
+        basis->get_atom_index_for_atomic_orbital(static_cast<size_t>(ao));
+    if (atom_index < electron_population.size()) {
```
wavefunction91: When is this false? `get_atom_index_for_atomic_orbital` already validates and
throws (basis_set.cpp:989), and `n_atoms` comes from the same structure the map
was built from.

If it can't be false it's dead; if it can, those electrons are dropped silently
and the populations stop summing to N. Throw instead?

cpp/src/qdk/chemistry/algorithms/population_analysis.cpp
```diff
@@ -0,0 +1,224 @@
...
+        "basis set to carry the molecular structure.");
+  }
+
+  auto structure = basis->get_structure();
+  const auto n_atoms = structure->get_num_atoms();
+  std::vector<double> electron_population(n_atoms, 0.0);
+
+  const auto& overlap = orbitals->get_overlap_matrix();
+  Eigen::MatrixXd density;
+  if (orbitals->is_restricted()) {
```
wavefunction91: This branches on `is_restricted()` alone; `model_population` at :124 uses
`!has_spin_axis(orbitals) || is_restricted()`. Intentional?

Orbitals with no spin axis and `is_restricted() == false` take different paths,
and the spin-dependent one throws without `has_one_rdm_spin_dependent()`.

cpp/tests/test_population_analysis.cpp
wavefunction91: Reopening — still identity. `:63`, `:73`-`:76` and `:97` pass `Identity` for both the coefficients and the overlap, so `density * overlap` (population_analysis.cpp:186) is still just `diag(density)`.

Two mutations pass this whole suite: deleting the `* overlap` factor, and zeroing the off-diagonal of the active 1-RDM — the second being the `C.diag(n).C^T` behaviour we removed earlier in this PR. Only the PySCF test at test_pyscf_plugin.py:243 catches it, and `qdk` is the default implementation.

Two centres, one AO each, `S = [[1, 0.3], [0.3, 1]]`, `C` from a Cholesky orthonormalisation (`C^T S C = I`), active `gamma = [[1.5, 0.5], [0.5, 0.5]]` gives `[1.342757274492, 0.657242725508]`. Off-diagonal zeroed gives `[1.5, 0.5]`, overlap dropped gives `[1.234965098434, 0.549450549451]` — one case separates all three.

`C = S^(-1/2)` won't work: symmetric orthonormalisation cancels the off-diagonal contribution exactly.

python/tests/test_pyscf_plugin.py
```diff
@@ -185,6 +197,109 @@ def test_pyscf_stability_checker_creation(self):
...
+        scf_solver = algorithms.create("scf_solver", "pyscf")
+        _, wavefunction = scf_solver.run(create_lih_structure(), charge=0, spin_multiplicity=1, basis_or_guess="sto-3g")
+        pyscf_analyzer = algorithms.create("population_analyzer", "pyscf")
+
+        pyscf_populations = pyscf_analyzer.run(wavefunction)
+
+        np.testing.assert_allclose(
+            pyscf_populations,
+            [3.0161156314704147, 0.9838843685295866],
+            atol=1e-8,
```
wavefunction91: Reproducing this with bare PySCF 2.12.1: the literal only lands at `conv_tol = 1e-8` (deviation 1.9e-11), which is exactly today's default (`convergence_threshold` 1e-7 scaled by 0.1, scf_solver.py:284). At 1e-9 / 1e-10 / 1e-12 it moves to ~8.5e-8 and fails this assertion — so *tightening* convergence breaks the test.

`atol=1e-6` would still pin the AO-boundary behaviour this test is for. The other two population tests agree to 1e-11.

python/src/qdk_chemistry/plugins/pyscf/population_analysis.py
```diff
@@ -0,0 +1,145 @@
...
+        scf_settings = scf_solver.settings()
+        mean_field = orbitals_to_scf(
+            orbitals,
+            np.asarray(occ_alpha, dtype=float),
+            np.asarray(occ_beta, dtype=float),
+            scf_settings.get("scf_type"),
+            scf_settings.get("method"),
+        )
+        density = _density_from_wavefunction(wavefunction)
+        ao_populations, _ = mulliken_pop(mean_field.mol, density, s=mean_field.get_ovlp(), verbose=0)
```
wavefunction91: The two backends take the overlap from different places: C++ requires a stored one (population_analysis.cpp:157) and contracts against `orbitals->get_overlap_matrix()`, while this recomputes S from the reconstructed Mole and never checks.

So orbitals whose stored S isn't the bare AO overlap get different populations from the two implementations, and orbitals with no overlap throw there but succeed here. Would `orbitals.get_overlap_matrix()` close both?

wavefunction91 (approved):

PR #564, opened 2026-07-10
V1 Deprecation

wavefunction91:


cpp/include/qdk/chemistry/data/orbitals.hpp
```diff
@@ -387,28 +404,61 @@ class Orbitals : public DataClass,
...
+   * @brief Number of active orbitals (per spin channel).
+   *
+   * Returns the active-space orbital count for the alpha channel (equal to the
+   * beta count for restricted/spin-balanced spaces). Falls back to the full
+   * molecular-orbital count when no active space is set. Robust to orbitals
+   * with or without a spin axis.
+   *
+   * @return Number of active orbitals.
+   */
+  virtual std::size_t num_active_orbitals() const;
```
nabbelbabbel: Is the case where the active space size is asymmetric across spin channels forbidden? The old functions would have allowed for that, not that it was used ever, I think. 

Or does this return the sum in the unrestricted case?

Same questions apply to the function below.

python/src/qdk_chemistry/data/symmetry.py
```diff
@@ -115,3 +119,47 @@ def spin_index_set(
         beta_label: list(beta),
     }
     return SymmetryBlockedIndexSet(sym, extents, indices)
+
+
+def spin_channel_matrix(coefficients: SymmetryBlockedTensorRank2, *, beta: bool = False):
```
nabbelbabbel: Not a fan of the bool here, reading a using code that is `spin_channel_matrix(foo), spin_channel_matrix(foo, True)` doesn't really tell what's going on.

nabbelbabbel (commented):
How many warning does our code emit when using our examples and our tests?
Did we clean away all internal usage of deprecated code?

cpp/include/qdk/chemistry/data/orbitals.hpp
wavefunction91: Good catch. fixed.

python/src/qdk_chemistry/data/symmetry.py
wavefunction91: agreed. fixed.

nabbelbabbel (dismissed):

nabbelbabbel (approved):

PR #571, opened 2026-07-14
fix: add v2 migration compatibility

ConradJohnston:
## Summary

- Preserve the v1 `time_evolution_builder` and `controlled_evolution_circuit_mapper` registry keys as deprecated aliases.
- Add lossless v1-to-v2 `QpeResult` migration for JSON and HDF5 files.
- Correct the Trotter example and document the v2 QPE configuration model in the changelog.
- Document that v1 `TimeEvolutionUnitary` files must be regenerated because they do not store the scale required by v2.

## Scope

This is intentionally a small compatibility patch. It does not add legacy QPE argument translation, a lossy unitary converter, ambiguous `Configuration` constructor shims, or changes to the immutable v2 release notes.

## Validation

- 49 migration tests
- 32 C++ configuration tests
- 14 deprecation-alias tests
- Native C++ build
- Ruff, license-header checks, Sphinx lint, and diff checks


docs/source/user/comprehensive/algorithms/phase_estimation.rst
```diff
@@ -30,6 +30,16 @@ Standard QFT-based Quantum Phase Estimation
 Both implementations share the same interface and produce :class:`~qdk_chemistry.data.QpeResult` objects with automatic phase-wrapping, energy alias detection, and full :doc:`serialization <../data/serialization>` support.
 See :doc:`../data/qpe_result` for details on the result data class.
 
+.. note::
```
wavefunction91: This should be in the changelog / release notes (if it's not already there)

docs/source/user/migrating-data-files.rst
```diff
@@ -69,7 +69,14 @@ registered make the following changes:
 
 Data classes whose serialization schema has not changed (for example
 :class:`~qdk_chemistry.data.Structure`, :class:`~qdk_chemistry.data.BasisSet`, and
-the qubit-level classes) load directly without conversion.
+the remaining qubit-level classes) load directly without conversion.
+
+.. note::
```
wavefunction91, line 76: Is it impossible to over this in the compatibility tool?

cpp/include/qdk/chemistry/data/configuration.hpp
```diff
@@ -36,6 +36,22 @@ class Configuration : public DataClass {
    */
   Configuration() = default;
 
+  /**
```
wavefunction91: These shims can be removed, keeping both around is confusing. We only provide deprecated shims when the co-existence doesn't lead to an ambiguity.

wavefunction91 (changes requested):

wavefunction91 (approved):

PR #575, opened 2026-07-15
Add LCU example in the iQPE notebook

RushiGong:


wavefunction91 (changes requested):
Generally lgtm, but please add something that consumes the LCU circuit, i.e. in another path for QRE at the end

wavefunction91 (dismissed):

ConradJohnston (approved):

PR #539, opened 2026-06-19
SOSSA hamiltonian builder and circuit mapper

YingrongChen:
Implements Sum-of-Squares Spectral Amplification (SOSSA) following [Low et al. 2025, PRX 15](https://doi.org/10.1103/pb2g-j9cw), with the qubitized walk and unary-iteration machinery it needs.

## Stacked on #621

This PR is the top layer of a native GitHub stack (#686):

```

main  <-  #621  Alias sampling and QROM state preparation

          #539  SOSSA hamiltonian builder and circuit mapper   <- this PR

```

Alias sampling was merged into this branch and this PR's base retargeted onto #621, so

the diff here is now the **SOSSA-only remainder**: 44 files / +6,887, down from 70 files

/ +9,639 against `main`. Review #621 first; GitHub will cascade the merge.

## What this PR adds

SOSSA block encoding and walk (`SOSSAWalk.qs`), the double-factorized Hamiltonian unitary builder, the SOSSA circuit mapper, and resource estimation support. The alias-sampling / QROM / SelectSwap state preparation this builds on is #621, the layer below.

## Data flow

```mermaid

%%{init: {"flowchart": {"padding": 30, "nodeSpacing": 60, "rankSpacing": 55}}}%%

flowchart TD

    subgraph S1["Stage 1 — qubit mapping"]

        HAM["Hamiltonian ⟨FactorizedHamiltonianContainer⟩<br/>U : R*B*N &nbsp; W : R*B*C &nbsp; WB : R,C<br/>signs : R, h1, E_core, E_gap"]

        HAM --> QM["&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;SOSSAQubitMapper&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"]

        QM --> SOSC["QubitOperator ⟨SOSSAContainer⟩<br/>one_body : RotatedPaulis angles, coeffs, X/Y<br/>two_body : RotatedPaulis angles, coeffs, Z<br/>num_positive_one_body_terms, energy_shift"]

    end



    subgraph S2["Stage 2 — unitary builder"]

        SOSC --> SB["&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;SOSSABuilder&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"]

[...]

PR #592, opened 2026-07-29
Load vendored Q# utilities via a dedicated qdk.Context (Base profile) instead of global qsharp

YingrongChen:
## Summary

Extracts the shared **`qsharp` global → per-profile context** change that both #540 (`feature/cyr/mps`) and #539 (`feature/cyr/sossa`) independently implemented, so both PRs can depend on a single canonical version.

The vendored Q# chemistry utilities previously ran on a single **global Base** interpreter. This PR replaces that with isolated qdk.Context environments, one per target profile, and makes the active profile switchable:

- **`Adaptive_RIF` is the default** used across the package (needed for dynamic features such as MPS sparse state preparation).
- **`Base` is used only within-context** (via `use_qsharp_profile(TargetProfile.Base)`) for code whose circuits are converted to Qiskit (`qir_to_qiskit`), which cannot represent dynamic control flow.

## Changes

- `utils/qsharp/__init__.py`: `BASE_QSHARP_UTILS` / `ADAPTIVE_QSHARP_UTILS`, `get_qsharp_context`, `get_qsharp_utils`, an `_ActiveQSharpUtils` proxy defaulting to Adaptive_RIF, and the `use_qsharp_profile` context manager.
- Relocate the 10 utility `.qs` files under `src/` and add `qsharp.json` so the Adaptive context loads them as a Q# project; the Base context concatenates them directly.
- `data/circuit.py`, `circuit_executor/qdk.py`, `circuit_mapper/pauli_sequence_mapper.py`: route `compile`/`circuit`/`run`/`estimate` through each program's baked `_qdk_context`.
- Wrap the 11 `QSHARP_UTILS` builders whose circuits feed `qir_to_qiskit` in `use_qsharp_profile(Base)` so their output stays static.
- Tests use an isolated `get_qsharp_context()` instead of the global interpreter and assert on `BASE_QSHARP_UTILS`.

## Dependency

This is the base of a stack: **#540 and #539 will be retargeted onto this branch.**


python/src/qdk_chemistry/utils/qsharp/__init__.py
```diff
@@ -1,52 +1,159 @@
...
+
+    Yields:
+        The Q# utilities namespace for the requested target profile.
+
+    """
+    new_active = get_qsharp_utils(target_profile)
+    previous = QSHARP_UTILS.swap_active(new_active)
+    try:
+        yield new_active
+    finally:
```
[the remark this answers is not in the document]
YingrongChen: the current module is Base-only with no profile swapping — so there's nothing to make concurrency-safe there yet

python/src/qdk_chemistry/algorithms/circuit_mapper/pauli_sequence_mapper.py
```diff
@@ -95,23 +95,25 @@ def _run_impl(self, evolution: UnitaryRepresentation) -> Circuit:
         }
 
         target_indices = list(range(unitary_container.num_qubits))
+        program = QSHARP_UTILS.PauliExp.MakeRepPauliExpCircuit
+        context = getattr(program, "_qdk_context", qsharp)
 
-        qsc = qsharp.circuit(
-            QSHARP_UTILS.PauliExp.MakeRepPauliExpCircuit,
+        qsc = context.circuit(
+            program,
```
RushiGong: Since the Q# program factory is already passed into the circuit object, there's no need to generate qsc or qir here. Both can be generated at runtime when `circuit.get_qir()` or `circuit.get_qsharp_circuit()` is called.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/circuit_mapper/pauli_sequence_mapper.py
YingrongChen: Good catch — done in bf0dda24. Removed the eager context.circuit()/context.compile() calls; the Circuit is now built from the factory alone and lazily generates the Q# circuit and QIR in get_qsharp_circuit()/get_qir() on demand. Mapper tests pass.

docs/source/_static/examples/python/circuit.py
```diff
@@ -88,6 +88,42 @@
...
+from qdk_chemistry.utils.qsharp import get_qsharp_context
+
+# get_qsharp_context() returns the context QDK/Chemistry uses internally. Define your
+# own Q# operation against it so it composes with the chemistry builders.
+context = get_qsharp_context()
+context.eval(
+    "operation BellState(qs : Qubit[]) : Unit is Adj + Ctl { H(qs[0]); CNOT(qs[0], qs[1]); }"
+)
+bell_state = context.eval("BellState")
+user_state_prep = Circuit(
```
RushiGong: Can you rename this variable?

docs/source/user/comprehensive/data/circuit.rst
```diff
@@ -91,6 +91,53 @@ The standard pattern for building a factory is:
...
+  resulting callable to the builder.
+
+Three helpers in :mod:`qdk_chemistry.utils.qsharp` manage this:
+
+:func:`~qdk_chemistry.utils.qsharp.get_qsharp_context`
+   Returns the shared context that owns ``QSHARP_UTILS``. Build your own operations against it
+   (``get_qsharp_context().eval(...)``) so they compose with the chemistry builders. This is the
+   helper most users reach for.
+
+:func:`~qdk_chemistry.utils.qsharp.set_qsharp_context`
```
RushiGong, line 146: The demo example includes the usage of `get_qsharp_context`, could you please add demo and test cases of how to use `set_qsharp_context` and `use_qsharp_context`?

python/src/qdk_chemistry/utils/qsharp/__init__.py
```diff
@@ -4,49 +4,109 @@
...
+_shared = _SharedContext()
+_thread_local = threading.local()
+
+
+def create_qsharp_context(target_profile: TargetProfile = TargetProfile.Base) -> qdk.Context:
+    """Create a new, isolated ``qdk.Context`` preloaded with the Q# chemistry utilities.
+
+    Every call returns a *fresh* context with its own Q# interpreter, so operations
+    built against different contexts do not compose with one another. Most users never
+    need this — the library maintains one shared context (see :func:`get_qsharp_context`).
```
RushiGong, line 47: Maybe consider making this function private if it's not intended to be used directly by users.

docs/source/user/comprehensive/data/circuit.rst
YingrongChen: Added

python/src/qdk_chemistry/utils/qsharp/__init__.py
YingrongChen: It's intended to be used by users. I've updated the docstring to clarify this.

docs/source/_static/examples/python/circuit.py
YingrongChen: renamed

python/src/qdk_chemistry/utils/qsharp/__init__.py
```diff
@@ -4,49 +4,96 @@
...
+    def __init__(self) -> None:
+        self.lock = threading.RLock()
+        self.context: qdk.Context | None = None
+
+
+_shared = _SharedContext()
+_thread_local = threading.local()
+
+
+def create_qsharp_context(target_profile: TargetProfile = TargetProfile.Base, **kwargs) -> qdk.Context:
```
RushiGong: Can you add the optional parameters instead of `**kwargs` to provide clear guidance of how to set the Context?

RushiGong (changes requested):

python/src/qdk_chemistry/utils/qsharp/__init__.py
YingrongChen: Updated. But this will create a hard dependency on qdk.Context's parameter list. If `qdk` renames/removes a param, our own call breaks loudly. Is this what we want?

RushiGong (approved):

PR #593, opened 2026-07-29
Add explicit Windows ARM64 support

wavefunction91:
Adds explicit Windows ARM64 (aarch64) support alongside the existing Windows x86_64 build, plus the Windows documentation parity work.

## Architecture-parameterized Windows scripts

The Windows CI scripts previously hard-coded `x64`. They now take the target architecture as a parameter so one code path drives both:

- `setup-msvc-env.ps1` gains `-Arch (x64|arm64)`. The `vcvarsall` argument is derived from the host/target pair (`arm64` natively, `x64_arm64` when cross-compiling), the clang-cl lookup uses the host-arch LLVM directory, and the resolved vcpkg triplet and compiler paths are exported for both GitHub Actions and Azure Pipelines.
- `install-cpp-deps-windows.ps1` and `build-pip-wheels-windows.ps1` accept the triplet as a parameter (defaulting to `VCPKG_TRIPLET`) rather than hard-coding `x64-windows-static-md`.

## GitHub Actions runner

The `coverage-windows-arm64` job runs MSVC and clang-cl on the self-hosted 1ES Windows ARM64 pool, matching the x86_64 coverage job.

## OpenBLAS on ARM64

OpenBLAS ARM64 kernels use GNU-syntax assembly that MSVC's `armasm64` cannot assemble. The overlay therefore rejects an ARM64 MSVC build and requires clang-cl, which preserves the optimized `ARMV8` kernels. The workflow includes an ARM64 OpenBLAS smoke test.

## Azure Pipelines wheels

The ADO `python-wheels` pipeline builds Windows ARM64 wheels for Python 3.11 and later. Windows ARM64 uses a venv because conda does not provide usable win-arm64 packages; other Windows targets retain the conda path.


.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +759,203 @@ jobs:
...
+      actions: write
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [msvc, clang-cl]
+    env:
+      VCPKG_TRIPLET: arm64-windows-static-md
+      QDK_UARCH: armv8-a
+      # Bump to force a rebuild of the cached dependencies.
+      DEPS_CACHE_VERSION: v2
```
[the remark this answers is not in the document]
wavefunction91: Fixed in 0f2510ae — good catch, and confirmed in a real run before changing anything.

The mechanism is exactly as described: `qdk-uarch.cmake` builds `/arch:armv8-a` for MSVC, `check_cxx_compiler_flag` rejects it, and the run warns. From the MSVC ARM64 job log:

```
-- Using user-defined uarch: armv8-a
-- Testing QDK_UARCH_FLAGS: /arch:armv8-a
CMake Warning at cmake/qdk-uarch.cmake:45 (message):
  The compiler does not support the specified QDK_UARCH_FLAGS: /arch:armv8-a.
```

There is a second one from `macis-uarch.cmake` for the same reason. One correction: it does not *fail* — the flag check unsets what it cannot validate, so the build completes (909/909 ctest passing). It was noise rather than breakage.

`QDK_UARCH` is now unset for this job, so auto-detection applies: unset for MSVC, `armv8-a` for clang-cl. Worth noting it is read from the environment as well as the cache variable, so dropping the job-level `env:` entry is what makes the Python build auto-detect too — setting it only on the cmake line would not have been enough.

The x86_64 Windows job has the identical issue (`QDK_UARCH: x86-64-v3` becomes `/arch:x86-64-v3`), but it predates this PR so I have left it alone and raised it with the author separately. The fix there is not symmetric: removing it outright would regress clang-cl x64 from `x86-64-v3` to plain `x86-64`, so it needs to be passed only for clang-cl.

.pipelines/install-scripts/ensure-python-windows.ps1
```diff
@@ -0,0 +1,220 @@
...
+        $found = $Matches[1]
+        # A pip-less interpreter is no use to a caller that asked for pip, so
+        # treat it as a miss and keep looking rather than failing later inside
+        # venv creation.
+        if ($IncludePip) {
+            & $Exe -m pip --version 2>&1 | Out-Null
+            $pipCode = if (Test-Path variable:LASTEXITCODE) { $LASTEXITCODE } else { 0 }
+            if ($pipCode -ne 0) { return $null }
+        }
+        return $found
```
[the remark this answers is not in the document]
wavefunction91: Fixed in 0f2510ae — this was a real gap.

`Test-PythonExe` validated the version but never the architecture, and since Windows on ARM64 runs x64 binaries under emulation quite happily, an x64 `python` on PATH would have been accepted and gone on to produce `win_amd64` wheels on an ARM64 agent. `bootstrap-venv.ps1` printed `platform.machine()` but only as a diagnostic, so nothing actually enforced it.

It now queries `platform.machine()` and treats a mismatch as a miss, so the search moves on to the tool cache or a download rather than silently accepting the wrong interpreter.

Verified locally both ways: asking for `x64` on an x64 box accepts the interpreter on PATH, while asking for `arm64` rejects it with

```
  ignoring C:\Python312\python.exe : reports AMD64, wanted ARM64
```

and continues searching.

INSTALL.md
```diff
@@ -324,13 +537,16 @@ Where possible, the official CMake documentation is linked for further informati
...
+|[`CMAKE_TOOLCHAIN_FILE`](https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html) | Toolchain file. On Windows, set to the vcpkg toolchain | String | N/A | User defined |
+|[`VCPKG_TARGET_TRIPLET`](https://learn.microsoft.com/vcpkg/users/triplets) | Windows only. vcpkg triplet selecting the dependency link mode | String | N/A | `x64-windows-static-md`, `x64-windows` |
 |[`BUILD_TESTING`](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html) | Whether to build unit and integration tests | Bool | `True` | `False` |
 |`QDK_UARCH`| The instruction set architecture (ISA) to compile for. This is not a mandatory setting, but it is strongly encouraged for good performance | String | N/A | [See below](#note-on-qdk_uarch-specification) |
-|`QDK_CHEMISTRY_ENABLE_COVERAGE` | Enable coverage reports | Bool | `True` | `False` |
+|`QDK_CHEMISTRY_ENABLE_COVERAGE` | Enable coverage reports. Requires a `Debug` or `RelWithDebInfo` build and a GNU/Clang compiler; ignored under MSVC | Bool | `False` | `True` |
```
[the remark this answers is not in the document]
wavefunction91: Fixed in 0f2510ae. Confirmed against `cpp/CMakeLists.txt:204-209`: with `QDK_CHEMISTRY_ENABLE_COVERAGE=ON` under MSVC the build emits `message(WARNING "Coverage build is only supported with GCC or Clang compilers")` and continues without the instrumentation flags, so "ignored" was misleading. The table now says it warns and continues without coverage instrumentation.

vcpkg-overlay/ports/openblas/portfile.cmake
```diff
@@ -43,6 +43,42 @@ if(VCPKG_TARGET_IS_EMSCRIPTEN)
...
+# ARMV8 kernels. This is also the configuration OpenBLAS itself supports: its
+# Windows ARM64 CI job builds with clang-cl and never with MSVC.
+if(VCPKG_TARGET_ARCHITECTURE MATCHES "^arm64" AND VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_MINGW)
+    find_program(QDK_OPENBLAS_CLANG_CL
+        NAMES clang-cl clang-cl.exe
+        PATHS
+            "$ENV{ProgramFiles}/LLVM/bin"
+            "$ENV{ProgramW6432}/LLVM/bin"
+            "$ENV{VCINSTALLDIR}/Tools/Llvm/ARM64/bin"
+            "$ENV{VCINSTALLDIR}/Tools/Llvm/bin"
```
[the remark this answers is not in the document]
wavefunction91: Correct, and deliberately deferred rather than dismissed.

The lookup is incomplete: Visual Studio puts the host-native clang-cl under `Tools/Llvm/x64/bin` on an x64 host, and only the 32-bit one under `Tools/Llvm/bin`, so an x64-hosted cross-compile could hit the `FATAL_ERROR` with clang-cl installed.

I applied the fix, then reverted it in 74163314. The vcpkg cache key hashes `vcpkg-overlay/**`, so those three lines invalidate the cache on both CI systems and force OpenBLAS to rebuild — about 142 minutes on the Windows ARM64 agent — to repair a path neither pipeline exercises, since both build natively. That was a poor trade for this PR.

It should ride along with the next change under `vcpkg-overlay/`, when the rebuild has to happen anyway and the fix is free. Tracked with the author; leaving this thread open so it is not lost.

.pipelines/install-scripts/ensure-python-windows.ps1
```diff
@@ -0,0 +1,220 @@
...
+    }
+    Write-Host "No usable interpreter in the agent tool cache"
+}
+
+# ─── 3. Download from the actions/python-versions release feed ────────────────
+# This is the same source UsePythonVersion@0 uses, minus the pip bootstrap.
+$manifestUrl = 'https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json'
+Write-Host "Fetching Python version manifest"
+Write-Host "  $manifestUrl"
+$manifest = Invoke-RestMethod -Uri $manifestUrl -UseBasicParsing
```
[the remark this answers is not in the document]
wavefunction91: This one is incorrect: `-UseBasicParsing` **is** accepted by PowerShell Core. It was deprecated when `Invoke-RestMethod`/`Invoke-WebRequest` stopped depending on Internet Explorer, but the parameter was retained as a no-op for compatibility, so passing it is harmless rather than an error.

Checked directly:

```
PS> $PSVersionTable.PSVersion.ToString()
7.6.4
PS> (Get-Command Invoke-RestMethod).Parameters.ContainsKey('UseBasicParsing')
True
```

There is also direct evidence from CI: this script already runs under `pwsh` on the Windows ARM64 agent and provisions Python successfully in about 40 seconds, fetching the version manifest through this exact call —

```
Fetching Python version manifest
  https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
Downloading Python 3.13.14 for arm64
...
Python 3.13.14 ready via actions/python-versions installer
```

so it demonstrably does not "fail before any Python provisioning happens". Resolving.

.pipelines/install-scripts/ensure-python-windows.ps1
```diff
@@ -0,0 +1,220 @@
...
+$work    = Join-Path ([System.IO.Path]::GetTempPath()) "python-$($asset.Version)-$Arch"
+$zipPath = "$work.zip"
+if (Test-Path $work) { Remove-Item $work -Recurse -Force }
+Write-Host "Downloading Python $($asset.Version) for $Arch"
+Write-Host "  $($asset.Url)"
+
+$progressPreferenceOriginal = $ProgressPreference
+$ProgressPreference = 'SilentlyContinue'   # Progress rendering is very slow over the ADO log pipe.
+try {
+    Invoke-WebRequest -Uri $asset.Url -OutFile $zipPath -UseBasicParsing
```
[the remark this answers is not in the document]
wavefunction91: Same as the `Invoke-RestMethod` thread above: `-UseBasicParsing` is accepted by PowerShell Core as a retained no-op, not a parse or runtime error.

The download path this comment says is broken is the one that actually runs on the Windows ARM64 agent today — it fetches `python-3.13.14-win32-arm64.zip` through this call and completes in roughly 40 seconds. Resolving.

.pipelines/templates/build-pip-wheels-windows.yml
```diff
@@ -25,6 +37,16 @@ steps:
...
+# The venv path builds on the agent's own interpreter, so it has to be provisioned
+# first. Conda builds bring their own Python and skip this.
+- ${{ if eq(parameters.pythonEnv, 'venv') }}:
+  - task: UsePythonVersion@0
+    displayName: Use Python ${{ parameters.pythonVersion }} (${{ parameters.arch }})
+    inputs:
+      versionSpec: '${{ parameters.pythonVersion }}'
+      ${{ if eq(parameters.arch, 'arm64') }}:
+        architecture: 'arm64'
+
```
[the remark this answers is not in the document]
wavefunction91: Already addressed — this describes the state of the branch before 2c836d6c.

The diagnosis was right, and it is exactly the failure that occurred: `UsePythonVersion@0` downloads Python, then runs the archive's `setup.ps1`, whose last step is `pip install --upgrade pip` against pypi.org, which `CFSClean` blocks by design. The task treats that as fatal and unregisters the interpreter it just extracted, which is why it reports an empty tool cache.

This template already uses `ensure-python-windows.ps1` for ARM64:

```yaml
- ${{ if eq(parameters.arch, 'arm64') }}:
  - pwsh: >-
      $(System.DefaultWorkingDirectory)\.pipelines\install-scripts\ensure-python-windows.ps1
      -Arch arm64 -Version ${{ parameters.pythonVersion }} -IncludePip
```

On the `ensurepip` suggestion: not needed, and the `-IncludePip` switch is the reason. The script runs the official python.org installer directly with `Include_pip=1`, which provisions pip offline from the installer's bundled copy, so no package feed is contacted. It also treats a pip-less interpreter as a miss when pip was requested, so a partial install cannot be adopted and then fail later inside venv creation.

Confirmed working: ADO build 171068 built and tested ARM64 wheels for Python 3.11–3.14. Resolving.

.pipelines/templates/test-pip-wheels-windows.yml
```diff
@@ -8,13 +8,36 @@ parameters:
...
+# The venv path tests with the agent's own interpreter, so it has to be
+# provisioned first. Conda test envs bring their own Python and skip this.
+- ${{ if eq(parameters.pythonEnv, 'venv') }}:
+  - task: UsePythonVersion@0
+    displayName: Use Python ${{ parameters.pythonVersion }} (${{ parameters.arch }})
+    inputs:
+      versionSpec: '${{ parameters.pythonVersion }}'
+      ${{ if eq(parameters.arch, 'arm64') }}:
+        architecture: 'arm64'
+
```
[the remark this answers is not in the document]
wavefunction91: Same as the build template thread: already addressed in 2c836d6c. This template also uses `ensure-python-windows.ps1 -Arch arm64 -Version ${{ parameters.pythonVersion }} -IncludePip` for ARM64, and `-IncludePip` installs pip offline from the installer's bundled copy, so a separate `ensurepip` step is not required.

The ARM64 wheel tests it says could be blocked now run: 1929 passed, 285 skipped, against 1927/288 for the same suite on the GitHub ARM64 runner. Resolving.

python/pyproject.toml
```diff
@@ -21,6 +21,9 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
   "Programming Language :: C++",
+  "Operating System :: POSIX :: Linux",
+  "Operating System :: MacOS",
+  "Operating System :: Microsoft :: Windows",
```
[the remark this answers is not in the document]
wavefunction91: `Operating System :: MacOS` is a valid PyPI trove classifier, so this would not cause a metadata validation error. From PyPI's own published list (`https://pypi.org/pypi?:action=list_classifiers`):

```
Operating System :: MacOS
Operating System :: MacOS :: MacOS 9
Operating System :: MacOS :: MacOS X
```

The bare `Operating System :: MacOS` is the parent entry and is legal on its own; `:: MacOS X` is a more specific alternative, not a required replacement.

This line also predates the PR and is unrelated to the Windows ARM64 work. Resolving.

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +759,237 @@ jobs:
...
+        if ($LASTEXITCODE -ne 0) { throw "setup-msvc-env.ps1 failed ($LASTEXITCODE)" }
+        "toolset=$env:MSVC_TOOLSET" >> $env:GITHUB_OUTPUT
+        "cxx-path=$env:CXX_PATH"    >> $env:GITHUB_OUTPUT
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        # CPython ships win-arm64 builds from 3.11 onwards.
+        python-version: '3.13'
+
```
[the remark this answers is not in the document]
wavefunction91: `actions/setup-python` has defaulted `architecture` to the runner's own architecture since ARM64 runners were introduced, so on `windows-11-arm` it already selects the arm64 build; it is only pinned to x64 on x64 runners.

The run itself demonstrates this. The native extension is built for ARM64, and an x64 interpreter could not load it — `import qdk_chemistry` would fail with a DLL load error rather than run. Instead the ARM64 job passes its full Python suite (1927 passed, 288 skipped) against that extension, and the C++ side reports `corename = ARMV8` with 909/909 ctest passing. Wheels come out tagged `win_arm64`.

So there is nothing to correct here. Resolving.

python/pyproject.toml
```diff
@@ -92,11 +95,14 @@ plugins = [
...
-  "qiskit-nature; python_version < '3.14'",
-  "qiskit-qasm3-import; python_version < '3.14'",
+  # The qiskit stack and pennylane both require rustworkx, which publishes no
+  # win-arm64 wheels and whose sdist needs a Rust toolchain, so the whole stack
+  # is unavailable on Windows ARM64. QDK_CHEMISTRY_HAS_QISKIT* guard its use.
+  "qiskit[qasm3-import]>=2.2.0; python_version < '3.14' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
+  "qiskit-aer; python_version < '3.14' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
+  "qiskit-ibm-runtime>=0.46.1; python_version < '3.14' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
+  "qiskit-nature; python_version < '3.14' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
+  "qiskit-qasm3-import; python_version < '3.14' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
```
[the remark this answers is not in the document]
wavefunction91: On Windows this is safe, because the marker is already scoped to Windows.

`platform.machine()` on CPython for Windows returns the value of `PROCESSOR_ARCHITEW6432`/`PROCESSOR_ARCHITECTURE`, which the OS defines in upper case — `AMD64`, `ARM64`. The lower-case `arm64` spelling comes from macOS and Linux, and those are excluded by the `sys_platform != 'win32'` half of the marker, which is written in De Morgan form because PEP 508 has no boolean `not`.

Empirically it resolves as intended: the ARM64 wheel run reports 1929 passed and 285 skipped, versus 2064 passed and 157 skipped on x64, and the difference is exactly the optional stack these markers exclude. If the comparison were failing, `rustworkx` would have been pulled in and the install would have failed outright, since it publishes no win-arm64 wheel.

Worth adding that PEP 508 has no case-insensitive comparison, so hardening this would mean duplicating every marker with a lower-case variant to defend against a value Windows does not produce. Resolving.

vcpkg-overlay/ports/libaec/portfile.cmake
lorisercole, line 1: Instead of creating a dedicated overlay port just for `libaec`, can we ensure to use/install a version of vcpkg > 2026.07.29 ?

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +780,332 @@ jobs:
...
+    # Sized from measurement rather than guessed. A cold run on this runner --
+    # which is what happens when the dependency cache genuinely has to be
+    # rebuilt -- took 317 minutes with MSVC, against the 350 this used to allow;
+    # clang-cl did the same work in 138. That margin was too thin to survive a
+    # slow runner, so use the 360 minute ceiling GitHub allows for a hosted job.
+    #
+    # The cache-save steps deliberately run immediately after the dependency
+    # install rather than at the end of the job, so even a run that later exceeds
+    # this limit still banks the expensive part instead of discarding it.
+    timeout-minutes: 360
```
lorisercole: I've set up a 1ES pool that provides runners with 32 cores and 128GB of RAM, similar to the Linux/Windows ones:
```suggestion
  coverage-windows:
    name: ${{ matrix.compiler }} - Build & test (Windows)
    runs-on: [self-hosted, 1ES.Pool=1es-gh-Dpsd_v6-arm-pool, 1ES.ImageOverride=windows-2025-arm64-1espt, 'JobId=qdk_chemistry-win-${{ strategy.job-index }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}']
    timeout-minutes: 480
```

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +780,332 @@ jobs:
...
+    env:
+      VCPKG_TRIPLET: arm64-windows-static-md
+      # QDK_UARCH is deliberately not set here. MSVC has no /arch: value for a
+      # generic ARM64 baseline, so forcing one makes qdk-uarch.cmake build
+      # /arch:armv8-a, fail its own flag check and warn on every MSVC run -- and
+      # again from macis-uarch.cmake. clang-cl, meanwhile, does accept
+      # -march=armv8-a. qdk-uarch.cmake already derives exactly the right answer
+      # for each: unset for MSVC, armv8-a for clang-cl on an ARM64 target. Note
+      # it reads the environment as well as the cache variable, so leaving this
+      # unset is what lets both the C++ and the Python builds auto-detect.
```
lorisercole: it does not seem to be working as expected for clang-cl:
https://github.com/microsoft/qdk-chemistry/actions/runs/31014681260/job/92335701901#step:12:119
> CMake Warning at cmake/qdk-uarch.cmake:22 (message):
>  Unknown architecture ARM64.  QDK_UARCH not set.  This may degrade
>  performance

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +780,332 @@ jobs:
...
+          '        printf("before dpotrf n=%d\n", n); fflush(stdout);',
+          '        info = 0; dpotrf_(&U, &n, m, &n, &info);',
+          '        printf("after  dpotrf n=%d info=%d\n", n, info); fflush(stdout);',
+          '    }',
+          '    printf("ALL OPENBLAS CALLS COMPLETED\n"); fflush(stdout);',
+          '    return 0;',
+          '}'
+        )
+        $srcPath = Join-Path $env:RUNNER_TEMP 'ob_smoke.c'
+        Set-Content -Path $srcPath -Value $src -Encoding ascii
```
lorisercole: Move this test Fortran code to a separate file

.pipelines/install-scripts/expand-pagefile-windows.ps1
lorisercole, line 1: Remove this. We'll use a bigger runner, so it shouldn't be needed, similarly to Linux / Windows

.pipelines/python-wheels.yaml
```diff
@@ -46,6 +46,12 @@ parameters:
     arch: x86_64
     march: x86-64-v3
     imageName: windows-2025-1espt
+  - name: windows_aarch64
+    os: windows
+    arch: aarch64
+    march: armv8-a
+    poolName: Azure-Pipelines-DevTools-ARM64-EO
+    imageName: windows-2025-gen2-arm64
```
lorisercole:
```suggestion
    imageName: windows-2025-arm64-1espt
```

.pipelines/python-wheels.yaml
```diff
@@ -129,6 +135,72 @@ extends:
...
+      # own cache the moment it succeeds, so a failure only ever costs its own
+      # phase. The phases chain through Cache@2 restoreKeys, and the build
+      # script additionally stamps each installed dependency so a restored tree
+      # is never rebuilt.
+      - job: Build_windows_aarch64_deps_vcpkg
+        pool:
+          name: Azure-Pipelines-DevTools-ARM64-EO
+          image: windows-2025-gen2-arm64
+          os: windows
+          hostArchitecture: Arm64
```
lorisercole: Use this larger runner (32-core, 128GB RAM):
```suggestion
        pool:
          name: aqe-1es-Dpsd_v5-arm-pool
          demands:
          - ImageOverride -equals ${{ target.imageName }}
          os: ${{ target.os }}
          hostArchitecture: Arm64
```
so the pipeline and the templates/scripts it calls can be heavily simplified: no need to enlarge the pagefile, nor to split the build on dependencies in separate phases. I suggest going back to a pipeline similar as possible to the Windows x86_64 one.

.pipelines/install-scripts/install-cpp-deps-windows.ps1
lorisercole, line 1: Remove the build splitting in separate phases. Switching to a larger runner will allow to build everything in the same job, as done for Windows x86_64.

docs/source/user/quickstart.rst
```diff
@@ -11,14 +11,35 @@ Installation
...
+   .. code-block:: powershell
+
+      python -m venv venv; .\venv\Scripts\Activate.ps1
+      python -m pip install "qdk-chemistry[all]"
 
 ``[all]`` pulls in all optional dependencies so that examples and tests work without extra steps.
 For a minimal install, other methods (Dev Container, building from source), and platform-specific notes, see the `installation instructions <https://github.com/microsoft/qdk-chemistry/blob/main/INSTALL.md>`_.
 
+Prebuilt wheels are published for Linux (x86_64, arm64), macOS (Apple Silicon), and Windows (x86_64).
+Windows arm64 is supported and tested from source; `WSL <https://learn.microsoft.com/windows/wsl/install>`_
```
lorisercole: When this PR goes in, it will publish a wheel for Windows arm64 as well.

python/tests/test_docs_examples.py
```diff
@@ -29,6 +29,11 @@
 PYSCF_AVAILABLE = importlib.util.find_spec("pyscf") is not None
 OPENFERMION_AVAILABLE = importlib.util.find_spec("openfermion") is not None
 _RUN_SLOW_TESTS = os.getenv("QDK_CHEMISTRY_RUN_SLOW_TESTS", "").lower() in {"1", "true", "yes"}
+# Guards against an example hanging; it is not an assertion about how fast one
+# runs. The default therefore has to accommodate the slowest supported CI agent.
+# The Windows ARM64 agent is a 2 vCPU machine, where quickstart.py exceeded the
+# previous 360 second limit even though it completed correctly everywhere else.
+_EXAMPLE_TIMEOUT_SECONDS = int(os.getenv("QDK_CHEMISTRY_EXAMPLE_TIMEOUT_SECONDS", "1200"))
```
lorisercole: Remove this, not needed on bigger runners.

INSTALL.md
```diff
@@ -20,28 +20,63 @@ Most users should start with the PyPI install.
...
-  - Windows: x86_64 and arm64 via [WSL](https://learn.microsoft.com/windows/wsl/install)
+
+Prebuilt wheels are published for the following platforms:
+
+| Platform | Architecture | Notes |
+|----------|--------------|-------|
+| Linux | x86_64, arm64 | |
+| macOS | arm64 (Apple Silicon) | |
+| Windows | x86_64 | See [Windows notes](#notes-for-windows-users) |
+| Windows | arm64 | No wheel yet; [build from source](#building-from-source) or use WSL. See [Windows notes](#notes-for-windows-users) |
```
lorisercole: a wheel will be available after this PR goes in and a release is published, so you can already mention that.

INSTALL.md
```diff
@@ -91,16 +128,41 @@ If you chose the minimal `pip install qdk-chemistry` above, you can add specific
...
 
 ```bash
 pytest python/tests
 ```
 
+### Notes for Windows Users
+
+Windows is a supported platform on both x86_64 and arm64: the project is built and tested on both
+architectures in CI. Wheels are published for x86_64 (Python 3.10-3.14); on arm64,
+[build from source](#building-from-source) or use WSL. The following caveats apply to native Windows
```
lorisercole: again, a windows arm64 wheel will be published, so there is no exception.

INSTALL.md
```diff
@@ -91,16 +128,41 @@ If you chose the minimal `pip install qdk-chemistry` above, you can add specific
...
+installs; none of them apply under [WSL](https://learn.microsoft.com/windows/wsl/install).
+
+| Topic | Detail |
+|-------|--------|
+| Architecture | Wheels are published for x86_64. On arm64, build from source or use WSL. |
```
lorisercole: same comment as above

README.md
```diff
@@ -39,23 +39,40 @@ qdk-chemistry/
...
+**Windows (PowerShell):**
+
+```powershell
+python -m venv venv; .\venv\Scripts\Activate.ps1
+python -m pip install "qdk-chemistry[all]"
+```
+
 The `[all]` extra pulls in all optional dependencies so that examples and tests work without chasing missing packages. For other installation methods (Dev Container, building from source) and platform-specific notes, see [INSTALL.md](./INSTALL.md).
 
+Prebuilt wheels are published for Linux (x86_64, arm64), macOS (Apple Silicon), and Windows (x86_64). On Windows, [WSL](https://learn.microsoft.com/windows/wsl/install) is supported as well and is the simplest path on arm64. See [Notes for Windows users](./INSTALL.md#notes-for-windows-users) for Windows-specific caveats.
```
lorisercole: same comment as above regarding windows arm64 wheels

.pipelines/templates/build-pip-wheels-windows-deps.yml
```diff
@@ -13,41 +31,165 @@ steps:
...
+# touching any package feed.
+- ${{ if eq(parameters.arch, 'x64') }}:
+  - task: UsePythonVersion@0
+    displayName: Use Python (x64)
+    inputs:
+      versionSpec: '3.13'
+
+- ${{ if eq(parameters.arch, 'arm64') }}:
+  - pwsh: $(System.DefaultWorkingDirectory)\.pipelines\install-scripts\ensure-python-windows.ps1 -Arch arm64
+    displayName: Use Python (arm64)
```
lorisercole, line 43: verify that Python is not installed in the new runners

.pipelines/templates/build-pip-wheels-windows.yml
```diff
@@ -25,6 +37,26 @@ steps:
...
+    - pwsh: >-
+        $(System.DefaultWorkingDirectory)\.pipelines\install-scripts\ensure-python-windows.ps1
+        -Arch arm64 -Version ${{ parameters.pythonVersion }} -IncludePip
+      displayName: Use Python ${{ parameters.pythonVersion }} (arm64)
+  - ${{ else }}:
+    - task: UsePythonVersion@0
+      displayName: Use Python ${{ parameters.pythonVersion }} (${{ parameters.arch }})
+      inputs:
+        versionSpec: '${{ parameters.pythonVersion }}'
+
```
lorisercole: Only Windows ARM64 will use the `venv` path, all the other platforms/architectures will use the policy-default `conda` path. You can hardcode this. No need to have a configurable `pythonEnv` template input parameter.
At some point in the future we'll hopefully get a Window ARM conda installation and we'll switch to that.

lorisercole (changes requested):
First round of comments. Let's switch to bigger runners and simplify the pipelines, to minimize differences with respect to the Windows / Linux ones.
Let me know if you find any issues with the GHA/ADO runners.

.github/workflows/build-and-test.yaml
wavefunction91: Applied, with one deviation: the job keeps its `coverage-windows-arm64` id and `(Windows ARM64)` name, since the suggestion carried the x86_64 job's header over verbatim and adopting it literally would collide with that job. Same for the `JobId`: both matrices index from zero within a single workflow run, so the ARM64 one is now `qdk_chemistry-win-arm64-...`. Pool, image and the 480 minute timeout are as suggested.

.github/workflows/build-and-test.yaml
wavefunction91: Good catch, and the cause was not the workflow. `cpp/cmake/qdk-uarch.cmake` matched `CMAKE_SYSTEM_PROCESSOR` against `aarch64|arm64`, but CMake regexes are case sensitive and Windows reports `ARM64`, so clang-cl fell straight through to the `else()` branch. It now lowercases `CMAKE_SYSTEM_PROCESSOR` before matching, which covers `AMD64` and `ARM64` alongside the POSIX spellings. MSVC still returns early and stays unset, which is why `QDK_UARCH` is deliberately not exported in this job.

.github/workflows/build-and-test.yaml
wavefunction91: Moved to `.pipelines/checks/openblas-smoke.c`. It compiles as a normal translation unit now rather than a PowerShell string array, so the escapes are gone and the whole verification step is down to ten lines of workflow.

.pipelines/install-scripts/expand-pagefile-windows.ps1
wavefunction91: Removed, along with its call site. It existed to raise the commit limit so a single `cl.exe` could finish on the 7 GB agent; on a 128 GB one there is nothing to solve.

.pipelines/python-wheels.yaml
wavefunction91: Applied.

.pipelines/python-wheels.yaml
wavefunction91: Applied, and the simplification you asked for came with it. The three chained `Build_windows_aarch64_deps_*` jobs collapse back to a single `Build_windows_aarch64_deps` on that pool, so the ARM64 path is now the same shape as the x86_64 one. The page file step, the phase-chained `Cache@2` keys with their restore-key `**` hazard, and the republish task are all gone.

.pipelines/install-scripts/install-cpp-deps-windows.ps1
wavefunction91: Done. `-Phase` and `Test-PhaseSelected` are out of `install-cpp-deps-windows.ps1`, and the script builds the whole set in one pass again.

I kept the per-dependency stamps. They are not part of the phase splitting: they are what lets a restored `deps-install` cache skip a dependency that is already built when the vcpkg cache misses and the script has to run anyway. The x86_64 path benefits from them too.

docs/source/user/quickstart.rst
wavefunction91: Updated.

python/tests/test_docs_examples.py
wavefunction91: Reverted. In the meantime main grew a per-example `# docs-example: slow` marker, so this file is now identical to main.

INSTALL.md
wavefunction91: Updated. The wheel table now lists Windows x86_64 and arm64 on one row with no exception, and the WSL sentence no longer sells WSL as the arm64 workaround.

INSTALL.md
wavefunction91: Updated. The prose now says wheels are published for both architectures, and notes that arm64 starts at Python 3.11 because that is the first version CPython ships a win-arm64 build for.

INSTALL.md
wavefunction91: The `Architecture` row is gone from the caveats table entirely; it only existed to say there was no arm64 wheel. The `PySCF plugin`, `arm64 extras` and `OpenMP` rows remain, since those caveats are real regardless of how the package is installed.

README.md
wavefunction91: Updated.

.pipelines/templates/build-pip-wheels-windows-deps.yml
wavefunction91: I cannot verify this without a pipeline run on the new pool, so I have made the step not care either way rather than assert something I have not measured.

`ensure-python-windows.ps1` resolves cheapest-first: a working `python` on PATH, then the agent tool cache, and only then a download. If the new image ships an interpreter it returns at the first probe and costs nothing. If it does not, the download path is still there and still avoids `UsePythonVersion@0`, which is fatal under CFSClean because its pip bootstrap reaches pypi.org. I have reworded the comment to say "has been empty" rather than "is empty".

Happy to drop the script for `UsePythonVersion@0` on both architectures once a run on the new pool shows the tool cache is populated.

.pipelines/templates/build-pip-wheels-windows.yml
wavefunction91: Removed from both `build-pip-wheels-windows.yml` and `test-pip-wheels-windows.yml`. The templates now derive it from `arch`, so Windows ARM64 gets `venv` and every other target gets the policy-default `conda`, with the reason stated in one place. Azure Pipelines has no inline conditional expression, so the mapping is a one-line `if` in the PowerShell step rather than a template expression. Switching to conda later is a one-word change.

vcpkg-overlay/ports/libaec/portfile.cmake
wavefunction91: Done, and the overlay is deleted rather than kept as a fallback. The vcpkg baseline in `vcpkg-configuration.json` now points at the `2026.07.29` release (`9e593bb18ea69cc5095e012465dcd675a822ed0d`), which contains microsoft/vcpkg#53051 -- the commit that moved libaec off the rate-limited gitlab.dkrz.de host and onto the GitHub mirror. That was the only reason the overlay existed.

Worth flagging: this moves every port version, not just libaec, and invalidates the vcpkg caches on all platforms. CI on this PR is the test of that.

.github/workflows/build-and-test.yaml
```diff
@@ -840,6 +830,24 @@ jobs:
       with:
         # CPython ships win-arm64 builds from 3.11 onwards.
         python-version: '3.13'
+        # Must be explicit. The action defaults to the runner process
+        # architecture, which is x64 under emulation on this image, and that
+        # image's tool cache holds an x64 interpreter it will happily select.
+        # The failure is silent until the very end: the extension module is
+        # built ARM64 by the toolchain configured above, pip resolves win_amd64
+        # dependencies, and the import fails with "%1 is not a valid Win32
+        # application" after the whole build has already run.
```
lorisercole: Remove

.github/workflows/build-and-test.yaml
```diff
@@ -805,19 +795,19 @@ jobs:
...
+      # Note it reads the environment as well as the cache variable, so leaving
+      # this unset is what lets both the C++ and the Python builds auto-detect.
       # Bump to force a rebuild of the cached dependencies.
       DEPS_CACHE_VERSION: v2
-      # The ARM64 runner image ships CMake 4.x, which errors out on projects
-      # declaring cmake_minimum_required below 3.5. Several vendored
-      # dependencies still do; this is the documented escape hatch.
+      # CMake 4.x errors out on projects declaring cmake_minimum_required below
+      # 3.5. Several vendored dependencies still do; this is the documented
+      # escape hatch, and it is a no-op on an image that ships CMake 3.x.
```
lorisercole: Remove

```suggestion
```

.github/workflows/build-and-test.yaml
```diff
@@ -782,20 +781,11 @@ jobs:
...
-    # clang-cl did the same work in 138. That margin was too thin to survive a
-    # slow runner, so use the 360 minute ceiling GitHub allows for a hosted job.
-    #
-    # The cache-save steps deliberately run immediately after the dependency
-    # install rather than at the end of the job, so even a run that later exceeds
-    # this limit still banks the expensive part instead of discarding it.
-    timeout-minutes: 360
+    # Same 1ES arrangement as the x86_64 job, on the ARM64 pool. The JobId is
+    # distinct from that job's so the two cannot collide: both matrices index
+    # from zero within the same workflow run.
```
lorisercole: Remove

```suggestion
```

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +779,285 @@ jobs:
...
+        & "${{ github.workspace }}\.pipelines\install-scripts\setup-msvc-env.ps1" `
+            -Compiler "${{ matrix.compiler }}" -Arch arm64
+        if ($LASTEXITCODE -ne 0) { throw "setup-msvc-env.ps1 failed ($LASTEXITCODE)" }
+        "toolset=$env:MSVC_TOOLSET" >> $env:GITHUB_OUTPUT
+        "cxx-path=$env:CXX_PATH"    >> $env:GITHUB_OUTPUT
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        # CPython ships win-arm64 builds from 3.11 onwards.
```
lorisercole:
```suggestion
```

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +779,285 @@ jobs:
...
+        & "${{ github.workspace }}\.pipelines\install-scripts\install-cpp-deps-windows.ps1" `
+            -SrcDir        "${{ github.workspace }}" `
+            -ClPath        "${{ steps.env.outputs.cxx-path }}" `
+            -BuildType     RelWithDebInfo `
+            -Triplet       "${{ env.VCPKG_TRIPLET }}" `
+            -DepsInstallDir "${{ github.workspace }}\deps-install-${{ matrix.compiler }}"
+        if ($LASTEXITCODE -ne 0) { throw "Dependency install failed ($LASTEXITCODE)" }
+
+    # Saved immediately after the install so a later timeout still leaves a warm
+    # cache for the next run — dependency builds dominate the job's runtime.
```
lorisercole:
```suggestion
```

.github/workflows/build-and-test.yaml
```diff
@@ -759,6 +779,285 @@ jobs:
...
+      run: |
+        cmake --build "cpp/build-${{ matrix.compiler }}"
+        if ($LASTEXITCODE -ne 0) { throw "CMake build failed ($LASTEXITCODE)" }
+
+    # Guards the OpenBLAS ARM64 build: confirms the library reports an ARMV8
+    # kernel set rather than the generic one, and that dgetrf returns above the
+    # getf2 crossover instead of looping forever. Both failure modes are silent
+    # -- the library builds, links, and computes dgemm correctly -- so without
+    # this check a regression surfaces only as the whole test suite timing out.
+    # Costs about a minute.
```
lorisercole: Make comment short: it's an OpenBLAS smoke test for ARM. Details are in the code

.pipelines/pip-scripts/build-pip-wheels-windows.ps1
```diff
@@ -16,44 +17,70 @@ param(
...
+    # among compiler processes. Without that reserve a 7 GB machine looks like it
+    # can run two jobs, and on Windows ARM64 the pair of cl.exe processes then
+    # exhaust the commit limit while instantiating libint2's engine.impl.h:
+    #
+    #   libint2\./engine.impl.h(643): fatal error C1060: compiler is out of heap space
+    #
+    # This leaves larger agents unaffected: a 16 GB machine still gets 4 jobs.
+    $usableGB = $ramGB - 2
+    $jobs  = [math]::Min($cpu, [math]::Max(1, [math]::Floor($usableGB / 3.5)))
+    Write-Host "CPUs=$cpu  RAM=${ramGB} GB (usable ${usableGB} GB)  -> CMAKE_BUILD_PARALLEL_LEVEL=$jobs"
```
lorisercole: Revert please, keep it simple for the pipeline.

lorisercole (changes requested):
- Please revert 5f1f2fa4337eed69f1a868ca24fc451d4513c0e8. Python caches are continuously reused by other runs, we should keep them. 
  ARM64 Windows build caches were not "evicted", they had never been saved in the first place due to missing gzip/tar commands on the runner image, as you can see from this warning: https://github.com/microsoft/qdk-chemistry/actions/runs/31764695122/job/94658003816#step:11:63 . I've just updated the images, so now caching should work properly and dependencies will be reused.
- Please ensure that caching is enabled for the `actions/setup-python` steps of all the jobs of the pipeline. Use these explicit lines:
  ```yaml
          cache: pip
          cache-dependency-path: python/pyproject.toml
  ```
- update all the `actions/cache/restore` and `actions/cache/save` actions to version `v6` throughout the pipeline
- Please make all comments shorter and on point.

.pipelines/python-wheels.yaml
```diff
@@ -129,6 +134,24 @@ extends:
           parameters:
             march: x86-64-v3
 
+      # Same rationale for Windows ARM64, on the 1ES ARM64 pool.
+      # hostArchitecture is required: the 1ES template otherwise defaults it to
+      # amd64, which provisions an x64-emulated agent on the ARM64 image.
+      - job: Build_windows_aarch64_deps
+        pool:
+          name: aqe-1es-Dpsd_v5-arm-pool
```
lorisercole:
```suggestion
          name: aqe-1es-Dpds_v6-arm-pool
```
I had to change the pool SKU due to an incompatibility with the Win arm image.

.pipelines/python-wheels.yaml
```diff
@@ -138,6 +161,14 @@ extends:
               - ImageOverride -equals ${{ target.imageName }}
               os: ${{ target.os }}
               hostArchitecture: Arm64
+          ${{ elseif and(eq(target.os, 'windows'), eq(target.arch, 'aarch64')) }}:
+            dependsOn: Build_windows_aarch64_deps
+            pool:
+              name: aqe-1es-Dpsd_v5-arm-pool
```
lorisercole:
```suggestion
              name: aqe-1es-Dpds_v6-arm-pool
```

lorisercole (dismissed):
Thanks for the update and for testing this in ADO.

Only these comments haven't been addressed yet: https://github.com/microsoft/qdk-chemistry/pull/593#pullrequestreview-4950030211
> * Please revert [5f1f2fa](https://github.com/microsoft/qdk-chemistry/commit/5f1f2fa4337eed69f1a868ca24fc451d4513c0e8). Python caches are continuously reused by other runs, we should keep them.
>   ARM64 Windows build caches were not "evicted", they had never been saved in the first place due to missing gzip/tar commands on the runner image, as you can see from this warning: https://github.com/microsoft/qdk-chemistry/actions/runs/31764695122/job/94658003816#step:11:63 . I've just updated the images, so now caching should work properly and dependencies will be reused.
> * Please ensure that caching is enabled for the `actions/setup-python` steps of all the jobs of the pipeline. Use these explicit lines:
>   ```yaml
>           cache: pip
>           cache-dependency-path: python/pyproject.toml
>   ```
> * update all the `actions/cache/restore` and `actions/cache/save` actions to version `v6` throughout the pipeline
> * Please make all comments shorter and on point.

wavefunction91:
Resolved the remaining items from review 5049359917 / 4950030211 in faf40c29: restored setup-python pip caching with \\cache: pip\\ and \\cache-dependency-path: python/pyproject.toml\\ across the GitHub workflows, kept all cache restore/save actions pinned to the workflow's v6.1.0 SHA, and shortened the ARM64 workflow comments.

.github/workflows/build-and-test.yaml
```diff
@@ -732,6 +756,254 @@ jobs:
...
+        key: windows-vcpkg-${{ env.VCPKG_TRIPLET }}-vs${{ steps.env.outputs.toolset }}-${{ hashFiles('vcpkg.json', 'vcpkg-configuration.json', 'vcpkg-overlay/**') }}
+
+    - name: Restore deps-install cache
+      id: cache-deps
+      if: ${{ !inputs.rebuild_dep_cache }}
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: deps-install-${{ matrix.compiler }}
+        key: >-
+          windows-deps-${{ matrix.compiler }}-${{ env.VCPKG_TRIPLET }}-vs${{ steps.env.outputs.toolset }}-${{ env.DEPS_CACHE_VERSION }}-${{ steps.deps-fp.outputs.fingerprint }}-${{ hashFiles('cpp/manifest/qdk-chemistry/cgmanifest.json', 'external/macis/manifest/cgmanifest.json', 'cpp/cmake/patches/**', '.pipelines/toolchains/windows.cmake') }}
```
RushiGong: Should this deps cache key also include the vcpkg inputs? These deps are built against `vcpkg_installed`, but this key doesn't include `vcpkg.json`, `vcpkg-configuration.json`, or `vcpkg-overlay/**`. If vcpkg changes, this cache can still restore old libint2/ecpint/gauxc next to a new vcpkg tree.

.pipelines/windows/windows-build-msvc-cmake.ps1
```diff
@@ -266,9 +283,13 @@ Write-Host ""
 if (-not $SkipCpp) {
     if (-not $SkipConfigure) {
         Write-Host "=== Step 1: Configure C++ build ===" -ForegroundColor Yellow
+        # QDK_UARCH is omitted entirely (rather than passed empty) when unset above:
+        # cmake -DQDK_UARCH="" still counts as user-defined and bypasses auto-detection
+        # in qdk-uarch.cmake, producing an empty and invalid /arch: flag.
+        $uarchArgs = if ($QDK_UARCH) { @("-DQDK_UARCH=$QDK_UARCH") } else { @() }
```
RushiGong: Looks like this may become a string instead of an array when `$QDK_UARCH` is set.

wavefunction91: Real bug, thanks for catching it. `$x = if ($cond) { @(single-item) } else { @() }` unwraps the single-item array to a plain string on assignment, and splatting a string with `@` iterates its characters as separate positional args -- confirmed locally (both in `pwsh` and Windows PowerShell 5.1). Fixed in 6e8b6187 by wrapping the whole `if`/`else` in `@(...)`, which forces array context in both branches regardless of item count.

.github/workflows/build-and-test.yaml
wavefunction91: Real gap, thanks. This key hashed the FetchContent-built deps (libint2/ecpint/gauxc via cgmanifest.json) but not the vcpkg inputs the two caches share a build against. Fixed in 6e8b6187: added `vcpkg.json`, `vcpkg-configuration.json` and `vcpkg-overlay/**` to the deps-install cache key, matching what the vcpkg cache key already hashes.

.pipelines/install-scripts/install-cpp-deps-windows.ps1
```diff
@@ -32,18 +35,83 @@ param(
...
 $buildDir = "$SrcDir\deps-build-msvc"
+
+# ─── Resumability ────────────────────────────────────────────────────────────
+# Each dependency records a stamp naming the exact version that was installed,
+# so a restored cache can be inspected and only the missing work redone.
+$stampDir = Join-Path $DepsInstallDir '.deps-stamps'
+
+function Set-DepStamp([string]$Name, [string]$Id) {
+    New-Item -ItemType Directory -Force -Path $stampDir | Out-Null
+    Set-Content -Path (Join-Path $stampDir "$Name.stamp") -Value $Id -NoNewline
```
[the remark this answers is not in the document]
wavefunction91: `Set-Content -NoNewline` has been supported since Windows PowerShell 5.0, well before 5.1. Confirmed directly on this machine:

```
PS> powershell.exe -Command "$PSVersionTable.PSVersion; (Get-Command Set-Content).Parameters.Keys -contains 'NoNewline'"
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      26100  8875
True
```

It also works in `pwsh` (which is what actually runs these steps in CI -- every `shell:` in this workflow is `pwsh`). Resolving.

docs/source/user/quickstart.rst
```diff
@@ -11,14 +11,35 @@ Installation
 
 Install from PyPI with all optional dependencies:
 
-.. code-block:: bash
+.. tab:: Linux / macOS
```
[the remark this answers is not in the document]
wavefunction91: `.. tab::` is the correct, standalone directive for the `sphinx_inline_tabs` extension this repo uses (see `docs/source/conf.py`'s `extensions` list) -- unlike `sphinx-tabs`/`sphinx-design`, it does not require a parent `.. tabs::`/`.. tab-set::` container; that's the extension's whole point (one directive per tab, no wrapper). This has already built cleanly in CI. Resolving.

python/tests/test_deprecated_aliases.py
```diff
@@ -25,6 +25,7 @@
 from qdk_chemistry.algorithms import ExpectationEstimator, QdkExpectationEstimator, state_preparation
 from qdk_chemistry.algorithms.state_preparation import SparseIsometryStatePreparation
 from qdk_chemistry.data import AlgorithmRef, Circuit, QubitHamiltonian, QubitOperator, Settings
+from qdk_chemistry.plugins.qiskit import QDK_CHEMISTRY_HAS_QISKIT
```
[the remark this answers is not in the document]
wavefunction91: This module never imports Qiskit at module scope -- `qdk_chemistry.plugins.qiskit.__init__` only calls `importlib.util.find_spec("qiskit")` to set the flag; the actual `import qiskit` only happens inside `load()`, called elsewhere. Importing `QDK_CHEMISTRY_HAS_QISKIT` here at collection time is safe, and matches the exact pattern already used by ~15 other test files in this suite (`test_circuit.py`, `test_circuit_mapper.py`, `test_state_preparation.py`, etc.). Resolving to keep this file consistent with the rest of the suite.

python/pyproject.toml
```diff
@@ -98,11 +101,15 @@ plugins = [
...
-  "qiskit-qasm3-import",
+  # The qiskit stack and pennylane both require rustworkx, which publishes no
+  # win-arm64 wheels and whose sdist needs a Rust toolchain, so the whole stack
+  # is unavailable on Windows ARM64. QDK_CHEMISTRY_HAS_QISKIT* guard its use.
+  "qiskit[qasm3-import]>=2.2.0; sys_platform != 'win32' or platform_machine != 'ARM64'",
+  # cp314 manylinux aarch64 wheel not available
+  "qiskit-aer; (python_version < '3.14' or platform_machine != 'aarch64' or sys_platform != 'linux') and (sys_platform != 'win32' or platform_machine != 'ARM64')",
+  "qiskit-ibm-runtime>=0.46.1; sys_platform != 'win32' or platform_machine != 'ARM64'",
+  "qiskit-nature; sys_platform != 'win32' or platform_machine != 'ARM64'",
+  "qiskit-qasm3-import; sys_platform != 'win32' or platform_machine != 'ARM64'",
```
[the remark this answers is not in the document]
wavefunction91: Already addressed on this exact marker in an earlier round (see the discussion on `python/pyproject.toml` further up this PR): `platform.machine()` on CPython for Windows returns `PROCESSOR_ARCHITEW6432`/`PROCESSOR_ARCHITECTURE`, which the OS defines in upper case only (`AMD64`, `ARM64`). The lower-case `arm64`/`aarch64` spellings come from macOS/Linux, and those are already excluded by the `sys_platform != 'win32'` half of the marker. PEP 508 has no case-insensitive comparison or boolean `not`, so this is written in De Morgan form intentionally. Empirically verified: the ARM64 wheel run installs cleanly with this marker (no `rustworkx` pulled in, which publishes no win-arm64 wheel and would fail outright if it were). Resolving.

.pipelines/checks/openblas-smoke.c
```diff
@@ -0,0 +1,101 @@
...
+ *      zero-valued blocking parameters produce.
+ *
+ * The Fortran entry points are declared here directly rather than through a
+ * vendor header: no header is needed for the f77 interface, and this is the same
+ * interface blaspp itself calls.
+ */
+#include <stdio.h>
+
+extern char *openblas_get_config(void);
+extern char *openblas_get_corename(void);
```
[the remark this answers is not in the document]
wavefunction91: Good catch, applied in 6e8b6187: both now declared `const char *`, matching how the result is used (read-only, printed only, never written).

python/src/pybind11/data/path_utils.hpp
```diff
@@ -24,28 +24,27 @@ namespace qdk::chemistry::python::utils {
...
+  if (py::hasattr(path_obj, "__fspath__")) {
+    py::object fspath_result = path_obj.attr("__fspath__")();
+    if (py::isinstance<py::str>(fspath_result) ||
+        py::isinstance<py::bytes>(fspath_result)) {
+      return fspath_result.cast<std::string>();
     }
   }
+
+  throw std::runtime_error(
+      "Path argument must be a string or pathlib Path object");
```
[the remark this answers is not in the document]
wavefunction91: The `str(path_obj)` fallback was removed deliberately (see the comment above this function): the exception-driven form it enabled -- catching `py::cast_error` for every non-str/bytes/PathLike argument -- crashed with an access violation under clang-cl on Windows ARM64, which is what this PR is adding coverage for. Reintroducing a `py::str(path_obj)` last-resort would reintroduce that code path for exactly the inputs it doesn't already handle (custom `__str__`-only types), bringing the crash back for them.

If there are known callers passing string-convertible-but-not-PathLike objects, flagging here rather than silently leaving it -- happy to add an explicit, non-exception-driven check for that case if so (e.g. a type predicate before calling `str()`), but wanted to confirm that's an intended API surface first rather than assume.

agamshayit-ms (dismissed):

.github/workflows/build-and-test.yaml
```diff
@@ -592,22 +595,51 @@ jobs:
...
       with:
         path: deps-install-${{ matrix.compiler }}
+        # github.workspace is part of the key, not just the fingerprint: libint2 (and
+        # other FetchContent deps built by install-cpp-deps-windows.ps1) bake their
+        # install prefix as an absolute path into their exported CMake package config,
+        # since they are not relocatable installs. Runners in the same self-hosted pool
+        # can map the workspace to different drives (observed: C:\a\_work vs N:\_work),
+        # so a cache saved on one and restored on the other leaves
+        # Libint2Config.cmake pointing at a path that does not exist on this machine,
+        # failing CMake configure with "includes non-existent path".
```
lorisercole: I'm not sure about the reasoning here. It makes me think that something is not as expected.
For Windows x86_64 we never needed to track the workspace directory. We should not need to. You should be caching the LibInt2 library that cmake installs, not the build folder.
I don't think that the cmake config files are part of the installation, are they?

(I kind of remember seeing this problem when I worked on the first Windows build, and I think it happened because I was trying to cache the build directory.)

.pipelines/python-wheels.yaml
```diff
@@ -129,6 +134,24 @@ extends:
           parameters:
             march: x86-64-v3
 
+      # Same rationale for Windows ARM64, on the 1ES ARM64 pool.
+      # hostArchitecture is required: the 1ES template otherwise defaults it to
+      # amd64, which provisions an x64-emulated agent on the ARM64 image.
```
lorisercole:
```suggestion
```

python/pyproject.toml
```diff
@@ -57,9 +60,13 @@ coverage = [
   "gcovr"
 ]
 discovery = [
-  "azure-ai-discovery==1.0.0",
-  "azure-identity>=1.19.0,<2",
-  "azure-storage-blob>=12.25.0,<13"
+  # azure-identity (via msal) and azure-storage-blob both pull in cryptography,
+  # which publishes no win-arm64 wheel and needs a Rust toolchain plus OpenSSL
+  # dev headers to build from source -- neither is provisioned on the ARM64
+  # runners. Skip discovery there, matching the qiskit-extras/plugins pattern.
```
lorisercole:
```suggestion
  # azure-identity (via msal) and azure-storage-blob both pull in cryptography, which publishes no win-arm64 wheel
```

python/pyproject.toml
```diff
@@ -98,11 +105,15 @@ plugins = [
...
 ]
 qiskit-extras = [
-  "qiskit[qasm3-import]>=2.2.0",
-  "qiskit-aer; python_version < '3.14' or platform_machine != 'aarch64' or sys_platform != 'linux'",  # cp314 manylinux aarch64 wheel not available
-  "qiskit-ibm-runtime>=0.46.1",
-  "qiskit-nature",
-  "qiskit-qasm3-import",
+  # The qiskit stack and pennylane both require rustworkx, which publishes no
+  # win-arm64 wheels and whose sdist needs a Rust toolchain, so the whole stack
+  # is unavailable on Windows ARM64. QDK_CHEMISTRY_HAS_QISKIT* guard its use.
```
lorisercole:
```suggestion
  # The qiskit stack and pennylane both require rustworkx, which publishes no win-arm64 wheels
```

python/pyproject.toml
```diff
@@ -113,8 +124,11 @@ test = [
   "qdk-chemistry[coverage,discovery,jupyter,networkx-extras,openfermion-extras,plugins,qiskit-extras,qre]",
   "nbclient",
   "nbformat",
-  "pennylane",
-  "rdkit",
+  # pennylane also requires rustworkx (see qiskit-extras); the sample test skips
+  # when it is absent.
```
lorisercole:
```suggestion
```

.github/workflows/build-and-test.yaml
```diff
@@ -849,8 +857,16 @@ jobs:
...
         key: >-
-          windows-deps-${{ matrix.compiler }}-${{ env.VCPKG_TRIPLET }}-vs${{ steps.env.outputs.toolset }}-${{ env.DEPS_CACHE_VERSION }}-${{ steps.deps-fp.outputs.fingerprint }}-${{ hashFiles('cpp/manifest/qdk-chemistry/cgmanifest.json', 'external/macis/manifest/cgmanifest.json', 'cpp/cmake/patches/**', '.pipelines/toolchains/windows.cmake') }}
+          windows-deps-${{ matrix.compiler }}-${{ env.VCPKG_TRIPLET }}-vs${{ steps.env.outputs.toolset }}-${{ env.DEPS_CACHE_VERSION }}-${{ github.workspace }}-${{ steps.deps-fp.outputs.fingerprint }}-${{ hashFiles('cpp/manifest/qdk-chemistry/cgmanifest.json', 'external/macis/manifest/cgmanifest.json', 'cpp/cmake/patches/**', '.pipelines/toolchains/windows.cmake') }}
```
lorisercole: same comment as above about `github.workspace`

lorisercole (changes requested):

vcpkg-overlay/ports/openblas/cmake-project-include.cmake
```diff
@@ -25,21 +25,47 @@ elseif(DYNAMIC_ARCH)
...
+    # MSVC cannot assemble the GNU-syntax ARM64 kernels, and the only fallback
+    # that does build, TARGET=GENERIC, produces blocking parameters that make
+    # the blocked LU factorisation loop forever (getrf_single.c steps its loops
+    # by GEMM_P / GEMM_Q / GEMM_R; a zero step never terminates). That failure
+    # is silent -- the library builds, links, and computes dgemm correctly, then
+    # hangs inside dgetrf above the getf2 crossover -- so it must not be allowed
+    # to happen by accident. The portfile selects clang-cl for this target.
+    message(FATAL_ERROR
+        "OpenBLAS for Windows ARM64 must be built with clang-cl: MSVC cannot "
+        "assemble the ARM64 kernels, and the GENERIC fallback hangs in getrf.")
```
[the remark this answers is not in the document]
wavefunction91: Good catch, applied in f1345a31 -- reworded to match the general "Windows ARM" wording used by the second guard just below.

.github/workflows/build-and-test.yaml
wavefunction91: Good instinct -- something was indeed not as expected, just not in the caching setup. Checked libint2 2.9.0's own CMakeLists.txt (the v2.9.0 release tarball) directly:

```cmake
target_include_directories(libint2 INTERFACE
    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include> $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/libint2>
    $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${LIBINT2_INSTALL_INCLUDEDIR}>
```

`LIBINT2_INSTALL_INCLUDEDIR` is just `"include"` (relative). The relocatable pattern would be `$<INSTALL_INTERFACE:${LIBINT2_INSTALL_INCLUDEDIR}>`, which CMake rewrites to `${_IMPORT_PREFIX}/include` in the exported `Libint2Targets.cmake` -- resolved fresh, relative to wherever that file is actually loaded from, every time. Instead libint2 concatenates `${CMAKE_INSTALL_PREFIX}` -- the absolute path from *install* time -- directly into the generator expression. That's what ends up baked into the exported target, which is exactly the error we saw: `includes non-existent path "N:/_work/.../deps-install-msvc/include"`.

So this isn't a build-folder-vs-install-folder issue -- we are caching the actual install tree (`deps-install-msvc`, produced by `cmake --install`), and `cmake install` did put the CMake config files there, same as it would for any package. libint2's own exported package just isn't relocatable, for this specific interface include directory. ecpint and gauxc weren't checked as closely but I'd guess they're fine, since it's specifically this line in libint2 that fails to resolve.

Given that, keying the cache on `github.workspace` is the correct fix unless we also want to patch libint2's CMakeLists.txt (doable, but bigger scope -- happy to open that as a follow-up if you'd rather fix it there instead of working around it). Shortened the in-code comment to cite this directly rather than assert it in f1345a31.

.pipelines/python-wheels.yaml
wavefunction91: Applied in f1345a31.

python/pyproject.toml
wavefunction91: Applied in f1345a31.

python/pyproject.toml
wavefunction91: Applied in f1345a31.

python/pyproject.toml
wavefunction91: Applied in f1345a31.

.github/workflows/build-and-test.yaml
wavefunction91: Good instinct -- something was indeed not as expected, just not in the caching setup. Checked libint2 2.9.0's own CMakeLists.txt (the v2.9.0 release tarball) directly:

```cmake
target_include_directories(libint2 INTERFACE
    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include> $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/libint2>
    $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${LIBINT2_INSTALL_INCLUDEDIR}>
```

`LIBINT2_INSTALL_INCLUDEDIR` is just `"include"` (relative). The relocatable pattern would be `$<INSTALL_INTERFACE:${LIBINT2_INSTALL_INCLUDEDIR}>`, which CMake rewrites to `${_IMPORT_PREFIX}/include` in the exported `Libint2Targets.cmake` -- resolved fresh, relative to wherever that file is actually loaded from, every time. Instead libint2 concatenates `${CMAKE_INSTALL_PREFIX}` -- the absolute path from *install* time -- directly into the generator expression. That's what ends up baked into the exported target, which is exactly the error we saw: `includes non-existent path "N:/_work/.../deps-install-msvc/include"`.

So this isn't a build-folder-vs-install-folder issue -- we are caching the actual install tree (`deps-install-msvc`, produced by `cmake --install`), and `cmake install` did put the CMake config files there, same as it would for any package. libint2's own exported package just isn't relocatable, for this specific interface include directory. ecpint and gauxc weren't checked as closely but I'd guess they're fine, since it's specifically this line in libint2 that fails to resolve.

Given that, keying the cache on `github.workspace` is the correct fix unless we also want to patch libint2's CMakeLists.txt (doable, but bigger scope -- happy to open that as a follow-up if you'd rather fix it there instead of working around it). Shortened the in-code comment to cite this directly rather than assert it in f1345a31.

.github/workflows/build-and-test.yaml
lorisercole: ok, yes, that's a bug on libint2 side indeed.
We've never seen this bug before because normally GHA runners use the same work directory. This time the workflow ran in a different region and the runner used a different work directory. So it can happen.
I think this was fixed in more recent version of libint2. I'm working on an update anyway.

lorisercole (approved):
Looks good, thanks!
All tests green, [ADO build](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=177947&view=results) successful.

nabbelbabbel (approved):

PR #595, opened 2026-07-30
Add `EffectiveHamiltonianConstructor` interface

eimrek:
This PR adds the `EffectiveHamiltonianConstructor` interface planned to be used in the near future by the following downfolding algorithms:
- Schrieffer-Wolff PT2;
- DUCC (Double unitary coupled cluster);
- CT-F12 (Canonical Transcorrelation F12)

The current (initial) design:
- Inputs:
  - Ref. wavefunction - used for the reference state, and to define the target space for the downfolding (P);
  - Hamiltonian - full input Hamiltonian containing the target space P and the external space Q.
- Output:
  - Effective Hamiltonian defined only the in the target space P.

**However, there is an issue: if you calculate a multideterminant wavefunction, and want to set a different target active space, you lose the CI information in the wavefunction.**

One solution against this is to have a separate (optional?) `Orbitals` object that defines the target space, as the input.

Ping: @wavefunction91 @agamshayit-ms @mmoerchen 

Edit: Another point of discussion, is how to make diagnostic data available via this interface. For example, the discarded high-order term norm; or for SW-PT2, intruder states, which the user might want to programmatically access to enlarge the active space.


cpp/include/qdk/chemistry/algorithms/effective_hamiltonian.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * @brief Factory for effective-Hamiltonian constructor instances.
+ */
+struct EffectiveHamiltonianConstructorFactory
+    : public AlgorithmFactory<EffectiveHamiltonianConstructor,
+                              EffectiveHamiltonianConstructorFactory> {
+  static std::string algorithm_type_name() {
+    return "effective_hamiltonian_constructor";
+  }
+  static void register_default_instances();
+  static std::string default_algorithm_name() { return ""; }
```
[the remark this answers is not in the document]
eimrek: will be updated after the first implementation has been added.

cpp/include/qdk/chemistry/algorithms/effective_hamiltonian.hpp
```diff
@@ -0,0 +1,73 @@
...
+      std::shared_ptr<data::Hamiltonian> hamiltonian) const override = 0;
+};
+
+/**
+ * @brief Factory for effective-Hamiltonian constructor instances.
+ */
+struct EffectiveHamiltonianConstructorFactory
+    : public AlgorithmFactory<EffectiveHamiltonianConstructor,
+                              EffectiveHamiltonianConstructorFactory> {
+  static std::string algorithm_type_name() {
```
mmoerchen, line 159: Add Docstrings

cpp/include/qdk/chemistry/algorithms/effective_hamiltonian.hpp
```diff
@@ -0,0 +1,73 @@
...
+    return "effective_hamiltonian_constructor";
+  }
+  static void register_default_instances();
```
mmoerchen, line 162: General comment: all functions should have docstrings.

python/src/pybind11/algorithms/effective_hamiltonian.cpp
```diff
@@ -0,0 +1,81 @@
...
+};
+
+void bind_effective_hamiltonian_constructor(py::module &m) {
+  py::class_<EffectiveHamiltonianConstructor,
+             EffectiveHamiltonianConstructorBase, py::smart_holder>
+      constructor(m, "EffectiveHamiltonianConstructor", R"(
+Abstract base class for effective-Hamiltonian construction.
+
+Concrete implementations construct an effective Hamiltonian from a reference
+wavefunction and an input Hamiltonian.
```
mmoerchen: Add a minimal example how to create a custom effective Hamiltonian

python/src/pybind11/algorithms/effective_hamiltonian.cpp
```diff
@@ -0,0 +1,81 @@
...
+)");
+
+  constructor.def(py::init<>());
```
mmoerchen: Add a docstring to all user facing functions and properties

mmoerchen (changes requested):

cpp/include/qdk/chemistry/algorithms/effective_hamiltonian.hpp
```diff
@@ -0,0 +1,73 @@
...
+   * @brief Construct an effective Hamiltonian from a reference wavefunction.
+   *
+   * @param reference Reference wavefunction providing the reference state and
+   *        defining the target active space.
+   * @param hamiltonian Input Hamiltonian to transform.
+   * @return The effective Hamiltonian.
+   */
+  virtual std::shared_ptr<data::Hamiltonian> _run_impl(
+      std::shared_ptr<data::Wavefunction> reference,
+      std::shared_ptr<data::Hamiltonian> hamiltonian) const override = 0;
```
agamshayit-ms: We should also include the P-space indices as a `SymmetryBlockedIndexSet` argument

agamshayit-ms (dismissed):

eimrek:
Thanks @agamshayit-ms!

I also adapted the docs a bit and renamed `p_space_indices` -> `p_indices` (less verbose, the 'space' seems redundant).

python/tests/test_effective_hamiltonian_constructor.py
```diff
@@ -0,0 +1,62 @@
+"""Tests for the effective-Hamiltonian constructor interface."""
```
mmoerchen: Is there a reason we're testing the abstract base class here?
Also, I'd suggest moving the registration test to a dedicated C++ test file.

mmoerchen (changes requested):
Do we want to add `docs/` in this PR, or until we have an actual implementation PR?

python/tests/test_effective_hamiltonian_constructor.py
mmoerchen: @agamshayit-ms

eimrek: good catch. the python tests for the abstract classes seem to be in `test_algorithms.py` and `test_algorithms_registry.py` and currently c++ tests for this are missing. I'll clean this up.

eimrek:
> Do we want to add `docs/` in this PR, or until we have an actual implementation PR?

Fine with this, I'll add some high-level docs.

cpp/tests/test_effective_hamiltonian.cpp
```diff
@@ -0,0 +1,98 @@
...
+  EXPECT_TRUE(EffectiveHamiltonianConstructorFactory::unregister_instance(key));
+
+  available = EffectiveHamiltonianConstructorFactory::available();
+  EXPECT_TRUE(std::find(available.begin(), available.end(), key) ==
+              available.end());
+}
+
+// run() forwards the reference and the (restricted or unrestricted) P-space
+// index set unchanged to _run_impl, and returns its result.
+TEST(EffectiveHamiltonianConstructorTest, RunForwardsArguments) {
```
mmoerchen: Is there a reason we're testing the abstract base class here? I'd expect the relevant behavior to be covered by tests of the concrete implementations instead.

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
```diff
@@ -0,0 +1,47 @@
...
+==================================
+
+The ``EffectiveHamiltonianConstructor`` algorithm defines a common interface
+for downfolding a Hamiltonian from a larger orbital window into a smaller
+target space.
+
+Problem definition
+------------------
+
+Let the orbitals be partitioned as ``W = P union Q``, where ``P`` is the
```
mmoerchen: This should be an actual formula

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
```diff
@@ -0,0 +1,47 @@
+Effective Hamiltonian construction
+==================================
+
```
mmoerchen, line 3: Can we use the same doc layout as the other algorithm docs?

python/tests/test_algorithms.py
```diff
@@ -624,6 +644,30 @@ def test_coupled_cluster_calculator_inheritance(self, test_ansatz):
...
         assert np.isclose(t2_ab[0], 0.005)
 
+    @pytest.mark.parametrize(
+        "p_indices",
+        [
+            pytest.param(spin_index_set(4, [1, 2], [1, 2]), id="restricted"),
+            pytest.param(spin_index_set(4, [0, 2], [1, 3], equivalent=False), id="unrestricted"),
+        ],
+    )
+    def test_effective_hamiltonian_constructor_inheritance(self, p_indices):
```
mmoerchen: Is there a reason we're testing the abstract base class here? I'd expect the relevant behavior to be covered by tests of the concrete implementations instead.

mmoerchen (changes requested):

mmoerchen (approved):

PR #602, opened 2026-08-01
Add DUCC to qdk-chemistry

agamshayit-ms:
Add initial DUCC implementation.
Uses BTAS code to evaluate wick-contracted BCH up to second order


cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <qdk/chemistry/data/settings.hpp>
+#include <string>
+
+namespace qdk::chemistry::algorithms::microsoft {
+
+class DuccSettings : public data::Settings {
+ public:
+  DuccSettings() {
+    set_default("ducc_level", static_cast<std::int64_t>(2),
+                "DUCC approximation: 0 selects A(1), 1 selects A(4), and 2 "
```
mmoerchen, line 20: Could you add a short comment explaining what A(1), A(4), and A(7) represent?

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
```diff
@@ -84,8 +84,52 @@ See :doc:`Settings <settings>` for a general treatment of algorithm settings in
...
+   H_{\mathrm{eff}}^P = \hat P e^{-\sigma} H_W e^{\sigma} \hat P,
+   \qquad
+   \sigma = T_{\mathrm{ext}} - T_{\mathrm{ext}}^\dagger,
+
+where :math:`T_{\mathrm{ext}}` excludes amplitudes whose indices are all in :math:`P`.
+The input Hamiltonian must be Hermitian and span the full orbital window, but it can use any Hamiltonian container that provides four-center integrals.
+The reference wavefunction must contain real coupled-cluster amplitudes built on a single determinant, use the same full-space orbitals as the Hamiltonian, and occupy a contiguous prefix of orbitals in each spin channel.
+The target P-space must be non-empty in each spin channel.
+
+The ``ducc_level`` setting selects the perturbatively consistent A(1), A(4), and A(7) approximations of :cite:`Bauman2022DUCC`.
```
mmoerchen, line 106: Should we really use A(1), A(4), and A(7) as names here? Later we use "Level" instead. Could we pick a single terminology and use it consistently throughout?

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
```diff
@@ -84,8 +84,52 @@ See :doc:`Settings <settings>` for a general treatment of algorithm settings in
...
+   \mathrm{level\ 1\ (A(4)):}\quad
+      H_{\mathrm{eff}}^P &\simeq \hat P \left(H + [H_N,\sigma]
+      + \frac{1}{2} [[F_N,\sigma],\sigma]\right) \hat P, \\
+   \mathrm{level\ 2\ (A(7)):}\quad
+      H_{\mathrm{eff}}^P &\simeq \hat P \left(H + [H_N,\sigma]
+      + \frac{1}{2} [[H_N,\sigma],\sigma]
+      + \frac{1}{6} [[[F_N,\sigma],\sigma],\sigma]\right) \hat P.
+   \end{aligned}
+
+Level 1 is the second-order perturbatively consistent expansion in Eq. (62) of :cite:`Bauman2019DUCC`; level 2 retains the full double commutator and the Fock-only triple commutator required for third-order consistency.
```
mmoerchen, line 124: What about Level 0?

cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include <variant>
+#include <vector>
+
+#include "ducc_equations.inc"
+
+namespace qdk::chemistry::algorithms::microsoft {
+
+namespace {
+
+btas::Tensor<double> _to_btas(const Eigen::MatrixXd& source) {
```
mmoerchen, line 39: Would it make sense to make these tensor conversions generally available?

For example, by templating the conversions and moving the BTAS <-> Eigen conversion helpers to `utils/`.

cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.cpp
```diff
@@ -0,0 +1,321 @@
...
+      source.data(), static_cast<Eigen::Index>(source.extent(0)),
+      static_cast<Eigen::Index>(source.extent(1)));
+}
+
+Eigen::VectorXd to_eigen_vector(const btas::Tensor<double>& source) {
+  return Eigen::Map<const Eigen::VectorXd>(
+      source.data(), static_cast<Eigen::Index>(source.size()));
+}
+
+std::vector<std::uint32_t> _positions_in_window(
```
mmoerchen, line 73: These functions seem generally useful across effective Hamiltonians. Would it make sense to move them into a shared `detail` namespace in the base effective-Hamiltonian code?

cpp/CMakeLists.txt
```diff
@@ -139,6 +139,22 @@ else()
...
+if(NOT BTAS_FOUND)
+    include(FetchContent)
+    FetchContent_Declare(BTAS
+        GIT_REPOSITORY https://github.com/BTAS/btas.git
+        GIT_TAG 287b145ead818a0332f2b7ce0b7375a83d328bae
+    )
+    # BTAS's linalg kit force-writes the BUILD_TESTING it sees back into the cache.
+    set(_qdk_build_testing "${BUILD_TESTING}")
+    set(BUILD_TESTING OFF CACHE BOOL "Build the testing tree." FORCE)
+    FetchContent_MakeAvailable(BTAS)
```
mmoerchen, line 153: Could `BTAS` use the normal dependency gate, or be installed through the Windows dependency script?

The Windows jobs configure with `QDK_ALLOW_DEPENDENCY_FETCH=OFF`, but this direct `FetchContent_MakeAvailable()` appears to bypass that mechanism. As a result, a clean offline/no-fetch configuration still tries to download `BTAS` from GitHub.

cpp/CMakeLists.txt
```diff
@@ -139,6 +139,22 @@ else()
...
+    set(BUILD_TESTING "${_qdk_build_testing}" CACHE BOOL "Build the testing tree." FORCE)
```
mmoerchen, line 154: Could this preserve whether BUILD_TESTING was originally undefined, rather than only preserving its value?

Restoring an empty saved value appears to create a cache entry, which prevents include(CTest) from establishing its usual ON default. As a result, tests can be silently omitted from the build.

cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.cpp
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  const auto& coefficients =
+      input.coefficients()->block({data::axes::alpha(), data::axes::alpha()});
+  const std::optional<Eigen::VectorXd> energy =
+      energies ? std::optional<Eigen::VectorXd>(
+                     energies->block({data::axes::alpha()}))
+               : std::nullopt;
+  return std::make_shared<data::Orbitals>(
+      coefficients, coefficients, energy, energy, overlap,
```
mmoerchen, line 164: Could p_space and inactive be rebuilt using the same spin symmetry as these coefficient blocks?

For restricted inputs, the index sets retain `input.symmetries()`, while the coefficient storage becomes unrestricted. As a result, copying or serializing the resulting `Orbitals` object changes its symmetry metadata and can cause exact subset checks to fail.

cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.cpp
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  // Store the P-space scalar and spin-resolved one- and two-body blocks in the
+  // canonical container consumed by downstream algorithms.
+  auto container =
+      std::make_unique<data::CanonicalFourCenterHamiltonianContainer>(
+          to_eigen_matrix(output_one_a), to_eigen_matrix(output_one_b),
+          to_eigen_vector(output_two_aa), to_eigen_vector(output_two_ab),
+          to_eigen_vector(output_two_bb), std::move(orbitals), output_scalar,
+          Eigen::MatrixXd{}, Eigen::MatrixXd{});
```
mmoerchen, line 240: Could frozen-Q DUCC outputs be made JSON round-trippable?

Occupied Q orbitals are marked inactive, but the canonical JSON loader appears to reject inactive indices when no `inactive_fock_matrix` is present. A frozen-core round-trip test would likely catch this deterministically.

cpp/src/qdk/chemistry/algorithms/microsoft/ducc/ducc.cpp
```diff
@@ -0,0 +1,321 @@
...
+      break;
+    default:
+      throw std::runtime_error("ducc: unsupported ducc_level " +
+                               std::to_string(level));
+  }
+
+  // Store the P-space scalar and spin-resolved one- and two-body blocks in the
+  // canonical container consumed by downstream algorithms.
+  auto container =
+      std::make_unique<data::CanonicalFourCenterHamiltonianContainer>(
```
mmoerchen, line 236: Could the DUCC two-body convention either match the canonical container contract or be represented explicitly?

The current tests seem to rely on a antisymmetrized same-spin representation, but neither the container nor its serialized form appears to encode that convention. Could we also add a test that would fail if a consumer interpreted those integrals as the canonical representation?

python/tests/test_effective_hamiltonian_ducc.py
```diff
@@ -0,0 +1,438 @@
...
+    ci = _ci_unrestricted if unrestricted else _ci_restricted
+    energy_casci = ci(cas_ham, n_active_a, n_active_b)
+
+    # FCI on the DUCC output. Every level emits the spin-blocked container whose
+    # same-spin block is the half-antisymmetrized representative, which a
+    # conventional solver misreads, so it is diagonalized through the native
+    # qubit mapper.
+    energy_fci = _qubit_ground_state(out_ham)
+
+    assert np.isclose(energy_casci, energy_fci, atol=ci_energy_tolerance), (
```
mmoerchen, line 249: Could all new `np.isclose()` calls explicitly specify `rtol`? NumPy's default `rtol=1e-5` can dominate these absolute tolerances, making the effective threshold much larger than intended.

PR #603, opened 2026-08-02
Allow dense_pure_state to accept a dense amplitude vector

wavefunction91:
## Summary

`state_prep/dense_pure_state` currently only accepts a `Wavefunction`. Its `_run_impl` calls `get_configuration_set()` / `get_active_determinants()` for one purpose: to build a dense statevector. Every line after that point (`StatePreparationParams` -> `MakeStatePreparationOp` -> `Circuit`) is already independent of chemistry, and `PreparePureStateD` itself has no notion of orbitals or determinants.

Because `ConfigurationSet.__init__` requires `Orbitals` and raises if they are `None`, a caller who already holds amplitudes has to fabricate orbitals that carry no meaning for their problem in order to reach a routine that never looks at them.

**NEEDED-BY:** preparing an externally computed guiding state — a normalized amplitude vector over a qubit register with no molecular structure — fails with `AttributeError: 'numpy.ndarray' object has no attribute 'get_configuration_set'`. The only workarounds are to fabricate a molecule, or to duplicate the ~25 neutral lines of `_run_impl` outside the library, which then silently drifts from it.

## Change

- `_run_impl` accepts `Wavefunction | Sequence[float] | np.ndarray`.
- Both paths route through a new shared `_circuit_from_statevector` helper, so they are identical by construction rather than by maintenance.
- The `Wavefunction` path is behaviourally untouched.
- Amplitude vectors are normalized; non-power-of-two lengths, zero-norm vectors, `>32` qubits, and genuinely complex amplitudes are rejected with explanatory `ValueError`s. A complex array whose imaginary part is zero is accepted.

## Validation

Run against the released `2.0.0` wheel's compiled `_core` with these Python sources overlaid, because no local C++ toolchain was available:

- `tests/test_state_preparation.py` — **34 passed** (26 pre-existing and unmodified, plus 8 new).
- `tests/test_algorithms.py` — **30 passed**, no regression.
- Equivalence is asserted directly rather than assumed: for the `wavefunction_4e4o` fixture, feeding the statevector that the `Wavefunction` path produces back through the amplitude path yields the same `numQubits`, the same `rowMap`, `stateVector` agreeing to `1e-12` (measured max abs difference `0.0`), and identical QASM.
- `ruff format --check` clean. `ruff check` on the touched files is clean; the one error it reports (`PLR0917` at `sparse_isometry.py:739`) is pre-existing and in a file this PR does not modify.

## Notes for reviewers

- I could not build the C++ extension in my environment, so CI will be the first full-matrix build of this branch.
- `_circuit_from_statevector` is a private `@staticmethod`. If you would prefer it exposed publicly (e.g. as a module-level `circuit_from_statevector`), say so and I will adjust.
- The base `StatePreparation.run` type hint was widened to match, with the docstring noting that accepting amplitudes is implementation-specific rather than required of every implementation.

### DIFF
--- a/python/src/qdk_chemistry/algorithms/state_preparation/dense_pure_state.py
+++ b/python/src/qdk_chemistry/algorithms/state_preparation/dense_pure_state.py
@@ -5,6 +5,8 @@
 # Licensed under the MIT License. See LICENSE.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+from collections.abc import Sequence
+
 import numpy as np
 
 from qdk_chemistry.data import Wavefunction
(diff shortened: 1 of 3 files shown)


wavefunction91:
Withdrawing this — thanks for your time. On reflection the need this addresses is better met downstream in the consuming library rather than by widening the interface here, so I'm closing rather than asking you to carry the change. Apologies for the noise.

PR #608, opened 2026-08-03
Add remote algorithm execution

nabbelbabbel:
Depends on: #636, #633


python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -67,18 +67,26 @@ def run(
...
-        the underlying ``algorithm.run()`` directly.
+        Without cache or remote options this is equivalent to calling the
+        underlying ``algorithm.run()`` directly.
 
         Args:
             *args: Positional arguments for the algorithm.
             cache: Cache backend — a :class:`CacheBackend`, a path
                 (``str`` / ``Path`` → :class:`FolderCache`), or ``None``.
+            local_cache: Cache backend used by the submitting process.
+            remote_cache: Cache backend reachable from the compute node.
```
wavefunction91: This separation is incredibly confusing. What is the difference between the `cache` and `local_cache`? Why do we need a separate `local_cache` and `remote_cache`?

python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -67,18 +67,26 @@ def run(
...
+            remote: Remote backend name or instance, or ``None`` for local execution.
+            poll_interval: Seconds between remote job status checks.
```
wavefunction91: Is there a reason to have algorithm specific polling intervals? I feel like this can be global config and/or part of the base settings instance. I'd prefer to keep type-generic things like integers out of the optional args here, it just makes things confusing.

docs/source/conf.py
```diff
@@ -244,15 +244,17 @@ def autodoc_skip_imports(app, what, name, obj, skip, options):
...
 
-        # Skip re-exported members (e.g., qdk_chemistry.algorithms re-exporting data types)
+        # Index remote objects only in their defining modules to avoid duplicate targets.
         if (
             module
             and target_module
             and module != target_module
             and module.startswith("qdk_chemistry")
-            and target_module.startswith("qdk_chemistry")
+            and target_module.startswith("qdk_chemistry.remote")
```
wavefunction91: This narrows the skip condition from `target_module.startswith("qdk_chemistry")`
to `qdk_chemistry.remote`, and the comment above changed from "Skip re-exported
members (e.g., qdk_chemistry.algorithms re-exporting data types)" to a
remote-specific rationale.

Doesn't that re-enable documentation of re-exported members across the *entire
rest* of the package — the `qdk_chemistry.algorithms` case the original guard
was written for? The `target_module` derivation on line 247 also changed
globally.

If the goal is to fix duplicate targets under `remote`, could that be done
without relaxing the existing behavior everywhere else?

docs/source/conf.py
```diff
@@ -321,11 +323,27 @@ def normalize_autodoc_signature(
 
 def normalize_autodoc_docstring(app, what, name, obj, options, lines):
     """Rewrite internal module references that appear inside docstrings."""
+    if name.startswith("qdk_chemistry.remote"):
```
wavefunction91: Could you double-check what this does to bullet lists? After dedent, a
continuation line and a bullet under a param are both 8-space indented, so the
`line.startswith("        ")` test can't tell them apart. Running this branch
over `proxy.py`'s `RemoteAlgorithmProxy.__init__` docstring gives:

    **remote_config: Backend-specific configuration options (only used if
    remote is a string). Options include: - host: Remote host to connect to
    - python_path: Remote Python interpreter path - timeout: Maximum execution
    time in seconds - remote_workdir: Working directory on remote system
    - Other backend-specific settings

i.e. the option list renders as one run-on sentence rather than a list.

More broadly — this is a shared config file, and the branch is gated on
`name.startswith("qdk_chemistry.remote")`, so docstring handling now behaves
differently depending on which module you're in. There are ~100 wrapped `Args:`
continuations elsewhere in `python/src` (plugins/pyscf/scf_solver.py,
data/qubit_operator.py, plugins/qiskit/conversion.py, ...) that won't get this
treatment. Is a module-scoped special case the right place for it, or should
the fix be either global or in the docstrings themselves?

Related: the `priority=400` change on line 572 reorders this handler for the
whole build, not just for `remote`.

python/src/qdk_chemistry/remote/backends/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+def _load_plugin_backends() -> None:
+    """Auto-discover and register backends from entry points."""
+    try:
+        from importlib.metadata import entry_points  # noqa: PLC0415
+
+        eps = entry_points(group="qdk_chemistry.remote_backends")
+        for ep in eps:
+            backend_cls = ep.load()
+            register_backend(ep.name)(backend_cls)
+    except Exception:  # noqa: BLE001
```
wavefunction91: The `try` wraps the whole loop, so one third-party package with a broken entry
point aborts discovery for every backend after it — and the user sees only
"No backend registered with name 'x'", with nothing pointing at the real cause.

`_import_algorithm_plugins` in `qdk_chemistry/__init__.py:146` already solved
this correctly: try/except *inside* the loop, with a `warnings.warn` naming the
failing entry point. Could this follow that pattern? Two discovery mechanisms
landing in the same PR with opposite failure semantics seems worth reconciling.

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,354 @@
...
+
+@dataclass
+class JobStatus:
+    """Status of a remote job.
+
+    Returned by :meth:`RemoteBackend.check` and related helpers.
+    """
+
+    job_id: str
+    status: str  # "submitted", "running", "succeeded", "failed", "canceled"
```
wavefunction91: This documents the status vocabulary as lowercase, but `submit_and_wait` below
(line 159-162) compares against `("Succeeded", "Failed", "canceled")`, and both
built-in backends emit the capitalized forms. Nothing breaks today because the
two shipped backends happen to agree with the comparisons rather than with this
comment — but a third-party backend written against this line would poll
forever, and a lowercase "succeeded" reaching the cached path in proxy.py is
rejected as a failure.

Given that `@register_backend` and the entry-point group in this PR are aimed
squarely at out-of-tree backends, could the accepted set be normalized in one
place (an enum, or a case-insensitive terminal-status helper on `JobStatus`) so
the published contract and the comparisons can't drift apart?

python/src/qdk_chemistry/remote/backends/ssh.py
```diff
@@ -0,0 +1,305 @@
+"""SSH backend for remote execution of QDK/Chemistry algorithms.
+
+This backend transfers files and executes scripts over SSH, making it
+suitable for remote servers, HPC login nodes, and cloud VMs.
```
wavefunction91: I'd like to push back on shipping this as a core module, and propose it lands
as an example instead — though I think the longer-term question is worth a
proper discussion rather than a yes/no on this PR.

What this file encodes is a deployment policy, not a library capability: launch
under `nohup`, track via a PID file in `/tmp/qdk_remote`, poll with `kill -0`,
scrape `tail -50 stderr.log`, assume `python3`. Every one of those is a site
decision. This line markets it as "suitable for ... HPC login nodes" — but
running a long chemistry job under `nohup` on a login node is the thing cluster
admins ban; a real HPC path goes through SLURM/PBS, which this bypasses.

The part that worries me most is testability. Of the 13 methods on
`SSHBackend`, `test_ssh_backend.py` exercises two — `check` and `cancel` — and
both are mocked at `_run_remote`. Everything that actually invokes ssh/scp
(`connect`, `upload`, `download`, `_ssh_cmd`, `_run_remote`, `_submit`,
`fetch`) has no coverage, and no CI job stands up an SSH target. The tests we do
have assert exact command strings via `run_remote.call_args_list`, so they're
pinned to the current implementation — they'll break on a refactor that changes
nothing observable, while still not telling us whether a real submission works.
That's the profile of a module that silently rots: in-core, so we own it, but
unexercised, so we won't find out it broke until a user does.

As an example, that's fine — an example is *expected* to be illustrative and
version-drifting, and the reader supplies their own environment. As a shipped
backend it's a support commitment we can't currently verify.

The secondary argument is that moving it *earns* the extension point this same
PR introduces. `@register_backend` plus the `qdk_chemistry.remote_backends`
entry-point group currently have zero out-of-tree consumers — `ssh` and `local`
are both hard-imported in `backends/__init__.py:21`. Shipping SSH as an example
turns the plugin mechanism from speculative into demonstrated.

The repo already has the precedent and the slot:
`docs/source/_static/examples/python/custom_plugin.py` opens with "demonstrates
how to extend QDK/Chemistry with custom plugins: 1. Adding a new backend for an
existing algorithm type", and `docs/source/user/comprehensive/plugins.rst` pulls
it in with `literalinclude` + start/end cell markers. A
`custom_remote_backend.py` alongside it would fit that pattern exactly. `local`
stays in-core as the tested reference implementation. The move would touch
`backends/__init__.py:8,21`, `remote/__init__.py:11`,
`base.py:85,286,291,310,331,341`, `proxy.py:46,357,363`, and relocate
`test_ssh_backend.py`.

But I'd rather open this up than dictate it: what's our long-term intent for
remote execution targets? If we do want first-class transports in-core, what
would it take to test them for real — a containerized sshd in CI, a nightly
integration job against a known host, something else? And is SSH-plus-nohup the
transport we'd pick, or is a scheduler-aware backend the one that actually earns
core status? Worth a wider conversation before we commit to owning this surface.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,484 @@
...
+
+        backend = self._resolved_backend
+        try:
+            return backend.submit_and_wait(payload)
+        finally:
+            if self._backend_name is not None:
+                backend.disconnect()
+                self.__dict__.pop("_resolved_backend", None)
+
+    def _build_payload(self, *args: Any, **kwargs: Any) -> dict:
```
wavefunction91: `_build_payload_for` (line 219) looks line-for-line identical to this method
apart from `self._algorithm` vs the `algorithm` parameter. Same for the static
`_poll_until_done` at line 176 vs the module-level one at line 276 — and I
can't find any caller of the static version, or of `run_cached` (line 164).

Could the duplicates be collapsed onto the module-level functions, and the dead
ones dropped? Two copies of the hashing/payload logic will drift the first time
someone changes what goes into `run_hash`.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,484 @@
...
+        """Return string representation of the proxy."""
+        if self._backend_name:
+            backend_name = self._backend_name
+        elif self._backend is not None:
+            backend_name = self._backend.name
+        else:
+            backend_name = "unknown"
+        return f"<RemoteProxy({self._algorithm.name()}) -> {backend_name}>"
+
+    def __del__(self):
```
wavefunction91: Two things here worth a second look:

1. This disconnects a backend the caller owns. In `_run_uncached` a temporary
   proxy is constructed around a caller-supplied `RemoteBackend`; when that
   proxy is collected, this closes the caller's backend out from under them.
   `run()` already disconnects in its `finally` for the name-constructed case,
   which is the only case that owns anything.

2. `import contextlib` is *outside* the `suppress`, so at interpreter shutdown
   this raises `ImportError: sys.meta_path is None` rather than being ignored.

Is the finalizer needed at all given the `finally` in `run()`?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,588 @@
...
+        """Get the DataClass type for a type name, with lazy loading."""
+        if type_name in cls._dataclass_registry:
+            return cls._dataclass_registry[type_name]
+
+        # Lazy load common QDK Chemistry data types
+        cls._lazy_load_dataclasses()
+        return cls._dataclass_registry.get(type_name)
+
+    @classmethod
+    def _lazy_load_dataclasses(cls) -> None:
```
wavefunction91, line 259: This registers 11 types, but `qdk_chemistry.data` exports 70. Missing are most
of the algorithm *return* types — `QpeResult`, `PauliOperator`, `MeasurementData`,
`EnergyExpectationResult`, `StabilityResult`, `NuclearGradients`, `LatticeGraph`.

Could you check what happens to a remote phase-estimation run? As far as I can
tell the inputs serialize fine and the worker runs, and then `fetch()` raises
`TypeError: Unknown DataClass type: qpe_result` — i.e. the failure lands after
the compute has already been paid for.

Is there a reason this is a hand-maintained allowlist rather than a lookup
driven by the `DataClass` type name at deserialization time? An allowlist means
every new data type silently breaks remote execution until someone remembers
to edit this function.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,588 @@
...
+            directory,
+            f"arg_{i}",
+            arg,
+            cache=remote_cache_backend,
+            content_hash=chash,
+        )
+        if chash and "content_hash" not in entry:
+            entry["content_hash"] = chash
+        manifest["args"].append(entry)
+        if entry.get("file"):
```
wavefunction91: `files_created` is only appended when the *top-level* entry has a "file" key.
For a list/tuple/dict argument the entry is `{"type": "list", "items": [...]}`,
so the nested HDF5 files that `serialize_value` just wrote to `directory` are
never added — and `SSHBackend._submit` uploads only what this list returns.

Separately, the settings loop above (line 401) never appends to `files_created`
at all, so any setting that serializes to a file has the same problem.

Could you check `algo.run([structure_a, structure_b])` end-to-end? I think the
manifest ends up referencing files the worker never receives. Would collecting
files by walking the manifest recursively (or having `serialize_value` return
its created paths) be more robust than checking one level?

python/src/qdk_chemistry/remote/backends/ssh.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def check(self, backend_state: dict) -> JobStatus:
+        """Check whether the remote process is still running."""
+        remote_job_dir = backend_state["remote_job_dir"]
+
+        # Read PID
+        pid_path = shlex.quote(f"{remote_job_dir}/pid")
+        pid_result = self._run_remote(f"cat {pid_path}", timeout=10)
+        if pid_result.returncode != 0:
+            return JobStatus(job_id="", status="unknown", error="Could not read PID file")
```
wavefunction91: Both of these return `status="unknown"` (here for an unreadable PID file, and
on line 233 for an unparseable one), but "unknown" is treated as terminal by
neither polling loop — `submit_and_wait` in base.py:160 breaks only on
`("Succeeded", "Failed", "canceled")`, and `Job.is_terminal` doesn't recognize
it either. Neither loop has an overall deadline.

So if the PID file is missing or truncated — a job that failed before the
`echo $! > pid` landed, or a wiped `/tmp` — the client polls every 5s forever
rather than surfacing the `error` string that's being set right here.

Should "unknown" be terminal after some number of consecutive occurrences, or
should the polling loops carry a timeout so no backend can hang the caller
indefinitely?

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,484 @@
...
+    if cache is not None:
+        resolved_cache = resolve_cache(cache)
+        resolved_remote_cache = resolved_cache
+    else:
+        resolved_cache = resolve_cache(local_cache)
+        resolved_remote_cache = resolve_cache(remote_cache)
+
+    # No cache — just run
+    if resolved_cache is None:
+        return _run_uncached(algorithm, remote, args, kwargs)
```
wavefunction91, line 246: Tracing this branch: if a caller passes `remote_cache=` but leaves
`local_cache=None`, then `resolved_cache` is None and we return here — which
throws away `resolved_remote_cache` entirely, so the compute node never sees
the cache and `payload["remote_cache"]` is never set.

That reads as a direct contradiction of the docstring above, which says
`local_cache=None` means "skip local caching" (not "disable remote caching
too"). Could you double-check whether `remote_cache`-only is meant to be a
supported combination? If it is, the guard probably needs to key off
`resolved_cache is None and resolved_remote_cache is None`.

This is also the concrete reason behind my comment on the `run()` signature in
`algorithms/registry.py` — my objection there isn't only that three cache
parameters are hard to explain, it's that one of the three documented
combinations currently does nothing at all. If the local/remote split were
collapsed, this failure mode wouldn't be expressible in the first place.

python/src/qdk_chemistry/algorithms/registry.py
wavefunction91: Following up on my own comment: it's not just awkwardly placed, it's
unreachable on the uncached path. `RemoteBackend.submit_and_wait` at
base.py:159 does `getattr(self, "poll_interval", 5)`, and no backend ever sets
that attribute, so `run(..., remote=backend, poll_interval=0.1)` sleeps 5s
regardless. It only takes effect when a local cache is also configured.

wavefunction91 (changes requested):

python/src/qdk_chemistry/algorithms/registry.py
nabbelbabbel: Good catch it's a remnant from before having the tiered cache idea added, with that implemented one argument should be enough and the internal label of caches being remote accessible or not should let us figure this out internally.

python/src/qdk_chemistry/remote/backends/ssh.py
nabbelbabbel: > "suitable for ... HPC login nodes"

That line is certainly wrong. 
This is meant to be the "ssh-accessible remote machine with python3 and our package in the default environment" simple case only. 
This is decidedly not the solution for queue managed clusters.
Those need their own dedicated plugins, presumably even dedicated to individual setups as we won't be able to predict the extra wrinkles of local security around them either, generally speaking.

I am fine with moving this into an example, it is not much more than that.

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Yes, this needs a larger fix, I think this one may have gotten lost when pulling apart branches. Will update that with a cleaner solution that works for plugins all up.

python/pyproject.toml
```diff
@@ -121,6 +121,13 @@ Repository = "https://github.com/microsoft/qdk-chemistry"
 Source = "https://github.com/microsoft/qdk-chemistry"
 Changelog = "https://github.com/microsoft/qdk-chemistry/releases"
 
+[project.entry-points."qdk_chemistry.plugins"]
+geometric = "qdk_chemistry.plugins.geometric:GeometricPlugin"
+networkx = "qdk_chemistry.plugins.networkx:NetworkXPlugin"
+openfermion = "qdk_chemistry.plugins.openfermion:OpenFermionPlugin"
+pyscf = "qdk_chemistry.plugins.pyscf:PySCFPlugin"
```
wavefunction91: These are optional deps, but routing them through `_load_plugins` means a
missing one warns (`plugins/__init__.py:126`) instead of being skipped. Main
did the opposite — `contextlib.suppress(ImportError)` per plugin.

So a plain `pip install qdk-chemistry` now emits up to five UserWarnings on
import. Related: `load()` sets `_loaded = True` before the failing import, so
the old suppressed block at `__init__.py:122-141` is dead and installing pyscf
later in the session won't retry.

Can the bundled plugins stay off this path?

python/src/qdk_chemistry/data/registry.py
```diff
@@ -0,0 +1,97 @@
...
+    import qdk_chemistry.data  # noqa: F401, PLC0415
+
+    stack = list(_PythonDataClass.__subclasses__()) + list(_CoreDataClass.__subclasses__())
+    seen: set[int] = set()
+    while stack:
+        dataclass_type = stack.pop()
+        if id(dataclass_type) in seen:
+            continue
+        seen.add(id(dataclass_type))
+        if dataclass_type.__dict__.get("_data_type_name"):
```
wavefunction91: `__dict__.get` is declaration-only, but the serializer reads the attribute
normally (`serialization.py:165`, `hasattr` at :121), which follows the MRO:

    Structure(DataClass)         declares "structure"   registered   -> arg_0.structure.h5
    SpecialStructure(Structure)  inherits it            NOT reg'd    -> arg_0.structure.h5
    NewType(DataClass)           declares nothing       NOT reg'd    -> arg_0.None.h5

So an inheriting subclass round-trips as its parent — a silent type downgrade —
and a class with no declared name writes `.None.h5`.

Reachable today: the `data.symmetry` exports declare no name, and the
`SymmetryBlockedTensorRank*` bindings share one C++ name, so they can't all
register. `test_all_public_dataclasses_have_canonical_loader` only walks
`data.__all__`, so it misses them.

Could this derive from the existing `get_data_type_name()` virtual instead of a
hand-mirrored attribute? And should a `None` name raise rather than serialize?

python/src/qdk_chemistry/plugins/__init__.py
```diff
@@ -78,3 +78,59 @@
...
+
+
+def _distribution_name(entry_point: EntryPoint) -> str | None:
+    """Return a normalized distribution name for an entry point, when available."""
+    distribution = getattr(entry_point, "dist", None)
+    name = getattr(distribution, "name", None)
+    return name.lower() if isinstance(name, str) else None
+
+
+def _legacy_entry_point_enabled(entry_point: EntryPoint) -> bool:
```
wavefunction91: Only one of the three gated groups has shipped. At the merge base (`d4d8927c`)
`cache_backends` exists; `algorithm_plugins` and `remote_backends` were both
added earlier in this PR.

For those two there's no released package to stay compatible with, so the
`plugins.rst` line describing them as supported may be promising more than we
mean. Would it be simpler to keep the legacy handling for `cache_backends` and
drop the other two?

python/src/qdk_chemistry/remote/cache/__init__.py
```diff
@@ -106,8 +106,12 @@ def _load_plugin_caches() -> None:
     try:
         from importlib.metadata import entry_points  # noqa: PLC0415
 
+        from qdk_chemistry.plugins import _legacy_entry_point_enabled  # noqa: PLC0415
+
         eps = entry_points(group="qdk_chemistry.cache_backends")
         for ep in eps:
+            if not _legacy_entry_point_enabled(ep):
```
wavefunction91: The surrounding `try` still wraps the whole loop (`except Exception: pass` at
line 117), so one broken cache entry point would skip every one after it, with
no warning. `remote/backends/__init__.py` now handles this per entry point.

The swallow predates this PR, but since this function is being edited here —
worth bringing the two loaders into line while it's open?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,568 @@
...
+            items = []
+            for i, item in enumerate(value):
+                items.append(cls.serialize_value(directory, f"{name}_item_{i}", item))
+            return {"type": "tuple", "items": items}
+
+        # Handle dicts
+        if isinstance(value, dict):
+            entries = {}
+            for k, v in value.items():
+                entries[str(k)] = cls.serialize_value(directory, f"{name}_{k}", v)
```
wavefunction91: `str(k)` isn't injective, and the key also builds the filename:

    {1: "integer", "1": "string"}   ->  one entry '1'; round-trips to {'1': 'string'}
    {"/": blob_a, "_": blob_b}      ->  both become files__.blob.h5; second overwrites first

Both lose data silently. Could the manifest keep the key's type (or reject
non-str keys), and could filenames come from an index or content hash instead of
the key text?

docs/source/conf.py
wavefunction91: Reopening — the branch is unchanged, and I think the bullet-flattening case is
still live. Running it over the current `proxy.py` `__init__` docstring:

    '    **remote_config: Backend-specific configuration options (only used if remote is a string). Options include:'
    ' - host: Remote host to connect to - python_path: ... - Other backend-specific settings'

The blank line before the bullets doesn't protect them; they join onto it,
giving a run-on line with a one-space indent.

Was this closed deliberately because the built page renders fine? If so I'll drop
it — I'd just rather confirm against a build than assume.

python/src/qdk_chemistry/remote/__init__.py
```diff
@@ -18,8 +18,23 @@
...
+    create_remote,
+    get_backend,
+)
 from qdk_chemistry.remote.cache import resolve_cache
+from qdk_chemistry.remote.job import Job
+from qdk_chemistry.remote.proxy import RemoteAlgorithmProxy, run
 
 __all__ = [
+    "Job",
+    "RemoteAlgorithmProxy",
```
wavefunction91: `RemoteAlgorithmProxy` is in `__all__`, but `proxy.py:28` says "Internal proxy
... Not intended for direct use."

One of the two should change. If it's internal, drop it from `__all__` and the
import on line 29; if it's public, fix the docstring.

docs/source/user/comprehensive/plugins.rst
```diff
@@ -190,12 +190,89 @@ Community-developed plugins are also welcome. See :ref:`adding-plugins` for guid
...
-1. Implementing a new backend for an existing algorithm type (e.g., integrating an external quantum chemistry package)
-2. Defining an entirely new algorithm type with its own factory and implementations
+- Implementations of existing algorithm types
+- New algorithm types and their implementations
+- :class:`~qdk_chemistry.data.DataClass` types used in algorithm inputs or outputs
+- Remote execution backends
+- Cache backends
 
-The following sections provide comprehensive examples of each approach.
+The following sections provide complete remote backend and algorithm examples. The same plugin object can register multiple capabilities through :class:`~qdk_chemistry.plugins.PluginRegistrar`.
```
wavefunction91, line 201: Two gaps a reader can't close from the code:

1. Line 197 lists DataClass types as a plugin capability, but nothing says the
   class must declare `_data_type_name` *on itself* — inherited doesn't count,
   and a missing one means it can't be remotely serialized.

2. Nothing covers how to *use* the feature — `run(..., remote=..., cache=...)`,
   what `is_shared` means, what happens on a cache hit. It's docstring-only, and
   `api/python_api.rst` has no `remote` section.

Could both land here?

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,318 @@
...
+not submit work through a queue scheduler such as SLURM or PBS.
+"""
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+################################################################################
+# start-cell-custom-remote-backend
```
wavefunction91, line 16: As an example its job is to teach someone to write their own backend, and I'm not
sure it separates the framework's contract from this example's own choices.

A reader adapting it to SLURM can't easily tell that `connect`/`disconnect`/
`upload`/`download` are abstract while `_submit`/`check`/`cancel` are concrete
and skippable, or that `backend_state` persists through `Job` and so has to stay
JSON-serializable.

Statuses may be worth a look too: `check()` returns `"unknown"` (lines 203, 208),
which isn't in `_TERMINAL_STATUSES` or `is_successful_status`, so that path polls
to the timeout instead of surfacing the `error` it sets.

Could it say which statuses are recognized and what happens to one that isn't,
and mark the nohup/PID/`kill -0` block as this transport's choice rather than the
framework's model?

python/src/qdk_chemistry/plugins/base.py
```diff
@@ -0,0 +1,73 @@
...
+        """Register a cache backend."""
+        from qdk_chemistry.remote.cache import register_cache  # noqa: PLC0415
+
+        register_cache(name)(backend_type)
+
+
+class QdkChemistryPlugin(ABC):
+    """Base class for plugins discovered through ``qdk_chemistry.plugins``."""
+
+    api_version = 1
```
wavefunction91, line 109: Bigger-picture question about this class rather than the line.

`QdkChemistryPlugin` and the `qdk_chemistry.plugins` group are a public extension
contract — once a third-party package pins to it, the base class, group name,
`register()` signature and `api_version` get hard to change. The bundled plugins
don't exercise it meaningfully either: they were direct-imported before and still
are, via `__init__.py:122-141`.

Three things I'd want settled before that's frozen:

- `api_version` defaults to 1 and `plugins/__init__.py:121` compares `!=`, so it
  can't fire today and invalidates every plugin at once when it becomes 2. Should
  it accept a range?
- `_UNIFIED_PLUGIN_DISTRIBUTIONS` is filled during `_load_plugins()` but read at
  module import (`remote/backends/__init__.py:75`), and `register_remote_backend`
  imports `remote.backends` lazily — so the legacy loader can fire mid-iteration
  and run the legacy code it was meant to skip, depending on `entry_points()`
  order.
- It adds import warnings on the default install (`pyproject.toml:128`).

The direction's fine — one group instead of four is a good end state. But I don't
think it should land here. Remote execution doesn't need it, and I'd rather we
didn't settle a public extension contract as a side effect of shipping something
else. My ask is that it comes out of this PR and lands on its own.

wavefunction91: On how separable the two halves are, if useful — it's four call sites:

    remote/backends/__init__.py:59   _legacy_entry_point_enabled
    remote/cache/__init__.py:109     _legacy_entry_point_enabled
    plugins/base.py:49               register_backend
    plugins/base.py:55               register_cache

The dependency runs plugins -> remote: `PluginRegistrar` consumes remote's own
API, which predates the unified group. Remote's only need is
`_legacy_entry_point_enabled`, which exists because of that group — without it
there's nothing to suppress.

None of `test_remote.py`, `test_remote_backend_discovery.py` or
`test_dataclass_registry.py` import `qdk_chemistry.plugins`, so remote keeps its
coverage either way. `data/registry.py` and the `_data_type_name` bindings would
stay — they're needed for serialization and have no plugin dependency.

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,401 @@
...
+
+    job_id: str
+    status: str  # Case-insensitive: submitted, running, succeeded, failed, canceled, retrieved.
+    logs: str = ""
+    error: str | None = None
+    elapsed_seconds: float | None = None
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+    @staticmethod
+    def normalize_status(status: str | None) -> str:
```
wavefunction91, line 84: These three have a summary line but no `Args:`, where most public functions
elsewhere in `python/src` document their parameters.

It matters here because they define the status vocabulary a backend author has to
satisfy — the same question as line 34. What counts as a valid `status`, and what
`None` means, is what an `Args:` line would settle.

Same on the rest of the new public surface: `plugins/base.py:22,28,41,47,53,66`,
`serialization.py:119,307`, `data/registry.py:78`, `worker.py:100,125`,
`job.py:230`. Nothing flags it — D417 only fires when an `Args:` section exists
and omits a parameter.

python/pyproject.toml
nabbelbabbel: Reverted that to avoid warnings.

python/src/qdk_chemistry/plugins/base.py
nabbelbabbel: > It adds import warnings on the default install (pyproject.toml:128).

This has been resolved as per the comment above.

> Should it accept a range?

Can you give an example where a range would be appropriate? Or are you just looking to flag minor changes with a `minor` version? This seems binary on the API contract to me, either we break the API, number needs to change and it should fail/require an update, or it doesn't and the number stays the same.

If we want backwards compatibility, then I think that is something to decide on when/if we move to the next iteration if that is something we can support.

python/src/qdk_chemistry/plugins/__init__.py
lorisercole, line 1: This module doc string looks outdated.

docs/source/conf.py
nabbelbabbel: Yea, all of this just needed to go.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: Updated the example and added explainers. Made _TERMINAL_STATUSES more visible as a source of allowed statuses for users to reference to.

python/src/qdk_chemistry/plugins/__init__.py
nabbelbabbel: Good catch, this slipped through branch mangling on my end, removed legacy stuff that wasn't actually legacy w.r.t. main.

python/src/qdk_chemistry/remote/cache/__init__.py
nabbelbabbel: Removed it here too

docs/source/user/comprehensive/plugins.rst
nabbelbabbel: Updated the text, `_data_type_name` may change based on the other commets.

python/src/qdk_chemistry/algorithms/registry.py
lorisercole, line 1: **`register_instance` silently shadows same-named implementations** (defined in `algorithms/base.py`, reached via `register()` in this file).

`AlgorithmFactory.register_instance` does `self._registry[generator().name()] = generator` unconditionally, so two implementations registering the same `name()` under the same algorithm type overwrite each other — no error, no warning, and `available()` still shows a single entry. This is asymmetric with `register_factory`, which raises `ValueError` on a duplicate algorithm *type*.

With entry-point auto-discovery this is more likely and less visible than before: the winner depends on plugin load order rather than an explicit user import, and can flip after an unrelated install. Suggest at least warning on overwrite:

```python
def register_instance(self, generator):
    name = generator().name()
    if name in self._registry:
        warnings.warn(
            f"Algorithm '{name}' of type '{self.algorithm_type_name()}' is already "
            f"registered; overwriting.", UserWarning, stacklevel=2,
        )
    self._registry[name] = generator
```

A warning (rather than a hard raise) keeps intentional overrides working while surfacing accidental clashes. (Filed at file level since `register_instance` lives in `algorithms/base.py`, which isn't part of this PR's diff.)

python/src/qdk_chemistry/__init__.py
```diff
@@ -116,6 +116,9 @@ def _setup_resources() -> None:
 # Defer plugin imports until after module initialization
 def _import_plugins() -> None:
     """Import pre-packaged plugins after module initialization."""
+    from qdk_chemistry.plugins import _load_plugins  # noqa: PLC0415
+
+    _load_plugins()
```
lorisercole, line 151: **Built-in vs third-party precedence is undefined here.** `_load_plugins()` (third-party entry points) runs *before* the built-in `pyscf`/`qiskit`/`openfermion`/`networkx` `.load()` calls below. Combined with last-write-wins in `AlgorithmFactory.register_instance`, a third-party plugin that deliberately registers under a built-in name — e.g. an alternative SCF integration named `"pyscf"` so existing `create("scf_solver", "pyscf")` keeps working — is **silently overwritten by the built-in**, i.e. the override is discarded (the opposite of what installing an override implies).

The built-ins occupy exactly the short names a replacement would reuse: `"pyscf"`/`"pyscf_stabilized"`, `"qiskit"`, `"openfermion"`, `"qiskit_aer_simulator"`, `"nx_commuting"`. If overriding a built-in is supported, load built-ins first (and warn on overwrite); if not, reserve and document those names. Either way precedence should be explicit rather than an accident of call order.

python/tests/test_algorithms_registry.py
```diff
@@ -23,6 +26,79 @@
     PYSCF_AVAILABLE = False
 
 
+def test_algorithm_plugin_entry_points_load_independently(monkeypatch):
```
lorisercole: **No coverage for duplicate-name registration.** This file tests duplicate *factory type* registration (`test_register_factory_duplicate`) but nothing registers two implementations with the **same `name()` under the same type** — the custom-algorithm tests all use unique names.

Because the current behavior is silent last-write-wins, this gap means a future change to warn/raise on collisions wouldn't be caught either. Suggest adding a test that registers two same-named implementations and asserts the intended contract (warning emitted, or which one wins).

docs/source/user/comprehensive/plugins.rst
```diff
@@ -190,12 +190,89 @@ Community-developed plugins are also welcome. See :ref:`adding-plugins` for guid
...
+   class CustomPlugin(QdkChemistryPlugin):
+       def register(self, registrar: PluginRegistrar) -> None:
+           registrar.register_algorithm(lambda: CustomAlgorithm())
+           registrar.register_dataclass(CustomResult)
+           registrar.register_remote_backend("custom", CustomRemoteBackend)
+           registrar.register_cache_backend("custom", CustomCacheBackend)
+
+When an algorithm uses plugin-defined ``DataClass`` inputs or outputs, register those classes explicitly with ``register_dataclass`` or pass them through the ``data_classes`` argument of ``register_algorithm``. Python return annotations are not used for discovery.
+
+The existing direct registration functions, the :class:`~qdk_chemistry.plugins.base.ChemistryPlugin` compatibility alias, and the ``qdk_chemistry.algorithm_plugins``, ``qdk_chemistry.remote_backends``, and ``qdk_chemistry.cache_backends`` entry-point groups remain supported. New plugin packages should use :class:`~qdk_chemistry.plugins.base.QdkChemistryPlugin` and the unified entry point.
```
lorisercole: This paragraph documents that the legacy groups and the unified entry point coexist, but the docs never state what happens when two plugins register the **same algorithm/backend/cache name**, or how built-in vs third-party names resolve. Since auto-discovery makes the winner load-order dependent, a short "Naming and precedence" note here — implementation names should be unique/namespaced, plus a description of override behavior — would save plugin authors from silent shadowing surprises.

lorisercole:
A few notes on the unified plugin system, all around one theme: **implementation-name collisions are silent**.

`AlgorithmFactory.register_instance` overwrites an existing name without any warning (unlike `register_factory`, which raises on duplicate *types*). With the new entry-point auto-discovery the "winner" now depends on plugin load order rather than an explicit user import, so an accidental clash — or a deliberate override of a built-in like `"pyscf"`/`"qiskit"` — resolves unpredictably and invisibly. See inline comments for: (1) adding an overwrite warning, (2) making built-in vs third-party precedence explicit in `_import_plugins`, (3) a missing duplicate-name test, and (4) a short docs note on naming/precedence.

lorisercole (dismissed):

python/src/qdk_chemistry/remote/__init__.py
nabbelbabbel: Removed from `__all__`.

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Resolved by storing key and values with type info and storing by index

python/src/qdk_chemistry/plugins/__init__.py
nabbelbabbel: Updated the files docs

python/tests/test_algorithms_registry.py
nabbelbabbel: I have added coverage for all 4 types and the factories.

python/src/qdk_chemistry/algorithms/registry.py
nabbelbabbel: You are right to call out the name collisions; an override order is annoying to track though. Especially if it is partially determined by the package managers resolution of the order of packages.

I have updated this to be errors if anything clashes. This should be something that has to be resolved intentionally.
I have also added env var options to disable the loading of out shipped plugins even if their dependencies are present and they would be auto-loaded. This way only the core things we ship are an enforced set, which I think is reasonable.

python/src/qdk_chemistry/__init__.py
nabbelbabbel: See my comment above and the changes in response to one of David's comments.

python/src/qdk_chemistry/remote/worker.py
```diff
@@ -0,0 +1,137 @@
...
+        cache_info = manifest.get("remote_cache")
+        if not cache_info or not cache_info.get("name"):
+            return None, run_hash
+
+        from qdk_chemistry.remote.cache import get_cache  # noqa: PLC0415
+
+        cache_name = cache_info["name"]
+        cache_config = {key: value for key, value in cache_info.items() if key != "name"}
+        return get_cache(cache_name, **cache_config), run_hash
+    except Exception:  # noqa: BLE001
```
lorisercole, line 48: **Silent broad `except` hides real cache/serialization errors.** This `except Exception` (and the ones at lines 59 and 96, in `_get_cached_result` / `_store_cached_result`) intentionally swallow everything so caching never breaks a run — reasonable as a policy, but a genuine failure (a cache write erroring, a malformed manifest) then vanishes with no trace: `_load_remote_cache` silently returns `(None, None)` and remote caching is disabled for the job; `_store_cached_result` silently drops the result.

Suggest a `logging.debug`/`warning` at each swallow point so these are diagnosable. `_load_remote_cache` also collapses two distinct outcomes into `(None, None)` — "no cache configured" vs "cache failed to load" — which further hides failures.

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,401 @@
...
+        name: The backend name (e.g., "custom" or "local").
+
+    Returns:
+        Decorator function that registers the backend class.
+
+    """
+
+    def decorator(cls: Type[RemoteBackend]) -> Type[RemoteBackend]:  # noqa: UP006
+        cls.name = name
+        _BACKENDS[name] = cls
```
lorisercole: **Same silent-overwrite hazard as the algorithm registry, now across two more registries.** `register_backend` does `_BACKENDS[name] = cls` unconditionally, and `register_cache` does the same (`_CACHES[name] = cls`, `remote/cache/__init__.py:40`). As with `AlgorithmFactory.register_instance` (see my earlier comment on `algorithms/registry.py`), two plugins registering the same backend/cache name silently shadow each other — last registration wins, with no error, warning, or trace in `available_backends()` / `available_caches()`. With entry-point auto-discovery the winner depends on load order.

Suggest the same treatment across all three registries: warn on overwrite (or require an explicit override flag). Built-in names like `"local"`, `"folder"`, `"tiered"` are the most likely to be unintentionally reused.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,568 @@
...
+        manifest["results"].append(entry)
+
+    files_created = [
+        directory / filename for entry in manifest["results"] for filename in get_serialized_file_names(entry)
+    ]
+
+    # Write manifest
+    manifest_path = directory / "manifest.json"
+    with open(manifest_path, "w") as f:
+        json.dump(manifest, f, indent=2)
```
lorisercole: **Minor: output manifest isn't written atomically.** `serialize_outputs` writes the result `.h5` files and then this `manifest.json` via a plain `json.dump`. `LocalBackend.check()` treats *exit 0 + `output/manifest.json` exists* as success, so a crash or partial write here can present a truncated manifest as a successful job (and `deserialize_outputs` would then raise mid-`fetch` rather than the job being reported as failed).

The cache layer already solves exactly this with temp-file + `os.replace` (`FolderCache._atomic_write_text`); applying the same pattern here — and to `serialize_inputs` at line 422 — would make the manifest a reliable commit marker.

python/src/qdk_chemistry/data/registry.py
nabbelbabbel: > Could this derive from the existing get_data_type_name() virtual instead of a
hand-mirrored attribute?

Yes, I added that, though it was a bit bigger of a change on the cpp side to make this work based on class rather than instance.

I have extended the test too.

docs/source/user/comprehensive/plugins.rst
nabbelbabbel: I updated the docs accordingly.

python/src/qdk_chemistry/remote/backends/base.py
nabbelbabbel: This has been fixed, using the same logic comment above.

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Fixed

python/src/qdk_chemistry/remote/worker.py
nabbelbabbel: Resolved this instance here, opened #637 for the same fix on cache code on main

python/src/qdk_chemistry/plugins/base.py
nabbelbabbel: See #632

python/src/qdk_chemistry/remote/__init__.py
```diff
@@ -1,15 +1,15 @@
...
 Usage:
     >>> from qdk_chemistry.algorithms import create
+    >>> from qdk_chemistry.remote import create_remote
     >>>
     >>> scf = create("scf_solver")
-    >>>
-    >>> # Local with caching
-    >>> energy, wfn = scf.run(structure, 0, 1, "cc-pvdz", cache="./cache")
+    >>> backend = create_remote("local")
+    >>> energy, wfn = scf.run(structure, 0, 1, "cc-pvdz", remote=backend)
```
mmoerchen: TODO: Would it make sense to show an example with caching and remote?

docs/source/user/comprehensive/plugins.rst
```diff
@@ -233,17 +237,96 @@ Each plugin-defined ``DataClass`` used in algorithm inputs or outputs must decla
...
+The discovered backend is available through the standard remote backend registry:
+
+.. literalinclude:: ../../_static/examples/python/custom_remote_backend.py
+   :language: python
+   :start-after: # start-cell-custom-remote-usage
+   :end-before: # end-cell-custom-remote-usage
+
+Algorithms created through :func:`qdk_chemistry.algorithms.create` accept ``remote`` and ``cache`` keyword arguments on ``run``. :func:`~qdk_chemistry.remote.backends.base.create_remote` returns a connected backend, which the caller disconnects when it is no longer needed:
+
+.. code-block:: python
```
mmoerchen: Could we also make this block a `literalinclude`?

docs/source/user/comprehensive/plugins.rst
```diff
@@ -233,17 +237,96 @@ Each plugin-defined ``DataClass`` used in algorithm inputs or outputs must decla
...
+         cache="./cache",
+      )
+   finally:
+      remote.disconnect()
+
+Passing a path as ``cache`` creates a local :class:`~qdk_chemistry.remote.cache.folder.FolderCache`. On a completed cache hit, ``run`` reconstructs and returns the result without submitting another remote job. If the cache contains an in-flight job for the same algorithm, settings, and inputs, polling resumes instead of creating a duplicate. Pass ``force_rerun=True`` to bypass the lookup and execute again.
+
+By default, a cache is local to the calling machine. Set ``is_shared=True`` only when the same backing store is reachable from both the calling machine and remote compute node, such as a network-mounted directory:
+
+.. code-block:: python
```
mmoerchen: Could we also make this block a `literalinclude`?

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+        if result.returncode != 0:
+            raise ConnectionError(f"SSH connection failed: {result.stderr}")
+
+        subprocess.run(
+            self._ssh_cmd(["mkdir", "-p", self.remote_workdir]),
+            check=True,
+            timeout=30,
+        )
+
+    def disconnect(self) -> None:
```
mmoerchen, line 120: Is there any cleanup we need to do before disconnecting, or is the empty implementation intentional?

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Only ``connect``, ``disconnect``, ``upload``, and ``download`` are abstract
+    in ``RemoteBackend``. The asynchronous job hooks ``_submit``, ``check``,
+    ``cancel``, and ``fetch`` have default implementations that raise
+    ``NotImplementedError``, so a backend can leave unsupported hooks unchanged.
+    This example overrides all four to support the complete asynchronous job
+    lifecycle.
+    """
+
+    def __init__(self, **config):
```
mmoerchen: Could you expand the documentation for the SSH backend? Especially, what would a typical configuration look like? An example would be very helpful.

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+        backend_state = {
+            "remote_job_dir": remote_job_dir,
+            "remote_output_dir": remote_output_dir,
+        }
+        return job_id, backend_state
+
+    def check(self, backend_state: dict) -> JobStatus:
+        """Override the optional async status hook to inspect the SSH worker.
+
+        Status matching is case-insensitive. ``submitted`` and ``running`` are
```
mmoerchen: Is there a reason we're using raw strings here instead of an enum?

python/src/qdk_chemistry/remote/worker.py
```diff
@@ -0,0 +1,143 @@
...
+logger = logging.getLogger(__name__)
+
+_CACHE_MISS = object()
+
+
+def _load_remote_cache(input_dir: Path) -> tuple[Any, str | None]:
+    """Create the cache described by an input manifest, when available."""
+    run_hash = None
+    try:
+        manifest = json.loads((input_dir / "manifest.json").read_text())
```
mmoerchen: Should we check the manifest version and validate it before we use it?

python/src/qdk_chemistry/remote/backends/local.py
```diff
@@ -0,0 +1,273 @@
...
+    if not handle:
+        return ctypes.get_last_error() == error_access_denied  # type: ignore[attr-defined]
+
+    try:
+        exit_code = wintypes.DWORD()
+        if not kernel32.GetExitCodeProcess(handle, ctypes.byref(exit_code)):
+            return False
+        return exit_code.value == still_active
+    finally:
+        kernel32.CloseHandle(handle)
```
mmoerchen: TODO: Shouldn't there be a default return value in this somewhere?

python/src/qdk_chemistry/remote/backends/local.py
```diff
@@ -0,0 +1,273 @@
...
+                self._processes.pop(pid, None)
+                status = "Succeeded" if return_code == 0 and manifest.exists() else "Failed"
+        elif _process_is_running(pid):
+            status = "running"
+        else:
+            status = "Succeeded" if manifest.exists() else "Failed"
+
+        # Read logs if available
+        logs = ""
+        stderr_log = Path(job_workdir) / "stderr.log"
```
mmoerchen: TODO: Should `stderr.log` be hardcoded?

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+        backend.connect()
+        return backend
+
+    def run(self, *args: Any, **kwargs: Any) -> Any:
+        """Execute the algorithm on the remote system.
+
+        This method:
+        1. Serializes all arguments to HDF5
+        2. Uploads to the remote system
+        3. Generates and executes a Python script
```
mmoerchen: What does "generates and executes a python script" mean? Could you explain that a bit more?

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,464 @@
...
+
+        """
+        from qdk_chemistry.remote.job import Job  # noqa: PLC0415
+
+        job_id, backend_state = self._submit(payload)
+
+        job = Job(
+            job_id=job_id,
+            backend=self.name,
+            backend_config=self.config,
```
mmoerchen: Could we validate persisted config, state, and settings before job submission?

Values such as `Path` or `AlgorithmRef` currently appear to fail only during job JSON serialization, after the job has already been launched.

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,464 @@
...
+    def cancel(self, backend_state: dict) -> None:
+        """Cancel a running or queued job.
+
+        Args:
+            backend_state: The opaque state dict produced by ``_submit()``.
+
+        """
+        raise NotImplementedError(f"Backend '{self.name}' does not support cancellation")
+
+    def fetch(self, backend_state: dict, local_dir: str | Path | None = None) -> dict:
```
mmoerchen: Could `fetch()` return `Any` or a generic result type? Algorithms already return scalars, tuples, and DataClass instances, so dict feels overly restrictive.

python/src/qdk_chemistry/remote/backends/local.py
```diff
@@ -0,0 +1,273 @@
...
+        super().__init__(**config)
+        self._workdir: Path | None = None
+        self._processes: dict[int, subprocess.Popen] = {}
+
+    def connect(self) -> None:
+        """Create a temporary working directory."""
+        self._workdir = Path(tempfile.mkdtemp(prefix="qdk_local_"))
+        self.remote_workdir = str(self._workdir)
+
+    def disconnect(self) -> None:
```
mmoerchen, line 209: Could active or persisted jobs own their directories independently of the connection?

Deleting the root directory here can invalidate later `Job.check()`/`fetch()` calls and could also remove files used by a still-running child job.

python/src/qdk_chemistry/remote/backends/local.py
```diff
@@ -0,0 +1,273 @@
...
+        import os  # noqa: PLC0415
+        import signal  # noqa: PLC0415
+
+        process = self._processes.get(backend_state["pid"])
+        if process is not None:
+            process.terminate()
+            return
+
+        with contextlib.suppress(ProcessLookupError):
+            os.kill(backend_state["pid"], signal.SIGTERM)
```
mmoerchen, line 368: Should we verify process identity beyond the PID before signaling it?

PID reuse could cause cancellation to terminate an unrelated process.

python/src/qdk_chemistry/remote/job.py
```diff
@@ -185,12 +188,79 @@ def discover(cls, directory: str | pathlib.Path) -> list[Job]:
...
+        finally:
+            backend.disconnect()
+
+        if (self.status or "").lower() == "retrieved" or self.output_hashes is not None:
+            self.status = "retrieved"
+            job_status.status = "retrieved"
+        else:
+            self.status = job_status.status
+        if self.file_path is not None:
+            self.save()
```
mmoerchen, line 295: Could status updates use an atomic write instead of `Job.save()`?

A crash or concurrent reader could observe partially written JSON and lose the only recovery handle.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+        "args": args,
+        "kwargs": kwargs,
+    }
+
+    with contextlib.suppress(Exception):
+        payload["run_hash"] = algorithm.hash(*args, **kwargs)
+
+    input_hashes: dict[str, str] = {}
+    for i, arg in enumerate(args):
+        input_hashes[f"arg_{i}"] = _item_content_hash(arg)
```
mmoerchen: Could input hashing be best-effort, similar to `run_hash`? `AlgorithmRef` is supported by the serializer but rejected by `_item_content_hash()`, which appears to prevent even uncached remote execution.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+    for key, val in kwargs.items():
+        input_hashes[key] = _item_content_hash(val)
```
mmoerchen: Could positional and keyword hashes use separate namespaces?

A keyword argument named `arg_0` appears to overwrite the positional hash, which could cause the shared cache to load the wrong input.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+    """Hash result items, persist DataClass blobs, update job in cache."""
+    from qdk_chemistry.data._hashing import collect_content_hashes  # noqa: PLC0415
+
+    job.output_hashes = collect_content_hashes(result)
+    job.status = "retrieved"
+
+    items = result if isinstance(result, tuple) else (result,)
+    for entry, item in zip(job.output_hashes, items, strict=False):
+        if "value" not in entry:
+            cache.put_data(entry["hash"], item)
```
mmoerchen, line 80: Could unsupported cache outputs skip caching rather than fail the run?

Remote transport accepts dictionaries, but `FolderCache` rejects them here. That means a run can complete successfully and then fail while writing to the cache.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+    items: list[Any] = []
+    for entry in job.output_hashes:
+        if "value" in entry:
+            items.append(entry["value"])
+        else:
+            data = cache.get_data(entry["hash"])
+            if data is None:
+                return None
+            items.append(data)
+    return items[0] if len(items) == 1 else tuple(items)
```
mmoerchen: Could metadata preserve the result container type and use a dedicated cache-miss sentinel?

Currently (x,) becomes x, () never hits, and cached None values are treated as missing.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+        return _run_uncached(algorithm, remote, args, kwargs)
+
+    payload = _build_payload_for(algorithm, args, kwargs)
+    run_hash = payload.get("run_hash")
+    if run_hash is None:
+        return _run_uncached(algorithm, remote, args, kwargs)
+
+    # 1) Check the cache (skip on force_rerun)
+    if not force_rerun:
+        job = resolved_cache.get_job(run_hash)
```
mmoerchen, line 255: Could the cache acquire an atomic claim before submission? Concurrent callers can both observe a miss and launch the same calculation.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+
+        if job is not None:
+            # 1a) Completed with outputs → reconstruct
+            if job.is_terminal and job.output_hashes:
```
mmoerchen: Could a succeeded job without usable cached outputs retry `fetch()`? An interruption between status persistence and retrieval currently causes duplicate execution.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,413 @@
...
+            job.run_hash = run_hash
+            resolved_cache.put_job(run_hash, job)
+            if _on_job_submitted is not None:
+                _on_job_submitted(job)
+
+            _poll_until_done(job)
+
+            if not job.is_successful:
+                resolved_cache.put_job(run_hash, job)
+                raise RuntimeError(f"Remote job {job.job_id} ended with status: {job.status}")
```
mmoerchen: Could polling retain the final `JobStatus` and include its error and logs here? The current exception discards potentially useful backend diagnostics.

python/src/qdk_chemistry/remote/worker.py
```diff
@@ -0,0 +1,143 @@
...
+    from qdk_chemistry.algorithms import create as create_algorithm  # noqa: PLC0415
+    from qdk_chemistry.remote.serialization import (  # noqa: PLC0415
+        deserialize_inputs,
+        serialize_outputs,
+    )
+
+    input_path = Path(input_dir)
+    output_path = Path(output_dir)
+    cache, run_hash = _load_remote_cache(input_path)
+    result = _get_cached_result(cache, run_hash)
```
mmoerchen: Should `force_rerun` propagate to the worker as well? Otherwise the caller bypasses its cache lookup, but the worker can still return a previous shared-cache result.

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+        ssh_command = ["ssh", *self._ssh_options(), self._ssh_target(), command]
+        return subprocess.run(
+            ssh_command,
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=timeout or self.timeout,
+        )
+
+    def _submit(self, payload: dict) -> tuple[str, dict]:
```
mmoerchen, line 217: Should completed or canceled jobs remove their remote directory? Otherwise inputs, outputs, PID files, and logs seem to accumulate indefinitely.

If retention is intentional, could we document the expected retention policy?

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def cancel(self, backend_state: dict) -> None:
+        """Override the optional async cancellation hook by signaling the worker PID."""
+        remote_job_dir = backend_state["remote_job_dir"]
+        pid_path = shlex.quote(f"{remote_job_dir}/pid")
+        pid_result = self._run_remote(f"cat {pid_path}", timeout=10)
+        if pid_result.returncode == 0:
+            pid = _parse_remote_pid(pid_result.stdout)
+            if pid is not None:
+                self._run_remote(f"kill {pid} 2>/dev/null", timeout=10)
```
mmoerchen, line 357: Could cancellation check the remote command's return code? A failed kill currently appears to succeed, so `Job.cancel()` records the job as canceled while the worker may still be running.

docs/source/_static/examples/python/custom_remote_backend.py
```diff
@@ -0,0 +1,343 @@
...
+            manifest_local = local_dir / "manifest.json"
+            self.download(f"{remote_output_dir}/manifest.json", manifest_local)
+
+            with open(manifest_local) as manifest_file:
+                manifest = json.load(manifest_file)
+
+            for entry in manifest.get("results", []):
+                for filename in get_serialized_file_names(entry):
+                    self.download(
+                        f"{remote_output_dir}/{filename}", local_dir / filename
```
mmoerchen, line 401: Could manifest filenames be validated to stay within `local_dir` before download? Absolute paths and `..` components can otherwise write files outside the target directory before deserialization rejects them.

mmoerchen (dismissed):
Could all changed docstrings document every parameter in the function signature? I found a number of cases where one or more parameters are omitted.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: No. But a remote cleanup was added on fetch.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: I reworked this to be actual arguments instead of the config dict.

python/src/qdk_chemistry/remote/proxy.py
nabbelbabbel: I expanded the explanation a bit

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: Added docs.

python/src/qdk_chemistry/remote/worker.py
nabbelbabbel: Good catch, propagated that.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: This is a small example to show the plugin architecture, hardening it to this degree just makes it harder to read which is against the purpose of being an easy to understand example.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: This is a small example to show the plugin architecture, hardening it to this degree just makes it harder to read which is against the purpose of being an easy to understand example.

docs/source/_static/examples/python/custom_remote_backend.py
nabbelbabbel: Updated that

python/src/qdk_chemistry/remote/proxy.py
nabbelbabbel: Yes, that is a possibility and means that the same run was called twice. At the same time, no claim about atomicity is made about general caches, it can generally not be enforced that caches are able to make/accept atomic manipulations. 

This is not a job scheduler with workers fetching, it is a user or agent triggered single job submission.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,481 @@
...
+            RuntimeError: If remote execution fails.
+            ConnectionError: If connection to remote system fails.
+
+        """
+        # Build the execution payload
+        payload = _build_payload_for(self._algorithm, args, kwargs)
+
+        backend = self._resolved_backend
+        try:
+            return backend.submit_and_wait(payload)
```
wavefunction91: This is the uncached route, which blocks inside `submit_and_wait`. The cached
route below submits through `submit`, builds a `Job`, and polls it with
`_poll_until_done` — the same three steps implemented a second time, including
a second polling loop.

One thing that falls out of the split: `submit_and_wait` accepts a `cleanup`
argument, this call never passes it, and the cached route has no equivalent, so
nothing can release backend artifacts. The uncached route also returns no
`Job`, leaving the caller no handle to cancel or clean up with.

Could both routes go through `submit`, the `Job`, and `fetch`, with blocking
just meaning "poll before returning"? Separately, `RemoteAlgorithmProxy` is
only ever constructed here and run immediately, so its other methods have no
reachable caller.

python/src/qdk_chemistry/algorithms/registry.py
```diff
@@ -89,6 +92,18 @@ def run(
             The algorithm result.
 
         """
+        if remote is not None:
```
wavefunction91, line 100: Before calling this a defect I would like to understand the intended contract.
The `cache=` argument is serving two roles: a memoization store, and, when
`is_shared` is set, a transport channel between the caller and the compute
node. What is a record in it meant to guarantee? In particular, should a result
computed by a plain local run be a hit for a later remote run over the same
inputs, and the reverse?

As far as I can tell that does not currently work, and it fails without any
warning. `_store_result` in this file writes a record with no `output_is_tuple`
field. The remote reader, `_reconstruct_from_cache`, requires that field and
treats its absence as a miss. The record's status is `retrieved`, which counts
as terminal but not as successful, so the resume branch and the re-fetch branch
are both skipped and the call re-submits to the backend. The result of that run
then overwrites the local record. So calling `scf.run(mol, ..., cache="./c")`
and then repeating the call with `remote=` added will recompute a result the
cache already contains.

If the two are meant to interoperate, the three writers (this one, the proxy's,
and the worker's) would need to agree on the record format, and the one that
needs changing is outside this diff. If they are not meant to interoperate,
what should a plain local run do when it is given a cache marked shared? At
present it writes records the remote side cannot read.

python/src/qdk_chemistry/remote/backends/base.py
```diff
@@ -0,0 +1,546 @@
...
+            *True* if the job completed successfully; otherwise, *False*.
+
+        """
+        return self.is_successful_status(self.status)
+
+
+class RemoteBackend(ABC):
+    """Abstract base class for remote execution backends.
+
+    Backends must implement:
```
wavefunction91: I think the enforcement here is inverted. The docstring says a backend has to
implement all nine of these, but Python only holds you to four, and they're the
lifecycle and file-transfer ones. Everything the framework actually reaches for
when it runs a job — submitting, polling, fetching, cancelling, cleanup — is a
concrete method that raises, so a subclass can skip the lot and Python won't
say a word.

Which means you can write a backend that looks finished and isn't. It
instantiates, it registers through the entry point, it turns up in
`available_backends()`, `create_remote` hands it back connected, and then it
dies on `NotImplementedError` in `_submit` the first time someone runs
something on it. The `SlurmBackend` sketch a bit further down this docstring
does exactly that, and so does `PluginRemoteBackend` over in `test_plugins.py`,
both of which are there to show people how to write a backend.

Could `_submit`, `check` and `fetch` become abstract, with `cancel` and
`cleanup_job` explicitly optional?

The other thing I'm unsure about is what `upload` and `download` are for.
Nothing in `python/src` calls them right now (`LocalBackend` serializes straight
into the job directory and reads its outputs back off disk), so I can't tell
from the code what a backend author is meant to do with them. Is the idea that
the base class will eventually drive the whole serialize, transfer, deserialize
sequence through those two? If so, requiring them makes sense and it's that
piece that's still to come. If they're more of a convenience for backend
authors, is requiring them the right call, given that a transport which isn't
file-copy based would just stub them out?

python/src/qdk_chemistry/remote/backends/local.py
```diff
@@ -0,0 +1,416 @@
...
+
+        process = self._processes.get(pid)
+        if process is not None:
+            return_code = process.poll()
+            if return_code is None:
+                status = "running"
+            else:
+                self._processes.pop(pid, None)
+                status = "Succeeded" if return_code == 0 and manifest.exists() else "Failed"
+        elif _process_is_running(pid):
```
wavefunction91: Worth a look at the ordering here. When a `Job` gets resumed in a different
process, `self._processes` is empty, so `check` falls to this branch and asks
whether the PID is alive before it looks at whether the manifest exists. If the
worker has already finished and the OS has recycled its PID onto something
else, this reports "running" for a job whose result is sitting there ready, and
the caller polls until it times out.

`_submit` already records `process_identity` for exactly this reason, and
`cancel` consults it before deciding it's safe to signal. Should `check` use it
too, or look for the manifest first?

Raising it mainly because this is the reference implementation people will
write their own backends against, so whatever `check` does here is likely to
get copied.

wavefunction91 (changes requested):
A few follow-ups after another pass, mostly around the shape of the execution path and the backend contract.

python/src/qdk_chemistry/remote/backends/base.py
wavefunction91: The reader side of this looks handled now. `normalize_status`,
`is_terminal_status`, and `is_successful_status` all casefold, so the
comparisons can no longer drift from the vocabulary.

The writer side hasn't changed, though. `JobStatus.status` is still typed `str`
with the vocabulary carried in a trailing comment, and every producer still
emits a literal: `LocalBackend.check` writes "Succeeded" and "Failed"
capitalized, `Job.cancel` writes "canceled", and both the proxy and the worker
write "retrieved". `JobState` is referenced only by the readers.

For an out-of-tree backend author that leaves three inconsistent signals: the
annotation says `str`, the comment lists lowercase names, and the one in-tree
example emits capitalized ones. Nothing rejects a value outside the set. If
someone writes "complete" or "finished", the job never registers as terminal
and the caller waits out the full default timeout before raising, on a job that
had already finished.

Would typing the field as `JobState` and having the backends emit enum members
be workable? Also worth a look: `CANCELED` and `CANCELLED` are separate
members, so a direct `status == JobState.CANCELED` is False for a backend that
wrote the other spelling.

python/src/qdk_chemistry/remote/backends/local.py
nabbelbabbel: Good catch, updated that.

python/src/qdk_chemistry/algorithms/registry.py
nabbelbabbel: The logic is intended like this: upload and download are the default ways of moving files to and from the remote. For the `local` backend they are thus empty as no movement is needed.

The cache is first and foremost for storage. However, if the cache is shared (both local and remote can access it), then uploaded/download and caching have us do redundant work in moving files across network. Hence, a shared cache should supersede the upload and download.

> In particular, should a result computed by a plain local run be a hit for a later remote run over the same inputs, and the reverse?

Yes, if there is a cache. The cache is something that is first a local thing, it can be local only or shared, there is no remote only. Hence, any request hitting the cache locally and being fully servable from cache will never go to remote.

The issue around the `output_is_tuple` is a good catch I fixed that.

python/src/qdk_chemistry/remote/proxy.py
nabbelbabbel: Updated the tags and the description, removed the `RemoteAlgorithmProxy` that was indeed a left over from merges and back and forth with other iterations of exposing and extending the cpp based algorithms.

python/src/qdk_chemistry/remote/backends/base.py
nabbelbabbel: Made changes accordingly, the explanation for upload/download is in one of the earlier comments.

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -0,0 +1,363 @@
...
+
+    from qdk_chemistry.remote.backends import get_backend  # noqa: PLC0415
+
+    owns_backend = isinstance(remote, str)
+    if owns_backend:
+        backend = get_backend(remote)
+        backend.connect()
+    else:
+        backend = remote
+
```
wavefunction91, line 171: The contract now says `cleanup_job` is optional and leaves it non-abstract, but
this calls `job.cleanup()` unconditionally, as do the two sites in `run`. A
backend that implements the seven required methods and skips the optional one
gets through submit, wait and fetch, then raises `NotImplementedError` here and
loses the result that was just retrieved.

Nothing in the suite covers it because every in-tree backend implements
`cleanup_job`, so this only shows up for out-of-tree backends.

Should cleanup be conditional on the backend supporting it, or is it required
after all, in which case making it abstract alongside the others would say so?

nabbelbabbel: Removed those calls, its the callers job to figure out if they want to clean or not.

wavefunction91 (approved):

PR #619, opened 2026-08-05
Add SW-PT2 downfolding

eimrek:
## Summary

Adds a second-order Schrieffer-Wolff perturbation theory (SW-PT2) implementation for constructing effective Hamiltonians over a reduced orbital space.

The new `qdk_swpt2` implementation of `EffectiveHamiltonianConstructor` folds an external orbital space \(Q\) into a caller-selected target space \(P\), producing a Hermitian zero-, one-, and two-body effective Hamiltonian for subsequent active-space calculations.

## Implementation

The downfold evaluates the second-order Schrieffer Wolff / Van Vleck effective Hamiltonian

$$
H_{\mathrm{eff}} =
P\left(H_{\mathrm{BD}} + \frac{1}{2}[S,H_{\mathrm{OD}}]\right)P,
$$

where the generator \(S\) is formed using semicanonical generalized-Fock energy denominators.

## Settings

- `regularizer_sigma2`: strength of the sigma-squared denominator regularizer; default `1.0`.
- `semicanonicalize`: diagonalize the generalized Fock matrix within each orbital-role block; default `true`.
- `fold_above_two_body`: retain reference 1-RDM pair contractions from generated three-body terms; default `true`.
- `max_folded_occupation_deviation`: maximum allowed deviation of a folded orbital from integer occupation; default `0.5`.

## Known limitation: two-body symmetry

The transformed two-body tensor has only 4-fold symmetry. It preserves Hermiticity and electron exchange,

$$
(pq|rs) = (qp|sr), \qquad (pq|rs) = (rs|pq),
$$

but does not generally preserve the Coulomb bra-swap symmetry

$$
(pq|rs) = (qp|rs).
$$

Consumers must therefore read the complete dense \(n^4\) tensor.

Two existing QDK consumers currently assume 8-fold ERI symmetry and are not compatible with SW-PT2 output:

- `Hamiltonian.to_fcidump_file()`, which writes only the canonical 8-fold-unique entries.
- The native `"qdk"` qubit mapper, whose restricted fast path reads canonical orbital-pair entries.

Using either path currently reconstructs a different operator without raising an error. This should be addressed separately by adding support for general 4-fold restricted tensors.

## Validation

- Focused C++ effective-Hamiltonian and SW-PT2 tests pass, including:
  - an independent determinant-space matrix oracle;
  - identity behavior when \(P = W\);
  - second-order convergence with decreasing coupling;
  - semicanonical rotation covariance;
  - open-shell spin preservation;
  - folding versus two-body truncation;
  - denominator cancellation and guarded-pseudoinverse boundaries;
  - 4-fold tensor symmetry.
- Focused Python API and registry tests pass after rebuilding the extension.
- Strict Sphinx documentation build passes.
- Formatting, linting, type checking, license, stub, and repository pre-commit checks pass.


cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2.cpp
```diff
@@ -0,0 +1,449 @@
...
+namespace sw = qdk::chemistry::algorithms::microsoft::swpt2;
+
+namespace {
+
+/// The reference's spin-traced 1-RDM over its own active space, from the
+/// active 1-RDM when available and from the reference determinant otherwise.
+Eigen::MatrixXd reference_active_density(const data::Wavefunction& reference,
+                                         std::size_t n_ref_active) {
+  const auto size = static_cast<Eigen::Index>(n_ref_active);
+  if (reference.has_active_one_rdm()) {
```
agamshayit-ms: Can we allow references that have `has_one_rdm_spin_traced()` but no `has_active_one_rdm()`?

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2.cpp
```diff
@@ -0,0 +1,449 @@
...
+  // collide with the folded window orbitals above.
+  std::vector<std::size_t> emit_inactive;
+  for (int i : inactive_spatial) emit_inactive.push_back(W_global[i]);
+  emit_inactive.insert(emit_inactive.end(), win_inactive_global.begin(),
+                       win_inactive_global.end());
+  std::sort(emit_inactive.begin(), emit_inactive.end());
+
+  const Eigen::MatrixXd empty_fock = Eigen::MatrixXd::Zero(0, 0);
+  return std::make_shared<data::Hamiltonian>(
+      std::make_unique<data::CanonicalFourCenterHamiltonianContainer>(
```
agamshayit-ms, line 444: Because the qubit mapper currently assumes 8-fold symmetry, feeding this into it will yield wrong energies. Can we at least document this explicitly, until we fix the qubit mapper in the future?

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
```diff
@@ -84,8 +84,190 @@ See :doc:`Settings <settings>` for a general treatment of algorithm settings in
...
+   :math:`(pq|rs) = (qp|sr)` and electron exchange :math:`(pq|rs) = (rs|pq)` survive the
+   transformation, but the bra swap :math:`(pq|rs) = (qp|rs)` of a genuine Coulomb integral
+   does not: the commutator is not an electron-repulsion operator. Consumers must be given
+   the full dense :math:`n^4` block; one that reads only the canonical 8-fold-unique
+   elements silently reconstructs a different operator, with no error raised.
+
+The constructor logs the active regularization, the derived active electron count, the largest
+folded occupation deviation, the minimum denominator and the maximum raw amplitude
+:math:`|c_T/\Delta_T|`. It warns when a folded deviation or the folded core's excess charge is
+large, and when the amplitude exceeds one, where the perturbation series stops contracting.
```
agamshayit-ms: Could `|c_T/Delta_T| > 1` be described as a heuristic warning rather than a
convergence boundary? SW convergence depends on the perturbation norm and
spectral gap, not one matrix-element ratio. Please narrow the same claim in the
Python docstring and implementation comment.

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2_kernel.cpp
```diff
@@ -0,0 +1,1023 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
agamshayit-ms, line 1: Can the blas code here be benchmarked against a computer algebra system such as SeQuant?

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2.cpp
eimrek: sure, thanks 👍

docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst
eimrek: 👍

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2_kernel.cpp
eimrek: The tests contain a independent determinant-space matrix construction for $H_{BD}$, $H_{OD}$ and the generator $S$. It evaluates the commutator directly using ordinary matrix multiplication, without relying on the production projected-Wick contraction machinery.

See https://github.com/microsoft/qdk-chemistry/blob/3a03c134e3a1702e59b1793dc99cfb6f6c7dbc01/cpp/tests/test_swpt2_kernel.cpp#L401

This independent implementation is then matched against the production BLAS wick implementation. See e.g. https://github.com/microsoft/qdk-chemistry/blob/3a03c134e3a1702e59b1793dc99cfb6f6c7dbc01/cpp/tests/test_swpt2_kernel.cpp#L618

I think this already is at a similar level as an independent SeQuant implementation, but happy to discuss further.

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2.cpp
eimrek: Yep, good point. I made this issue: #684, that should be addressed before this PR is merged.

Additionally, there is a warning in the docs about this:

https://github.com/microsoft/qdk-chemistry/blob/3a03c134e3a1702e59b1793dc99cfb6f6c7dbc01/docs/source/user/comprehensive/algorithms/effective_hamiltonian_constructor.rst#L206-L213

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2_kernel.cpp
agamshayit-ms: It will be good to have a confirmation that the commutator algebra implemented in BLAS is correct beyond small systems for which you can form the CI matrix and do the matrix-multiplication test. To do this, you can use a Wick engine of your choice and compare the expressions

eimrek:
Added the Wick-engine-based validation in `cpp/tests/test_swpt2_symbolic.cpp`.

I enumerated every operand channel of ½[S, H_OD] (10,368 of them) with a Wick engine, projected onto the active space, and canonicalized. That yields 664 surviving terms — 132 scalar, 356 one-body, 160 two-body, 16 rank-3. These terms are directly evaluated in the tests against the `downfold_blocked` at https://github.com/microsoft/qdk-chemistry/blob/eef04796c6b5665552853dd89b1f17559e09f64a/cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2_kernel.cpp#L747

@agamshayit-ms let me know if you like this, or did you have something else in mind.

cpp/src/qdk/chemistry/algorithms/microsoft/effective_hamiltonian/swpt2.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+  // The orbitals the window Hamiltonian folded into its core energy are
+  // inactive in the emitted operator too; they lie outside W, so they cannot
+  // collide with the folded window orbitals above.
+  std::vector<std::size_t> emit_inactive;
+  for (int i : inactive_spatial) emit_inactive.push_back(W_global[i]);
+  emit_inactive.insert(emit_inactive.end(), win_inactive_global.begin(),
+                       win_inactive_global.end());
+  std::sort(emit_inactive.begin(), emit_inactive.end());
+
```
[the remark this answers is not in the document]
eimrek: Not needed: `Orbitals` rejects overlapping active/inactive sets and `swpt2.cpp` requires the window core to be exactly the reference core outside W, so the two lists are disjoint by construction (0 duplicates across all 17 downfolds in the test suites); added `RejectsWindowCoreThatIsNotTheReferenceCore` to cover that precondition.

PR #623, opened 2026-08-07
Add fermionic_low_rank BLISS algorithm

BoqinZhang:
fermionic_low_rank BLISS is an algorithm that squeezes the 1-norm of a Hamiltonian without damaging its spectrum of on the subspace of N_e. Cooperating with double factorization (DF), the shifted Hamiltonian can keep the low-rank structure. Reference paper: Patel et al., arXiv:2409.18277

Added `hamiltonian_regularizer` algorithm type
Added `BlissRegularizer` class, and `fermionic_low_rank` implementation
Added public `rebuild_bliss_shifted_hamiltonian` for shifting Hamiltonian instance by set {mu1, mu2, xi}
Added util functions `double_factorize` and `hamiltonian_one_norm` for making the double factorization on a Hamiltonian, and measuring 1-norm (lambda) of a Hamiltonian


cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
```diff
@@ -0,0 +1,297 @@
...
+  BlissShift compute_shift(const data::Hamiltonian& hamiltonian,
+                           unsigned int n_alpha_electrons,
+                           unsigned int n_beta_electrons) const;
+
+  /**
+   * @brief Access the algorithm's name.
+   *
+   * @return The algorithm's name.
+   */
+  std::string name() const override { return "fermionic_low_rank"; }
```
RushiGong: Every other algorithm type in this directory declares an abstract base and puts the concrete implementation under `microsoft/` — `scf.hpp:140`, `mc.hpp:143`, `localization.hpp:153`, `hamiltonian.hpp:78`, `active_space.hpp:172` and the rest are all `virtual std::string name() const = 0`. Here the type and the implementation are the same class, and `name()` hardcodes `"fermionic_low_rank"`.

Could we split it now — abstract `HamiltonianRegularizer` here, `BlissRegularizer` + `BlissSettings` moved down to `microsoft/fermionic_low_rank/`? That'd also make the C++ and Python names agree (the binding already calls it `HamiltonianRegularizer`), and it'd fix the docs in `utils/double_factorization.hpp:44` and `utils/hamiltonian_one_norm.hpp:30,53` that reference a `HamiltonianRegularizer` symbol which doesn't currently exist in C++.

cpp/src/qdk/chemistry/algorithms/hamiltonian_regularizer.cpp
```diff
@@ -0,0 +1,150 @@
...
+  return std::make_unique<BlissRegularizer>();
+}
+
+}  // namespace
+
+void HamiltonianRegularizerFactory::register_default_instances() {
+  QDK_LOG_TRACE_ENTERING();
+
+  HamiltonianRegularizerFactory::register_instance(
+      &make_fermionic_low_rank_regularizer);
```
RushiGong: I think this needs a matching line in `cpp/src/qdk/chemistry/algorithms/algorithm_defaults.cpp` (the `REGISTER_FACTORY_SETTINGS_INIT` block, around lines 39-49) — that file isn't in this PR, so it'd need pulling in.

All 11 existing factories are listed there. Without `REGISTER_FACTORY_SETTINGS_INIT(HamiltonianRegularizerFactory)`, `resolve_algorithm_defaults("hamiltonian_regularizer", ...)` returns `nullptr`, and the caller in `data/settings.cpp` (`AlgorithmRef::_resolve_settings`) wraps it in a `catch (...)` that sets the settings to null. So anyone who later stores an `AlgorithmRef("hamiltonian_regularizer", "fermionic_low_rank")` inside another algorithm's settings gets no defaults and no error message.

The geometry-optimizer PR (`aa7eec75`) added its line there as part of the same change — worth mirroring?

cpp/src/qdk/chemistry/algorithms/microsoft/symmetry_shift/fermionic_low_rank.cpp
```diff
@@ -0,0 +1,212 @@
...
+// Step 2: per-fragment median shift, aggregated into a global (mu2, xi).
+// ---------------------------------------------------------------------------
+
+GlobalTwoBodyShift accumulate_fragment_shifts(
+    const std::vector<qdk::chemistry::utils::TwoBodyFragment>& fragments) {
+  GlobalTwoBodyShift result;
+
+  if (fragments.empty()) {
+    return result;
+  }
```
RushiGong: This early return happens before `result.xi` is sized on line 38, so `xi` stays a default 0×0 matrix.

It then flows into `solve_one_electron_shift`, which hits `coulomb -= norb * xi` in `TwoBodyBlissCorrection::add_coulomb_contraction` with `coulomb` at norb×norb and `xi` at 0×0. Eigen only checks that under `eigen_assert`, and our release toolchains set `NDEBUG`, so I think we get an out-of-bounds read rather than the clean throw in `rebuild_bliss_shifted_hamiltonian` (which never gets reached). It's reachable from settings alone: any `df_truncation_threshold` above the largest supermatrix eigenvalue empties the fragment list. For water/STO-3G that's about 3.38, so `threshold = 10.0` does it. Could we size `xi` before returning?

cpp/tests/test_symmetry_shift.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto ham = hamiltonian_constructor->run(wfn_HF->get_orbitals());
+
+  auto regularizer =
+      HamiltonianRegularizerFactory::create("fermionic_low_rank");
+  auto shifted_ham = regularizer->run(ham, 5, 5);
+
+  auto norm_after =
+      qdk::chemistry::utils::hamiltonian_one_norm(*shifted_ham, 0.0);
+
+  EXPECT_NEAR(norm_after.total, 27.590504297492, 1e-6);
```
RushiGong, line 196: The test is called `ReducesOneNorm` and the doc comment above says the shift "should reduce (or at least not increase) the ... 1-norm relative to the unshifted Hamiltonian" — but we never compute the unshifted norm, so the reduction property isn't actually asserted. The Python version does it properly (`test_hamiltonian_regularizer.py:99-107` compares `norm_after.total <= norm_before.total`).

Could we just mirror the Python test here? Something like:

```suggestion
  auto norm_before = qdk::chemistry::utils::hamiltonian_one_norm(*ham, 0.0);
  EXPECT_LE(norm_after.total, norm_before.total + 1e-10);
  EXPECT_NEAR(norm_after.total, 27.590504297492, 1e-6);
```

Also, could you add a line saying where `27.590504297492` came from? It matches the 27.6 that Loaiza & Izmaylov report for H2O/STO-3G in Table I of arXiv:2304.13772, which is a nice check — but without a note, the next person who sees this drift won't know whether they're looking at a real regression or a BLAS difference.

cpp/src/qdk/chemistry/algorithms/microsoft/fermionic_low_rank/fermionic_low_rank_regularizer.cpp
```diff
@@ -0,0 +1,212 @@
...
+      one_electron.lambda_1e, global_shift.lambda_df_shifted,
+      global_shift.lambda_df_baseline, global_shift.lambda_df_shifted,
+      one_electron.lambda_1e_baseline, one_electron.lambda_1e, one_electron.mu1,
+      global_shift.mu2);
+
+  BlissShift shift;
+  shift.mu1 = one_electron.mu1;
+  shift.mu2 = global_shift.mu2;
+  shift.xi = global_shift.xi;
+  return shift;
```
RushiGong: We compute `lambda_total_before` and `lambda_total_after` just above, log them, and then return the shift regardless of which is larger. Tried this on a batch of synthetic PSD two-electron tensors with the right 8-fold symmetry (small `norb`, sweeping the electron sector) and a fair number of them came back with the shifted 1-norm *higher* than the unshifted one — in the worst case by quite a lot. FCI energy stayed invariant to ~1e-16 throughout, which is the awkward part: the invariance tests can't catch this, because invariance holds for any (mu1, mu2, xi) whether or not the shift helps. But since we already have both numbers right here, would it be worth either falling back to a zero shift when `lambda_total_after > lambda_total_before`, or at least warning instead of only logging at debug level? Silently returning a shift that makes the objective worse seems like the one outcome a caller wouldn't expect.

cpp/src/qdk/chemistry/utils/double_factorization.cpp
```diff
@@ -0,0 +1,122 @@
...
+  Eigen::MatrixXd supermatrix_eigenvectors = supermatrix;
+  Eigen::VectorXd supermatrix_eigenvalues(pair_dim);
+  // lapack::syev overwrites its input buffer in place with the eigenvectors
+  // and only reads the lower triangle (matching Eigen::SelfAdjointEigenSolver's
+  // default convention). Job::Vec is needed here since the eigenvectors are
+  // used below to build each fragment's orbital matrix.
+  const int64_t supermatrix_info = lapack::syev(
+      lapack::Job::Vec, lapack::Uplo::Lower, static_cast<int64_t>(pair_dim),
+      supermatrix_eigenvectors.data(), static_cast<int64_t>(pair_dim),
+      supermatrix_eigenvalues.data());
```
RushiGong, line 55: This builds the dense (norb²×norb²) supermatrix and takes a full eigendecomposition with eigenvectors, then runs another norb×norb `syev` per fragment below. That's O(norb⁶) time and O(norb⁴) memory, and the truncation check on line 79 happens *after* this, so `truncation_threshold` can't reduce the dominant cost.

Extrapolating from timings of the equivalent decomposition, that's roughly a couple of seconds at norb=50 but minutes at norb=100 and well over an hour at norb=150, with peak dense storage in the tens of GB by norb=150 (there are also several simultaneous norb⁴ buffers live — this matrix, the copy on line 46, and the temporary from the symmetrization on line 44).

Given we already have a pivoted Cholesky of the same supermatrix in `microsoft/cholesky_hamiltonian.cpp` and a `CholeskyHamiltonianContainer` that stores norb²×L factors, could this consume those instead of re-deriving a full eigendecomposition? Related: because `compute_fermionic_low_rank_shift` calls `get_two_body_integrals()`, feeding it a Cholesky-backed Hamiltonian will densify the whole four-center tensor first, which rather defeats the point of storing it factored.

If reworking it is out of scope here, could we at least document the scaling and a practical orbital limit on `double_factorize()`? Right now nothing warns a caller before they try this on a production-sized active space.

cpp/include/qdk/chemistry/utils/hamiltonian_one_norm.hpp
```diff
@@ -0,0 +1,72 @@
...
+///      Heff_ij = h_ij + sum_k g[i,j,k,k] - 1/2 sum_k g[i,k,k,j],
+///    and take lambda_1e = sum_i |gamma_i| over its eigenvalues (Eq. 15).
+///
+/// NOTE ON CONVENTION: this decomposes the PHYSICAL coefficient V = 1/2 * g.
+/// qdk::chemistry::utils::double_factorize is elsewhere applied to the full
+/// tensor g (e.g. by HamiltonianRegularizer implementations), so the
+/// two_body value here is exactly half of that pipeline's aggregated
+/// lambda_df; both are internally consistent, but only V = 1/2 * g reproduces
+/// the electronic Hamiltonian exactly (and matches the paper's absolute
+/// 1-norms).
```
RushiGong: I think this note is out of date. It says `double_factorize` is applied elsewhere to the full `g`, but `fermionic_low_rank_regularizer.cpp:179` does `const Eigen::VectorXd two_body_coefficient = 0.5 * g_aaaa;` — the same `V = 1/2 * g` this function uses. So the "exactly half" relationship doesn't hold; the two should agree.

Since this is a public header, could we correct it to say both paths factorize `V = 1/2 * g`, and that the factor-of-two caveat only applies to a caller who hands the raw `g` straight to `double_factorize()` themselves?

cpp/src/qdk/chemistry/algorithms/symmetry_shift.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                 mu2 * num_electrons * num_electrons;
+
+  const Eigen::MatrixXd inactive_fock =
+      original.has_inactive_fock_matrix()
+          ? original.get_inactive_fock_matrix().first
+          : Eigen::MatrixXd(0, 0);
+
+  auto container = std::make_unique<CanonicalFourCenterHamiltonianContainer>(
+      h_tilde, g_tilde, original.get_orbitals(), core_energy_new, inactive_fock,
+      original.get_type());
```
RushiGong, line 139: The header says everything other than the integrals and core energy — including "Hamiltonian type" — is carried over from `original`, but the output is always a `CanonicalFourCenterHamiltonianContainer` regardless of what came in. `get_type()` is preserved; the container isn't.

For a Cholesky- or sparse-backed input that means either an undocumented throw from `get_two_body_integrals()` or a silent O(norb⁴) densification, and neither is in the `@throws` list on the declaration.

Could we say so explicitly in the doc comment — something like "only dense four-center Hamiltonians are supported, and the result is always a canonical four-center container"? Cheap, and it stops the next reader assuming round-trip preservation.

python/src/pybind11/algorithms/hamiltonian_regularizer.cpp
```diff
@@ -0,0 +1,237 @@
...
+// license information.
+
+#include <pybind11/eigen.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <qdk/chemistry.hpp>
+
+#include "factory_bindings.hpp"
+#include "qdk/chemistry/algorithms/microsoft/fermionic_low_rank/fermionic_low_rank_regularizer.hpp"
```
RushiGong: This include looks unused — nothing from `microsoft::fermionic_low_rank` is referenced in this file (the only other mentions of the name are inside docstrings). The sibling bindings pull in their `microsoft/` headers because they actually bind the concrete class (`active_space.cpp:11-14` → `py::class_<microsoft::OccupationActiveSpaceSelector, ...>`, same in `mc.cpp` and `localizer.cpp`), but here we only bind the base.

```suggestion
```

cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
```diff
@@ -0,0 +1,297 @@
...
+ * @param input_num_electrons Target number of active electrons (Ne); the
+ *        invariance guarantee only holds for an integer electron count.
+ * @return The BLISS-shifted Hamiltonian.
+ *
+ * @throws std::invalid_argument if `original` is unrestricted or `shift.xi`
+ *         is not norb x norb.
+ */
+std::shared_ptr<data::Hamiltonian> rebuild_bliss_shifted_hamiltonian(
+    const data::Hamiltonian& original, const BlissShift& shift,
+    unsigned int input_num_electrons);
```
RushiGong: Small one: the parameter is `input_num_electrons` here but the binding exposes it as `py::arg("num_electrons")`, and the doc comment above and the prose on lines 128-132 both call it `Ne`. The `input_` prefix doesn't appear anywhere else in the repo (`get_total_num_electrons`, `n_active_alpha_electrons`, etc.).

Since the Python `py::arg` name is a keyword argument and therefore part of the shipped API, could we align both on `num_electrons` before this goes out?

cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
```diff
@@ -0,0 +1,297 @@
...
+ *   no truncation (exact double factorization).
+ */
+class BlissSettings : public qdk::chemistry::data::Settings {
+ public:
+  BlissSettings() {
+    set_default("shift_method", std::string("fermionic_low_rank"),
+                "Method used to compute the BLISS shift (mu1, mu2, xi).",
+                data::ListConstraint<std::string>{
+                    {std::vector<std::string>{"fermionic_low_rank"}}});
+    set_default("df_truncation_threshold", 0.0);
```
RushiGong: Could you please add description and constrains for this setting?

cpp/include/qdk/chemistry/utils/hamiltonian_one_norm.hpp
```diff
@@ -0,0 +1,72 @@
...
+///
+/// @param hamiltonian Restricted Hamiltonian to analyze.
+/// @param df_truncation_threshold Fragments whose reshaped-supermatrix
+///        eigenvalue magnitude is below this are dropped from the two-body
+///        1-norm.
+///        Defaults to 0.0 (no truncation -- exact/lossless factorization
+///        unless the caller explicitly opts into compression).
+/// @return The one-body, two-body, and total DF fermionic 1-norm.
+/// @throws std::runtime_error if the Hamiltonian is not restricted.
+HamiltonianOneNorm hamiltonian_one_norm(
```
RushiGong, line 68: Seems like `hamiltonian_one_norm` doesn't have any production callers — grepping across `cpp/` and `python/`, the only hits are its own definition, its binding, and tests. Nothing in the BLISS pipeline calls it; the regularizer goes straight to `double_factorize`.

RushiGong (changes requested):
Do you plan to include doc and usage example in this PR?

python/src/pybind11/algorithms/hamiltonian_regularizer.cpp
BoqinZhang: After the restructure and rename, the current `python/src/pybind11/algorithms/symmetry_shift.cpp` exposes the constructor of `FermionicLowRankShifter`, so this header is necessary now

cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
BoqinZhang: A large restructure and rename is made. Now the structure is:
public abstract algorithm type `SymmetryShifter`, its header is in `cpp/include/`, the implementation of helpers is in `cpp/src/qdk/chemistry/algorithms/symmetry_shift.cpp`.
Algorithm `FermionicLowRankShifter` inherits `SymmetryShifter`. Its implementation is in `cpp/src/qdk/chemistry/algorithms/microsoft/symmetry_shift/`.

cpp/src/qdk/chemistry/algorithms/hamiltonian_regularizer.cpp
BoqinZhang: SymmetryShifterFactory is added in `hamiltonian_regularizer.cpp`.

cpp/src/qdk/chemistry/algorithms/microsoft/fermionic_low_rank/fermionic_low_rank_regularizer.cpp
BoqinZhang: A confirmation code and a corresponding warning message is added to handle the case where the lambda increases after shift

cpp/tests/test_symmetry_shift.cpp
BoqinZhang: The name and documentation of the test is modified to describe the intension of the test accurately

cpp/src/qdk/chemistry/algorithms/microsoft/symmetry_shift/fermionic_low_rank.cpp
BoqinZhang: A constructor of struct `GlobalTwoBodyShift` is implemented, in which the number of orbitals is a necessary input, and xi is initialized as (norb, norb) matrix.

cpp/include/qdk/chemistry/utils/hamiltonian_one_norm.hpp
BoqinZhang: The documentation has been corrected

cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
BoqinZhang: The description and constraint for `df_truncation_threshold` has been added

cpp/include/qdk/chemistry/algorithms/hamiltonian_regularizer.hpp
BoqinZhang: The `input_num_electrons` has been renamed as `num_electrons` to be consistent with the python interface

YingrongChen:
Thanks for the pointer — I've added Cholesky to #687 (`double_factorizer/cholesky`, plus a fast path that consumes `CholeskyHamiltonianContainer`'s three-center integrals directly instead of expanding them to a dense tensor and re-decomposing at O(N⁶)).

While porting the idea I hit a bug in this branch's version. It's independent of my PR, so flagging it here.

## `indefinite_tolerance` is an unscaled machine epsilon

`cpp/src/qdk/chemistry/utils/double_factorization.cpp:138`

```cpp
const double indefinite_tolerance = epsilon;   // std::numeric_limits<double>::epsilon()
```

used at `:152` as `minimum_value < -indefinite_tolerance`.

Roundoff in the residual diagonal accumulates like `O(k · eps · ‖M‖)` across the rank-one updates, so the minimum residual diagonal of a **genuinely PSD** supermatrix drifts below `-2.2e-16` well before the factorization is done. The check then reports it indefinite and bails.

I measured this on random PSD tensors of the same construction your tests use (`Σ_t F_t ⊗ F_t`, all signs positive, so PSD by construction). 200 random tensors per row; "bails" means it declared the input indefinite and fell back:

| norb | rank | worst min eigenvalue | bare `eps` bails | scaled bails | scaled recovers exact rank |
|---|---|---|---|---|---|
| 4 | 3 | −2.8e-15 | **73/200** | 0/200 | 200/200 |
| 6 | 3 | −4.6e-15 | **111/200** | 0/200 | 200/200 |
| 8 | 3 | −1.1e-14 | **136/200** | 0/200 | 200/200 |
| 10 | 3 | −1.3e-14 | **161/200** | 0/200 | 200/200 |
| 4 | 6 | −2.8e-15 | **147/200** | 0/200 | 200/200 |
| 6 | 8 | −6.7e-15 | **194/200** | 0/200 | 200/200 |
| 8 | 12 | −1.8e-14 | **199/200** | 0/200 | 200/200 |

Every one of these inputs is PSD to within 2e-14; only the tolerance is wrong. The rate climbs with both orbital count and rank, which is the signature of accumulated roundoff rather than of anything about the input — by norb=8 at rank 12 it is 99.5%.

(Worth noting it is *seed-dependent* at small sizes, so a single passing test case proves nothing here — that's part of why this is easy to miss.)

## Why it's worth fixing even though nothing is incorrect

The fallback to eigendecomposition still returns a correct factorization, and the only trace is a debug log. So there's no wrong answer and no exception — the failure mode is that **the Cholesky path silently never engages for any molecule large enough to care about**, and you keep paying O(N⁶). That's the kind of thing that survives a green CI indefinitely, which is why I thought it was worth writing up rather than just fixing on my side.

## Fix

Scale by both the magnitude and the dimension:

```cpp
const double scale = std::max(residual_diagonal.maxCoeff(), 1.0);
const double noise_floor = std::numeric_limits<double>::epsilon() * scale *
                           static_cast<double>(pair_dim);
```

and use it for the stopping threshold too, since `stop_threshold = max(truncation_threshold, epsilon)` at `:137` has the same problem when `truncation_threshold` is small.

I checked this doesn't blunt the detection: over **1798 genuinely indefinite** random tensors (min eigenvalue < −1e-10, six different sign patterns, norb 2–6), the scaled tolerance caught every single one — zero cases where it claimed a Cholesky factorization existed and the reconstruction was wrong. And as the table shows, it recovers the exact rank in 1400/1400 PSD cases.

## Two smaller things I ran into

- **`‖ε‖²` is not the pivot.** I initially assumed the fragment weight could be read off the pivot, which would have let Cholesky and eigen share a truncation threshold for free. It's wrong: `‖ε‖²` equals the squared norm of the whole Cholesky column, while the pivot is only `L[pivot,k]²`. In #687 I run the loop to the noise floor and then sort/filter by `‖ε‖²`, which makes `truncation_threshold` mean the same thing in both methods — identical ranks for norb 2..8 at 1e-12, 1e-6 and 1e-3. Worth being deliberate about, otherwise the same setting silently means two different things depending on which factorizer you picked.

- **The `N(N+1)/2` rank bound is only valid if `p↔q` symmetry actually holds.** Since the code imposes `(pq)↔(rs)` by averaging but not `p↔q`, I treat exhausting that bound with residual left over as a signal to fall back rather than as an invariant.

## Going forward

That makes four implementations of this math across #687, #623, #539 and `main`'s Cholesky container. Once #687 lands, `cholesky_decompose_two_body()` / `eigen_decompose_two_body()` are public C++ API you can call directly, which was the original point of extracting them — happy to do the consolidation myself if that's easier for you.

Reproduction scripts for all the numbers above are available if you want them.

PR #629, opened 2026-08-10
Correct IQPE doc from MSB-first to LSB-first

### LINKED ISSUE #618: Correct IQPE execution-order documentation from MSB-first to LSB-first
## Problem

The iterative phase-estimation implementation executes iteration `k = 0` first with power `U^(2^(m-1))`. In the implemented feedback recurrence, this first measurement determines the least-significant phase bit. Subsequent iterations proceed toward the most-significant bit. The implementation then reverses the execution-order list when constructing `QpeResult.bits_msb_first`.

Several source docstrings, general docs, and examples instead say that execution proceeds MSB-to-LSB or that iteration 0 corresponds to the MSB. This contradicts the implementation and the ground-state QPE tutorial, which correctly describes the first largest-power iteration as measuring the LSB.

## Implementation evidence

- `python/src/qdk_chemistry/algorithms/phase_estimation/iterative_phase_estimation.py` loops over `iteration = 0..m-1`, appends each measured bit, updates `phase_feedback = phase_feedback / 2 + pi * measured_bit / 2`, and returns `bits_msb_first=bits[::-1]`.
- Native and Qiskit iterative builders both use `power = 2 ** (total_iterations - iteration - 1)`.
- Therefore iteration 0 is executed first with the largest power and contributes the LSB; the final iteration uses `U` and contributes the MSB.

## Known affected locations

- `python/src/qdk_chemistry/algorithms/phase_estimation/iterative_phase_estimation.py` module summary
- `python/src/qdk_chemistry/algorithms/phase_estimation/circuit_builder/iterative_builder.py` iteration parameter docstring
- `python/src/qdk_chemistry/plugins/qiskit/phase_estimation_circuit_builder.py` iteration parameter docstring
- `docs/source/user/comprehensive/algorithms/phase_estimation.rst`
- `docs/source/user/features.rst`
- `examples/qpe_stretched_n2.ipynb`
- `examples/extended_hubbard.ipynb`

The `QpeResult.bits_msb_first` documentation and logs that describe the **returned** bit list as MSB-first are correct and should remain unchanged.

## Suggested resolution

1. Correct all execution-order descriptions to LSB-first, MSB-last.
2. Explicitly distinguish execution order from the conventional MSB-first returned bitstring.
3. Add or strengthen a focused regression/documentation test that makes this distinction visible.
4. Audit other IQPE examples and docs for equivalent wording.

This was identified while addressing technical review feedback on PR #604.

RushiGong:
close #618

### DIFF
--- a/examples/interoperability/qiskit/iqpe_no_trotter.py
+++ b/examples/interoperability/qiskit/iqpe_no_trotter.py
@@ -270,14 +270,6 @@ def _energy_from_phase(phase_fraction: float) -> float:
     return angle / T_TIME
 
 
-def _resolve_energy(raw_energy: float, reference: float) -> tuple[list[float], float]:
-    period = 2 * np.pi / T_TIME
-    candidates = sorted(
-        {s * raw_energy + period * k for k in range(-2, 3) for s in (1, -1)}
-    )
-    return candidates, min(candidates, key=lambda e: abs(e - reference))
-
-
 result = QpeResult.from_phase_fraction(
     method="iterative_exact",
     phase_fraction=phase_fraction,
@@ -286,19 +278,14 @@ result = QpeResult.from_phase_fraction(
     bits_msb_first=bits[::-1],
 )
 
-raw_energy = result.raw_energy
-candidate_energies, resolved_energy = _resolve_energy(raw_energy, casci_energy)
-estimated_total_energy = resolved_energy + core_energy
+estimated_electronic_energy = result.raw_energy
+estimated_total_energy = estimated_electronic_energy + core_energy
 
 Logger.info(f"Measured bits (MSB → LSB): {list(result.bits_msb_first or [])}")
 Logger.info(
     f"Phase fraction φ (measured): {result.phase_fraction:.6f} (angle = {result.phase_angle:.6f} rad)"
 )
-Logger.info(f"Raw energy_from_phase output: {raw_energy:+.8f} Hartree")
-Logger.info("Candidate energies (alias checks):")
-for energy in candidate_energies:
-    Logger.info(f"  E = {energy:+.8f} Hartree")
-Logger.info(f"Estimated electronic energy: {resolved_energy:.8f} Hartree")
+Logger.info(f"Estimated electronic energy: {estimated_electronic_energy:.8f} Hartree")
 Logger.info(f"Estimated total energy: {estimated_total_energy:.8f} Hartree")
 Logger.info(f"Reference total energy (CASCI): {casci_energy:.8f} Hartree")
 iterative_energy_error = estimated_total_energy - casci_energy
--- a/examples/interoperability/qiskit/iqpe_no_trotter.py
+++ b/examples/interoperability/qiskit/iqpe_no_trotter.py
@@ -127,8 +127,7 @@ def run_iterative_exact_qpe(
 
     Returns:
         Tuple ``(bits, phase_fraction)`` with the measured bits in execution order
-        (LSB first, MSB last) and
-        the resulting phase fraction.
+        (LSB first, MSB last) and the resulting phase fraction.
 
     Notes:
         This routine runs the iterativeqpe with the exact evolution
@@ -283,7 +282,8 @@ result = QpeResult.from_phase_fraction(
     method="iterative_exact",
     phase_fraction=phase_fraction,
     eigenvalue_from_phase=_energy_from_phase,
-    bits_msb_first=bits,
+    # ``bits`` come back in execution order (LSB first), so reverse them to get MSB first.
+    bits_msb_first=bits[::-1],
 )
 
 raw_energy = result.raw_energy
--- a/examples/interoperability/qiskit/iqpe_no_trotter.py
+++ b/examples/interoperability/qiskit/iqpe_no_trotter.py
@@ -126,7 +126,8 @@ def run_iterative_exact_qpe(
         synthesis: Matrix exponential synthesizer employed by Qiskit.
 
     Returns:
-        Tuple ``(bits, phase_fraction)`` with the measured bits (MSB to LSB) and
+        Tuple ``(bits, phase_fraction)`` with the measured bits in execution order
+        (LSB first, MSB last) and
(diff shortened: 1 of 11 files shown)


examples/interoperability/qiskit/iqpe_no_trotter.py
```diff
@@ -126,7 +126,8 @@ def run_iterative_exact_qpe(
         synthesis: Matrix exponential synthesizer employed by Qiskit.
 
     Returns:
-        Tuple ``(bits, phase_fraction)`` with the measured bits (MSB to LSB) and
+        Tuple ``(bits, phase_fraction)`` with the measured bits in execution order
+        (LSB first, MSB last) and
```
wavefunction91: The `Returns:` here now correctly says the bits come back in execution order (LSB first) — but line 286 below passes them straight into `QpeResult.from_phase_fraction(bits_msb_first=bits)` with no reversal, and line 293 logs the result as "MSB → LSB". The library path does `bits_msb_first=bits[::-1]` (iterative_phase_estimation.py:139).

Worth checking: execution-order bits [1,1,0,1] give phase_fraction 0.6875 = 1011b, so the MSB-first string is [1,0,1,1], not [1,1,0,1]. `QpeResult` doesn't cross-check bits against the phase fraction, so this stores silently.

I think this predates the PR, but the docstring fix makes the file internally inconsistent, and issue #618 lists auditing the IQPE examples as part of the resolution — so it may belong here. That does mean pulling line 286 into the diff.

python/tests/test_phase_estimation_iterative.py
```diff
@@ -541,3 +541,72 @@ def test_iterative_qpe_raises_on_negative_num_bits(two_qubit_phase_problem: Phas
...
+
+    scripted_executor = _ScriptedExecutor(execution_order_bits)
+    original_create_nested = iqpe._create_nested
+
+    def _patched_create_nested(setting_name: str):
+        if setting_name == "circuit_executor":
+            return scripted_executor
+        return original_create_nested(setting_name)
+
+    iqpe._create_nested = _patched_create_nested  # type: ignore[method-assign]
```
wavefunction91: Could you double-check this ignore comment? Running mypy with the package resolvable gives:

```
test_phase_estimation_iterative.py:594: error: Incompatible types in assignment (expression has type "Callable[[Arg(str, 'setting_name')], Any]", variable has type "Callable[[Arg(str, 'setting_key')], Any]")  [assignment]
test_phase_estimation_iterative.py:594: note: Error code "assignment" not covered by "type: ignore" comment
```

Two things: the code is `[assignment]`, not `[method-assign]`, so the ignore silences nothing; and the parameter is named `setting_name` where `Algorithm._create_nested` names it `setting_key`. This is the only new mypy error the branch adds (8 on main, 9 here). CI's pre-commit doesn't catch it because that hook runs mypy in an isolated env where `qdk_chemistry` is unresolvable.

Would substituting the executor the way `test_algorithms.py` does avoid the patch entirely? It defines a mock subclass and injects it via `algorithms.register(...)` / `unregister(...)` (test_algorithms.py:178-199, 636, 731-740), which keeps this on the public Settings path.

python/tests/test_phase_estimation_iterative.py
```diff
@@ -541,3 +541,72 @@ def test_iterative_qpe_raises_on_negative_num_bits(two_qubit_phase_problem: Phas
...
+            self.bitstring_counts = {str(bit): 1}
+
+    class _ScriptedExecutor:
+        """Returns the scripted bit for each successive iteration circuit."""
+
+        def __init__(self, bits: list[int]) -> None:
+            self._bits = list(bits)
+            self._index = 0
+
+        def run(self, circuit: Circuit, shots: int, noise: QuantumErrorProfile | None = None) -> _ScriptedResult:  # noqa: ARG002
```
wavefunction91: `run` ignores `circuit` and returns canned counts, so the measured bits don't depend on the circuit that was built. That seems to mean this test can't observe which power went with which iteration — changing `power = 2 ** (total_iterations - iteration - 1)` to `2 ** iteration` in iterative_builder.py:150 wouldn't move either assertion, even though "measures LSB first" is exactly a claim about that pairing.

The post-processing half also looks largely covered already: `test_iterative_phase_estimation_extracts_phase_and_energy` asserts `bits_msb_first == [1,1,0,1]` alongside `phase_fraction == 13/16`, and 1101b = 13, so dropping the `bits[::-1]` reversal fails it (reversed = 1011b = 11).

Is there a smaller version that pins the thing the name promises — e.g. asserting the builder's power directly at iteration 0 vs iteration n-1? Genuinely asking; if the intent is a simulator-independent statement of the convention, that's reasonable, it just may not need 69 lines and a private-method patch.

python/src/qdk_chemistry/algorithms/phase_estimation/iterative_phase_estimation.py
```diff
@@ -1,8 +1,10 @@
 """Iterative phase estimation implementation.
 
 This module implements the Kitaev-style iterative quantum phase estimation (IQPE)
-algorithm, which measures phase bits sequentially from most-significant to least-significant
-using a single ancilla qubit and adaptive feedback corrections.
+algorithm, which measures phase bits sequentially from least-significant to most-significant
+using a single ancilla qubit and adaptive feedback corrections. The first iteration applies
+the largest controlled power :math:`U^{2^{m-1}}` and therefore measures the least-significant
+bit; subsequent iterations proceed toward the most-significant bit.
```
wavefunction91: `m` isn't defined anywhere in this docstring, and the comprehensive doc this pairs with uses `n` ("where :math:`n` is the total number of bits", phase_estimation.rst:149). The tutorial does use `m`, so either symbol works — it just needs an anchor here.

```suggestion
the largest controlled power :math:`U^{2^{m-1}}` (``m`` = ``num_bits``) and therefore measures
the least-significant bit; subsequent iterations proceed toward the most-significant bit.
```

wavefunction91 (changes requested):

wavefunction91 (approved):

PR #610, opened 2026-08-03
Add amplitude amplification algorithm

YingrongChen:
Adds amplitude amplification as a registry algorithm — builds the amplified circuit $Q^k U_\psi$ from a state preparation and a good-state oracle. Executing that circuit and deciding which shots are good is left to the caller, which keeps the algorithm independent of phase estimation. `run(state_prep_oracle, good_state_oracle)` takes just the two oracles and reads the register width from a resource estimate of the state preparation.

The only setting is `rounds` (`int`, default `1`), the number of plain Grover rounds. The QPE circuit builder gains a `measure_phase` setting (`bool`, default `True`) that returns exactly the measurement-free adjointable circuit that amplification needs when set to `False`. Also included: `phase_marking_oracle`, a helper that takes a standard QPE circuit and marks a half-open range of phase bins, reading the register layout off that circuit.

The Q# primitives live in `python/src/qdk_chemistry/utils/qsharp/src/AmplitudeAmplification.qs`, with the docs page `docs/source/user/comprehensive/algorithms/amplitude_amplification.rst` and its tested example `docs/source/_static/examples/python/amplitude_amplification.py`.

PR #630, opened 2026-08-11
Make the phase marking oracle a registered algorithm class

YingrongChen:
`phase_marking_oracle` shipped in #610 as a module-level function, which sits outside the registry every other algorithm goes through.


RushiGong (changes requested):
For the AA related algorithms, can we create a folder for it and structure these algorithms to be consistent with other algorithms code/folder structure?

python/src/qdk_chemistry/algorithms/amplitude_amplification/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+from .qpe_subspace import (
+    QPESubspaceMarking,
+    QPESubspaceMarkingSettings,
+)
+
+__all__: list[str] = [
+    "AmplitudeAmplificationFactory",
+    "AmplitudeAmplificationSettings",
+    "QPESubspaceMarkingSettings",
+]
```
RushiGong, line 26: Can you double check the `__all__` export here? Seems like not everything imported are called in `__all__`

docs/source/user/comprehensive/algorithms/amplitude_amplification.rst
```diff
@@ -1,16 +1,16 @@
...
 =======================
 
-The :class:`~qdk_chemistry.algorithms.amplitude_amplification.AmplitudeAmplification`
+The :class:`~qdk_chemistry.algorithms.amplitude_amplification.amplitude_amplification.AmplitudeAmplification`
 algorithm increases the probability of measuring a state in a chosen subspace.
 It takes two :class:`~qdk_chemistry.data.Circuit` objects:
 
 - ``state_prep_oracle`` prepares the initial state.
-- ``good_state_oracle`` flips a flag qubit for the good subspace.
+- ``good_state_oracle`` reflects about the good subspace.
```
RushiGong: Does this `good_state_oracle` a state or it carries the reflect operations?

python/src/qdk_chemistry/utils/qsharp/src/AmplitudeAmplification.qs
```diff
@@ -11,137 +11,113 @@
...
-                Controlled X([lowerFlag, upperFlag], target);
-            }
+        use aboveLower = Qubit();
+        use belowUpper = Qubit();
+        within {
+            ApplyIfLessOrEqualL(X, IntAsBigInt(lowerBound), phase, aboveLower);
+            ApplyIfGreaterOrEqualL(X, IntAsBigInt(upperBound - 1), phase, belowUpper);
+        } apply {
+            Controlled X([aboveLower, belowUpper], target);
         }
```
RushiGong: Do we need bounds checking in the Q# layer or it is well handled at python layer? `lowerBound >= upperBound` marks nothing and raises nothing. `upperBound == 0` also ends up comparing against `IntAsBigInt(-1)`. 

The old code collapsed a full-range interval to a single `X(target)`; now every range costs two ancillas and two comparators. Any efficiency concern here?

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,199 @@
...
+
+        phase_bin_count = 1 << num_phase_qubits
+        ranges: list[tuple[int, int]] = []
+        for phase_bin in range(phase_bin_count):
+            if eigenvalue_from_phase(phase_bin / phase_bin_count) < target_energy:
+                continue
+            if ranges and ranges[-1][1] == phase_bin:
+                ranges[-1] = (ranges[-1][0], phase_bin + 1)
+            else:
+                ranges.append((phase_bin, phase_bin + 1))
```
RushiGong, line 123: This scans every bin, so it's `2**num_bits` evaluations. It's fine at the 3-4 bits the tests use, but large number of bits would lower the efficiency of Python calls. Do we need to search at all though? Both containers have closed-form laws (walk is `λcos(2πφ)`, trotter is `-angle/t`), and both are monotonic on each half of `[0, 1)`, so the accepted set is always at most two intervals. Could we bisect for the boundaries?

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,199 @@
...
+            QSHARP_UTILS.StatePreparation.MakePrepareNothingOp(),
+            controlled_unitaries,
+            num_bits,
+            list(range(num_bits)),
+            [index + num_bits for index in range(num_system_qubits)],
+            QSHARP_UTILS.StatePreparation.MakePrepareHadamardAllOp(),
+            num_ancilla_qubits,
+        )
+
+        container = self._create_nested("unitary_builder").run(qubit_hamiltonian).get_container()
```
RushiGong: This runs the unitary builder a full extra time just to get `eigenvalue_from_phase` , `_create_controlled_circuit` already ran it once per bit in the loop above, so it's `num_bits + 1` builds total. Maybe have `_create_controlled_circuit` return the container alongside the circuit, or cache it on first use. There maybe some efficiency improvement for the circuit builder.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,199 @@
...
+            RuntimeError: If the controlled unitaries do not carry Q# operations.
+
+        """
+        Logger.trace_entering()
+        num_bits = self._settings.get("num_bits")
+        if num_bits <= 0:
+            raise ValueError(f"num_bits must be a positive integer. Got {num_bits}.")
+        target_energy = float(self._settings.get("target_energy"))
+        if math.isnan(target_energy):
+            raise ValueError("The target_energy setting must be set to the energy of the subspace to mark.")
```
RushiGong: Here check for nan, can you combine the infinite check from `_mark_phase_bin` here as well?

docs/source/user/comprehensive/algorithms/qpe_circuit_builder.rst
```diff
@@ -78,6 +78,14 @@ Available Implementations
 
 QDK/Chemistry provides two primary implementations of :class:`~qdk_chemistry.algorithms.phase_estimation.circuit_builder.base.QpeCircuitBuilder`:
```
RushiGong, line 79: Maybe be more clear around this, two primary implementation for phase estimation, also other implementations. Not just mentioning the note. Also having a section talking about the qpe subspace builder would be useful and AmpAmp QPE section can use reference link to it.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Going to leave this as is. _create_controlled_circuit lives in the shared QpeCircuitBuilder base and is used by the standard, iterative, and Qiskit builders, so changing its return signature to hand back the container would ripple through all of them (and their overrides) just to serve this one call site.

The extra build is also cheap in practice — the expensive part is the circuit mapping in _create_controlled_circuit, which this extra call doesn't touch. StandardPhaseEstimation already resolves the container the same way (_create_nested("unitary_builder").run(...) before running the circuit builder).

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,205 @@
...
+        super().__init__()
+        self._set_default(
+            "target_energy",
+            "double",
+            math.nan,
+            "Lowest energy the marked subspace may hold. Required.",
+        )
+
+
+class QPESubspaceMarking(QpeCircuitBuilder):
```
RushiGong: Having some concern here for whether this algorithm should be registered under qpe circuit builder. 
It's fine for me to have a new algorithm, and it will have flexibility for data type used in `run()`. As state preparation is not used when buiding qpe subspace but it is still called.

python/src/qdk_chemistry/utils/qsharp/src/AmplitudeAmplification.qs
```diff
@@ -11,137 +11,139 @@
...
-            } apply {
-                Controlled X(signalAncillas + [inRange], target);
-            }
+        use phase = Qubit[numPhaseQubits];
+        use signalAncillas = Qubit[numSignalAncillas];
+        within {
+            qpe(phase + system + signalAncillas);
+        } apply {
+            MarkAcceptedPhase(lowerBounds, upperBounds, phase, signalAncillas, target);
         }
```
RushiGong, line 113: Could you check this against a target that isn't at the band edge?

`within { qpe } apply { mark }` gives `QPE† · (Π_window ⊗ |0⟩⟨0|_signal) · QPE`. That's a reflection on `system` only if the middle is diagonal in the eigenbasis QPE just wrote. Two ways it isn't:

- **Signal ancillas.** The walk eigenvectors are `|w±⟩ = (|0⟩_sig|E⟩ ± i|⊥⟩)/√2`, so each has weight on both `|0⟩_sig` and `|⊥⟩`. Conditioning on `signal = 0` projects *inside* the walk eigenspace rather than post-selecting a good block encoding, leaving `Adjoint qpe` nothing clean to undo.
- **Inexact QPE.** Off a bin centre the phase register is spread, the marking partially projects that spread, and the register doesn't come back to `|0⟩`.

Either way `system` stays entangled with the released ancillas, contradicting the summary above ("comes back as it was found").

What stopped me concluding the tests disprove this: they dodge both cases simultaneously.

```
state  E        E/lambda  phase   band edge?
 11    +pi/2     +1.0     0.000   YES (branches merge)
 01/10  0.0       0.0     0.250   no
 00    -pi/2     -1.0     0.500   YES
```

`target_energy = lambda/2` marks `[0, 1/6] ∪ [5/6, 1)`, and the only eigenstate in there is `|11⟩` at `E = +lambda`. That's `arccos(+1) = 0`, exactly where the two branches coincide — so `|0⟩_sig|E⟩` genuinely is an eigenvector of `W`, the signal ancilla does return to `|0⟩`, and phase 0 is exactly bin 0 at any `num_bits`. The trotter test dodges it differently: `pauli_sequence` has no signal ancillas and `phi = 3/4` is exactly bin 12/16.

So rounds 0/1/2 passing in `test_amplified_qpe_acceptance_follows_the_round_count` doesn't exercise the general case.

Cheapest way to settle it: add a Hamiltonian with an interior eigenvalue (e.g. `aZ + bX`, so `E/lambda` isn't `0` or `±1`), mark that eigenstate, and check `P(good)` still tracks `sin²((2k+1)·arcsin√a)` for k = 0..3. If it holds, I'm being paranoid and this comment is fine as written. If not, the likely fix is dropping `signalAncillas` from the marking condition and documenting that the oracle assumes exact QPE.

docs/source/_static/examples/python/amplitude_amplification.py
```diff
@@ -21,73 +21,52 @@
...
-# marks both signs of that phase. Here it selects bin 8 again.
-good_state_oracle = phase_marking_oracle(
-    state_prep_oracle,
-    target_energy_range=(-np.inf, -0.99 * qubit_hamiltonian.schatten_norm),
-    qubit_hamiltonian=qubit_hamiltonian,
-)
+    controlled_circuit_mapper=AlgorithmRef(
+        "controlled_circuit_mapper", "prepare_select_prepare"
+    ),
+    target_energy=qubit_hamiltonian.schatten_norm / 2,
```
RushiGong: The script used to take a range of energy, here changes to one target value. Why is this change?

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,205 @@
...
+            bin is marked twice.  Never empty.
+
+        Raises:
+            ValueError: If the encoding cannot represent the target energy, or if it
+                can but no bin of the register reaches it.
+
+        """
+        phase_bin_count = 1 << num_phase_qubits
+        run_starts = {0, (phase_bin_count // 2 + 1) % phase_bin_count}
+        for phase in container.phases_from_eigenvalue(target_energy):
```
RushiGong: A target below the band used to mark every bin; now it raises, and the `below-band` test case went away with it.

Is that deliberate? The two ends feel different to me: above the band nothing can match, so raising makes sense — but below the band *everything* matches, which seems a reasonable thing to ask for, and it's what you'd naturally pass if the bound came from a rough estimate.

Related: `test_marking_oracle_circuit_is_executable` now has to pass exactly `-schatten_norm` to mark everything, which feels fragile.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,205 @@
...
+            Sorted half-open bin ranges, separated by at least one bin so that no
+            bin is marked twice.  Never empty.
+
+        Raises:
+            ValueError: If the encoding cannot represent the target energy, or if it
+                can but no bin of the register reaches it.
+
+        """
+        phase_bin_count = 1 << num_phase_qubits
+        run_starts = {0, (phase_bin_count // 2 + 1) % phase_bin_count}
```
RushiGong: Tried a container with the same law but its cut at `phi = 1/4`, and this silently returns wrong bins for it (drops three, no error). Since this PR just moved the phase convention into the containers, could the discontinuities come from there too?

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,205 @@
...
+        num_bits = self._settings.get("num_bits")
+        if num_bits <= 0:
+            raise ValueError(f"num_bits must be a positive integer. Got {num_bits}.")
+        target_energy = float(self._settings.get("target_energy"))
+        if math.isnan(target_energy):
+            raise ValueError("The target_energy setting must be set to the energy of the subspace to mark.")
+        if math.isinf(target_energy):
+            raise ValueError(f"The target_energy setting must be a finite energy. Got {target_energy}.")
+
+        container = self._create_nested("unitary_builder").run(qubit_hamiltonian).get_container()
```
RushiGong: This builds the container, then the loop below builds it again once per bit: `num_bits + 1` total, of which `num_bits` are redundant. Two reasons they're avoidable — this call uses the default `power=1`, which is exactly what the last iteration (`bit = num_bits - 1`) rebuilds, and for LCU the PREPARE/SELECT structure doesn't depend on `power` at all. Not free either: a rough benchmark puts it around 90% of classical setup time for a large LCU Hamiltonian, since each build re-parses every Pauli term. Could `_create_controlled_circuit` return the container alongside the circuit?

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,205 @@
...
+                can but no bin of the register reaches it.
+
+        """
+        phase_bin_count = 1 << num_phase_qubits
+        run_starts = {0, (phase_bin_count // 2 + 1) % phase_bin_count}
+        for phase in container.phases_from_eigenvalue(target_energy):
+            # The crossing is solved in floating point, so it is only known to within a
+            # bin: splitting on either side of the bins it could land in keeps the runs uniform.
+            crossing = math.floor(phase % 1.0 * phase_bin_count)
+            run_starts |= {(crossing + offset) % phase_bin_count for offset in (-1, 0, 1, 2)}
```
RushiGong: The comment says the crossing "is only known to within a bin", which is what justifies the `(-1, 0, 1, 2)` window. Do we need to consider senario of a larget `num_bits`?

`arccos` is ill-conditioned at the band edge: a 1-ULP error in `E/scale` becomes a `sqrt(2·eps)` error in phase, not an `eps`-sized one. Worst-case crossing error is about `N · 3.35e-9` bins, which passes 1 bin around `num_bits = 29`:

```
n=20:  0.004 bins      n=28:  0.90 bins
n=29:  1.80 bins       n=32: 14.41 bins
```

Concretely, `scale = 0x1.55564f37c163dp+136`, `target = 0x1.55564f37c163cp+136`, `num_phase_qubits = 32`: computed crossing bin 10, true bin 12, giving `[(0, 2147483649), (4294967284, 4294967296)]` where brute force gives `[(0, 17), (4294967280, 4294967296)]` — ~2.1 billion bins marked that shouldn't be, silently.

That needs a target within ~1 ULP of the band edge. But the window is a constant where the error isn't, so it can't bound it in general. Could we derive the window from the conditioning, or reject `num_bits` above the width where `±1` is provably safe? Even just recording the limit in the comment beats the current claim.

python/tests/test_amplitude_amplification.py
```diff
@@ -238,134 +284,139 @@ def test_amplification_matches_the_closed_form_and_overshoots():
...
+        }
+        if not expected:
+            # A narrow register can fall short of an energy the encoding still represents.
+            with pytest.raises(ValueError, match="No phase bin"):
+                QPESubspaceMarking._marked_phase_bins(target_energy, container, num_phase_qubits)
+            continue
+        bins = QPESubspaceMarking._marked_phase_bins(target_energy, container, num_phase_qubits)
+        assert {phase_bin for start, stop in bins for phase_bin in range(start, stop)} == expected, context
+        # Sorted, non-empty and separated by at least one bin, so the ranges cannot double-flip
+        # the flag: MarkPhaseRanges applies each one independently.
```
RushiGong: `MarkPhaseRanges` doesn't exist anymore; it merged into `MarkAcceptedPhase`, which loops over the singular `MarkPhaseRange`. The point the comment makes is still right, just the symbol name sends the reader looking for something they won't find.

docs/source/_static/examples/python/amplitude_amplification.py
```diff
@@ -21,73 +21,52 @@
...
+).run(state_preparation, qubit_hamiltonian)[0]
 
-# 5. Amplify, then execute
+# 4. Amplify the initial state against the qpe subspace marking oracle.
 amplitude_amplification = create("amplitude_amplification", "qdk_base", rounds=2)
-circuit = amplitude_amplification.run(state_prep_oracle, good_state_oracle)
+circuit = amplitude_amplification.run(state_preparation, good_state_oracle)
 
+# 5. Run the circuit and measure. The |11> state should be amplified.
```
RushiGong: The example says "The |11> state should be amplified" so we should show how to analyze the result to demonstrate it.

Since amplification is precisely a *before/after* story, could we land the payoff? Even two lines would do it:

```python
print(counts)  # |11> dominates; without amplification it would be ~9% of shots
```

Better still, print the `rounds=0` and `rounds=2` counts side by side — that shows the reader what the algorithm bought them, which is the whole point of the page.

While here: `target_energy=qubit_hamiltonian.schatten_norm / 2` is unexplained magic. A short "halfway up the band, so only the top eigenvector is marked" would save the reader working it out.

RushiGong (changes requested):

docs/source/_static/examples/python/amplitude_amplification.py
YingrongChen: I limit the input for only energy lower bound for now.

python/src/qdk_chemistry/utils/qsharp/src/AmplitudeAmplification.qs
```diff
@@ -12,136 +12,152 @@
...
+            ApplyControlledOnInt(lowerBound, X, phase, target);
         } elif lowerBound == 0 {
-            use encodedUpper = Qubit[numPhaseQubits];
-            within {
-                ApplyXorInPlace(upperBound, encodedUpper);
-            } apply {
-                ApplyIfGreaterLE(X, encodedUpper, phaseRegister, target);
-            }
+            // phase >= 0 always holds, so only the upper bound has to be tested.
+            ApplyIfGreaterOrEqualL(X, IntAsBigInt(upperBound - 1), phase, target);
```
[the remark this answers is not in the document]
YingrongChen: Not a bug. Computes if (c <= x) { action(target) }, that is, applies action to target if a BigInt value c is less or equal to the little-endian qubit register x

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Per our discussion, I moved this out from under `qpe_circuit_builder` into a new `amplitude_amplification_oracle` algorithm type that takes the QPE builder as a setting and runs on the qubit Hamiltonian alone, so it can no longer be selected where a QPE circuit builder is required.

python/src/qdk_chemistry/utils/qsharp/src/AmplitudeAmplification.qs
YingrongChen: Per our discussion, I dropped the signal ancillas from the marking condition so the projector stays diagonal in the walk eigenbasis, and added `test_subspace_oracle_flags_an_interior_eigenstate` plus `test_amplified_qpe_acceptance_at_an_interior_eigenvalue` on a genuinely interior eigenvalue over k = 0..3 — inexact QPE remains a real limitation and is now called out in both the Q# and the class docstring.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Per our discussion, I kept the refusal at both ends because reflecting about every bin is the identity up to a phase and so amplifies nothing, and `test_marking_oracle_circuit_is_executable` no longer leans on an exact `-schatten_norm` — it now marks a real window and only varies where `|00>` sits.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Per our discussion, `_marked_phase_bins` now scans every bin against the container's phase-to-energy law rather than assuming where the law turns, so a law cut at φ = 1/4 gets the right bins — pinned by `test_marked_bins_follow_a_law_that_turns_away_from_phase_half`.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Per our discussion, the oracle now hands the whole QPE circuit to the nested builder and builds the unitary just once itself, so the redundancy drops from `num_bits` extra builds to one; removing that last one would mean changing the shared `QpeCircuitBuilder` contract, which we're keeping out of this PR.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Per our discussion, the search no longer inverts the phase law — it evaluates the law forward on each bin — so `arccos` conditioning at the band edge never enters, and `test_a_target_just_inside_the_band_edge_marks_only_the_top_bin` pins a one-ulp target.

python/tests/test_amplitude_amplification.py
YingrongChen: Per our discussion, the comment now names `MarkAcceptedPhase` and the singular `MarkPhaseRange` it loops over.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,212 @@
...
+
+        Raises:
+            ValueError: If no bin reaches the bound, or if every bin does, because neither
+                names a proper subspace to amplify.
+
+        """
+        phase_bin_count = 1 << num_phase_qubits
+        ranges: list[tuple[int, int]] = []
+        for phase_bin in range(phase_bin_count):
+            if container.eigenvalue_from_phase(phase_bin / phase_bin_count) < energy_lower_bound:
```
RushiGong: I think this comparison can silently undo the fix from previous commit 61f192cd.

That commit dropped `signalAncillas` from the marking, justified by the accepted bins being symmetric under `phi -> 1 - phi` so that both walk branches of an eigenspace are marked together. Bin `b` and bin `N-b` are the same energy mathematically, but they're evaluated at *different* phases here, so `cos` rounds them differently. When `energy_lower_bound` lands exactly on a bin's energy, `>=` accepts one side and rejects the other:

```
N=8, scale=1, bound=cos(pi/4)=0.7071067811865476
  bin 1: 0.7071067811865476  -> accepted (equal)
  bin 7: 0.7071067811865474  -> rejected (1 ULP low)
  result: [(0, 2)]   should be: [(0, 2), (7, 8)]
```

Sweeping bounds set exactly to a bin energy (walk law, scales 0.5–100, widths 1–10): **3970 of 10180 cases come out asymmetric.** Only one branch of the eigenspace gets marked, which is precisely the condition that leaves the ancillas dirty after `Adjoint qpe` — the "released while not in |0> state" failure your commit message describes.

Worth stressing this is invisible to a brute-force cross-check, because the scan agrees with brute force perfectly; it's the floating-point law that isn't symmetric, while the physics requires that it is.

A bound landing exactly on a bin energy isn't exotic either — it's what you get computing the bound from a known eigenvalue, which is the natural thing to do.

Evaluating mirrored bins at the same phase fixes it, since they then share an identical float:

```suggestion
            if container.eigenvalue_from_phase(min(phase_bin, phase_bin_count - phase_bin) / phase_bin_count) < energy_lower_bound:
```

That drops the asymmetric count to 0/10180 and returns `[(0, 2), (7, 8)]` for the case above, with `max|E_sym - E|` about 4e-15 across bins, so it doesn't move any other answer. It does bake the walk's evenness into generic code though, so if you'd rather keep this law-agnostic, symmetrizing the accepted set afterwards (or asserting the symmetry before returning) would work too — the important thing is that something enforces it, since the Q# correctness now depends on it.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
```diff
@@ -0,0 +1,212 @@
...
+
+        Returns:
+            The circuit to use as the ``good_state_oracle`` of ``AmplitudeAmplification``.
+
+        Raises:
+            TypeError: If the nested ``qpe_circuit_builder`` is not a standard (QFT-based) one,
+                the only kind whose circuit can be undone.
+            ValueError: If its ``num_bits`` is not positive, if ``energy_lower_bound`` is unset
+                or not finite, or if it names no proper subspace of the phase register.
+            RuntimeError: If the phase estimation does not carry a Q# operation.
```
RushiGong, line 151: Can you double check the error docstring here? The docstring lists `TypeError`, `ValueError` and `RuntimeError`, but a bad `qpe_circuit_builder` ref escapes as `KeyError` from `_create_nested` (e.g. `KeyError: "Algorithm 'qdk_missing' not found"`). Either add it to the list or let it surface as one of the documented types.

RushiGong (changes requested):

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Fixed in 6b2017f6 by evaluating mirrored bins through the same canonical phase only for QuantumWalkContainer, leaving generic and Trotter phase laws unchanged. I also added an exact-boundary regression test.

python/src/qdk_chemistry/algorithms/amplitude_amplification/qpe_subspace.py
YingrongChen: Documented the KeyError raised when the nested qpe_circuit_builder reference is not registered.

RushiGong (dismissed):

RushiGong (approved):

PR #632, opened 2026-08-11
Update Plugin Architecture

nabbelbabbel:
- Add static `data_type_name()` APIs to canonical native and Python data-class loaders.
- Expose native static type names through pybind.
- Add a data-class registry supporting registration, discovery, lookup, and duplicate-name validation.
- Add `QdkChemistryPlugin` and `PluginRegistrar` for registering algorithm factories, algorithms, and data classes.
- Discover installed plugins through the `qdk_chemistry.plugins` entry-point group.
- Isolate plugin loading failures so one invalid plugin does not prevent others from loading.
- Add coverage for public data-class discovery, canonical loader names, duplicate registration, API-version validation, and entry-point failures.

Depends on #631


cpp/include/qdk/chemistry/utils/string_utils.hpp
```diff
@@ -40,32 +40,4 @@ inline std::string to_snake_case(const char* input) {
...
- * @code
- * class ConfigurationSet : public DataClass {
- *   std::string get_data_type_name() const override {
- *     return DATACLASS_TO_SNAKE_CASE(ConfigurationSet);
- *   }
- *   // Returns "configuration_set"
- * };
- * @endcode
- */
-#define DATACLASS_TO_SNAKE_CASE(ClassName)                \
```
wavefunction91, line 64: This drops `DATACLASS_TO_SNAKE_CASE` and hand-writes ~20 literals in its
place. Could we generalize it to templates instead of retiring it?

Hand-written identifiers can't be checked, and #631 is already paying for
that — it derives a suffix from `Rank` and `is_complex_scalar_v`, which
discriminates the scalar only by complex-ness, so a `float` specialization
would collide with `double`. `SymmetryBlockedSparseMap<Rank, Scalar>`
(symmetry_blocked_sparse_map.hpp:153) and `SymmetryBlockedScalar<Scalar>`
(symmetry_blocked_scalar.hpp:125) are templates too and still return
constants — the second bound specialization of either makes discovery throw
at import and takes unrelated `get_dataclass_type()` lookups with it.

A shape that works, if it's useful. Call sites:

```cpp
DATACLASS_TO_SNAKE_CASE(Ansatz)                                        // "ansatz"
DATACLASS_TO_SNAKE_CASE(SymmetryBlockedTensor, NTTPTag<Rank>, Scalar)  // "symmetry_blocked_tensor_2_complex"
```

The class-name token has to reach the machinery as a template argument, which
takes a structural literal type — public array member, and a deduction guide:

```cpp
template <std::size_t N>
struct FixedString {
  char data[N]{};
  constexpr FixedString() = default;
  constexpr FixedString(const char (&s)[N]) { for (std::size_t i = 0; i < N; ++i) data[i] = s[i]; }
  static constexpr std::size_t size() { return N - 1; }   // N counts the NUL
  constexpr const char* c_str() const { return data; }
  constexpr std::string_view view() const { return {data, N - 1}; }
};
template <std::size_t N>
FixedString(const char (&)[N]) -> FixedString<N>;

// PascalCase -> snake_case. Sized, then filled: the length is a template
// argument of the result type, so it has to be computed first.
constexpr std::size_t snake_case_size(std::string_view s);
template <FixedString Token>
inline constexpr auto snake_case_v = /* FixedString<snake_case_size(Token.view()) + 1> */;
```

`WireTag<T>` is the one extension point. The primary template `static_assert`s,
so a type with no spelling is a compile error instead of a collision at import.
Tag first in the registration macro, so the type may contain commas:

```cpp
template <class T>
struct WireTag {
  static_assert(detail::always_false<T>,
                "no wire spelling registered; add QDK_REGISTER_WIRE_TAG(\"tag\", YourType)");
  static constexpr FixedString value{"?"};   // so the assert above is the only diagnostic
};
template <class T>
inline constexpr auto wire_tag_v = WireTag<T>::value;

#define QDK_REGISTER_WIRE_TAG(Tag, ...)                                     \
  namespace qdk::chemistry::utils {                                         \
  template <> struct WireTag<__VA_ARGS__> {                                 \
    static constexpr ::qdk::chemistry::utils::FixedString value{Tag};       \
  }; }                                                                      \
  static_assert(true, "")

QDK_REGISTER_WIRE_TAG("real",      double);
QDK_REGISTER_WIRE_TAG("real32",    float);
QDK_REGISTER_WIRE_TAG("complex",   std::complex<double>);
QDK_REGISTER_WIRE_TAG("complex32", std::complex<float>);
QDK_REGISTER_WIRE_TAG("uint",      std::size_t);
```

There is no mixed type/non-type parameter pack, so non-type arguments are
lifted into types and the whole list travels as one `class...`. That is what
lets the macro forward `__VA_ARGS__` wholesale and never count it:

```cpp
template <auto V>
struct NTTPTag {};

template <auto V>
struct WireTag<NTTPTag<V>> {
  static constexpr auto value = detail::decimal<V>();   // 2 -> "2", -17 -> "m17"
};
```

Assembly and the macro:

```cpp
namespace detail {
template <FixedString Token, class... Params>
inline constexpr std::size_t wire_name_size =
    snake_case_v<Token>.size() + ((1 + wire_tag_v<Params>.size()) + ... + 0);
}

// Namespace scope, not a static constexpr local: that is C++23. And keyed on
// the token -- key it only on the parameters and every non-template call site
// shares one cache, so ConfigurationSet comes back as "ansatz".
template <FixedString Token, class... Params>
inline constexpr FixedString<detail::wire_name_size<Token, Params...> + 1> wire_name_v = [] {
  FixedString<detail::wire_name_size<Token, Params...> + 1> out{};
  std::size_t k = 0;
  auto append = [&](std::string_view t) { for (char c : t) out.data[k++] = c; };
  append(snake_case_v<Token>.view());
  ((append("_"), append(wire_tag_v<Params>.view())), ...);
  out.data[k] = '\0';
  return out;
}();

#define DATACLASS_TO_SNAKE_CASE(ClassName, ...)                                          \
  (::qdk::chemistry::utils::detail::verify_params<ClassName __VA_OPT__(, ) __VA_ARGS__>(), \
   ::qdk::chemistry::utils::wire_name_v<::qdk::chemistry::utils::FixedString(#ClassName)  \
                                        __VA_OPT__(, ) __VA_ARGS__>.c_str())
```

That gives a `const char*` into static storage — no allocation, still a
constant expression.

`verify_params` is the part worth having. Inside the class body the
injected-class-name is the current instantiation, so the macro passes it as a
type, and the arguments are recovered from the type and compared against what
was declared:

```cpp
template <class...> struct ParamList {};

template <class T> struct RecoveredParams { using type = ParamList<>; static constexpr bool recovered = false; };

template <template <class...> class T, class... A>
struct RecoveredParams<T<A...>>    { using type = ParamList<A...>;             static constexpr bool recovered = true; };

template <template <auto, class...> class T, auto V, class... A>
struct RecoveredParams<T<V, A...>> { using type = ParamList<NTTPTag<V>, A...>; static constexpr bool recovered = true; };

template <class Enclosing, class... Declared>
constexpr void verify_params() {
  using R = RecoveredParams<Enclosing>;
  if constexpr (R::recovered && renderable_list_v<ParamList<Declared...>>) {
    static_assert(params_match_v<ParamList<Declared...>, typename R::type>, "...");
  }
}
```

Match on the rendered tag, not on type identity, so `NTTPTag<2>` satisfies a
`std::size_t` parameter. An omitted list is then the same check with an empty
left-hand side and needs no separate diagnosis. Rejected at compile time:

```
DATACLASS_TO_SNAKE_CASE(SymmetryBlockedTensor)                        // list omitted
DATACLASS_TO_SNAKE_CASE(SymmetryBlockedTensor, Scalar, NTTPTag<Rank>) // wrong order
DATACLASS_TO_SNAKE_CASE(SymmetryBlockedTensor, Scalar)                // wrong count
DATACLASS_TO_SNAKE_CASE(SymmetryBlockedTensor, NTTPTag<3>, Scalar)    // wrong value
long double with no QDK_REGISTER_WIRE_TAG                             // no wire spelling
```

so adding or reordering a template parameter breaks the build at the macro
rather than silently changing a wire name.

Where it stops: a partial specialization has to state the kind of every
parameter, so those two patterns cover `<class...>` and `<auto, class...>` —
every templated data class here — but not an interleaved shape like
`<class, auto>`, which is rendered but unchecked. Keep the patterns
non-overlapping; adding `<auto...>` makes a one-non-type-parameter template
match two and become ambiguous. `__VA_OPT__` needs `/Zc:preprocessor` on MSVC.

Tag spellings extend `_scalar_tag()` (symmetry_blocked_scalar.hpp:371-380),
which is what `j["scalar"]` writes: `double` -> `real` and
`complex<double>` -> `complex` are unchanged, and the 32-bit variants get
distinct spellings instead of sharing one. `to_snake_case()` stays as it is.

Happy to hand over the header and its tests.

If this lands, #631 can close — its names fall out of it, and so do the two
siblings it doesn't cover.

python/src/qdk_chemistry/plugins/base.py
```diff
@@ -0,0 +1,79 @@
...
+    def register_algorithm(
+        self,
+        factory: Callable[[], Any],
+        *,
+        data_classes: Iterable[Type[Any]] = (),  # noqa: UP006
+    ) -> None:
+        """Register an algorithm factory and its DataClass input or output types.
+
+        Raises:
+            DuplicateRegistrationError: If the algorithm name or a DataClass wire type is already registered.
```
wavefunction91, line 42: The docstrings here and on line 26 promise `DuplicateRegistrationError`.
None of the three paths raises it:

```
register_algorithm, duplicate name    -> nothing raised
register_algorithm_factory, duplicate -> ValueError (registry.py:644)
C++ factory path                      -> std::runtime_error
                                         (algorithm.hpp:307-330)
```

Worse than the wrong docstring: after the first case,
`create("term_grouper", "commuting")` returns the newly registered class,
so a plugin silently shadows a built-in.

`DuplicateRegistrationError` subclasses `ValueError`, so
`except DuplicateRegistrationError` catches none of them.

#608 documents this API as the way to register a custom backend
(remote/backends/base.py:139-163), so a plugin author is who finds out.

Should `register_algorithm` reject a duplicate name rather than replace it?

python/src/qdk_chemistry/plugins/base.py
```diff
@@ -0,0 +1,79 @@
...
+
+        """
+        from qdk_chemistry.algorithms import register  # noqa: PLC0415
+
+        register(factory)
```
wavefunction91, line 50: `register(factory)` commits the algorithm before the `data_classes` loop
below runs. A failure in that loop is caught and warned by `_load_plugins`
(plugins/__init__.py:77-82) with no rollback, so a plugin reported as failed
leaves its algorithm registered.

Since `register_instance` overwrites silently, that residue can also shadow
a healthy implementation registered later.

Validate the data classes before committing the algorithm, or unwind on
failure?

python/src/qdk_chemistry/plugins/__init__.py
```diff
@@ -1,80 +1,85 @@
...
+
+import warnings
+
+from qdk_chemistry._core import DuplicateRegistrationError
+from qdk_chemistry.plugins.base import ChemistryPlugin, PluginRegistrar, QdkChemistryPlugin
+
+DuplicateRegistrationError.__module__ = __name__
+
+
+def _load_plugins() -> None:
```
wavefunction91, line 62: Is the MCP work the consumer for this? Nothing uses the group yet — no
entry-points declaration here, in #636 or #608, and in #608 it appears only
in prose and a doctest (remote/backends/__init__.py:11, base.py:139-163).
If MCP is what consumes it, that answers it.

One thing either way: `api_version` is compared with `!=` on line 72, so a
bump rejects newer plugins as well as older ones. If the field is there to
let the API evolve, that inverts it.

wavefunction91 (changes requested):

python/src/qdk_chemistry/plugins/__init__.py
nabbelbabbel: No, nothing uses it because we don't publish plugins as separate packages. But, given the 5 individual registration mechanisms this is a single surface unification that should make things easier to use for others.

python/src/qdk_chemistry/plugins/base.py
nabbelbabbel: This was just missing code that got lost in transit when dismembering #608

python/src/qdk_chemistry/plugins/base.py
nabbelbabbel: Fixed

python/src/qdk_chemistry/plugins/__init__.py
nabbelbabbel: Added full round trip test.

python/src/qdk_chemistry/data/registry.py
```diff
@@ -0,0 +1,152 @@
...
+            seen.add(id(dataclass_type))
+            if (
+                dataclass_type is not _PythonDataClass
+                and dataclass_type not in excluded_types
+                and dataclass_type.__module__.startswith("qdk_chemistry.")
+                and declares_data_type_name(dataclass_type)
+            ):
+                _register_dataclass(dataclass_type)
+            stack.extend(dataclass_type.__subclasses__())
+        _DISCOVERY_COMPLETE = True
```
wavefunction91: _DISCOVERY_COMPLETE is set even when the scan ran with excluded_types. Plugins load at import, before anything else touches the registry, so a plugin that fails after validation latches discovery shut with the excluded types never registered — and those can be core loaders.

A plugin declaring data_classes=[Structure] whose algorithm registration fails:

```
1. third-party plugin failed to load: RuntimeError   (_load_plugins warns)
2. cache write succeeded: ['c7fa6171bf37a03c.structure.h5']
3. cache read of that exact entry -> None
```

FolderCache.get_data returns None when _resolve_dataclass_type misses (folder.py:116), so writes keep succeeding and reads silently miss for the rest of the process. In a control run get_dataclass_type("structure") returns _core.data.Structure.

Would only setting the latch when excluded_types is empty be enough?

python/src/qdk_chemistry/algorithms/base.py
```diff
@@ -375,15 +376,26 @@ def register_instance(self, generator: Callable[[], Algorithm]) -> None:
...
+        Raises:
+            DuplicateRegistrationError: If an algorithm name or alias is already registered in this factory.
 
         Examples:
             >>> factory = ScfSolverFactory()
             >>> factory.register_instance(lambda: MyCustomScf())
 
         """
-        self._registry[generator().name()] = generator
+        aliases = generator().aliases()
```
wavefunction91: register_instance writes an entry per alias, but unregister_instance (line 414) pops only the key it's given — so unregister-then-re-register, the usual way to swap an implementation, now leaves the old one live:

```
registered:                      ['my_estimator', 'mine']
create('mine')                -> v1
unregister('my_estimator')       ['mine']            <- alias survives
register(v2)                  -> DuplicateRegistrationError: 'mine'
create('mine')                -> v1                  <- still the old one
```

The new test's finally block already calls registry.unregister twice to work around this. Aliases also now appear in available(), which previously listed only the primary name.

algorithm.hpp:352-380 has the same register-all/erase-one shape — fix both sides or neither?

nabbelbabbel: Yea, I dropped that though it now requires additional checks that aliases on plugins don't clash either. Otherwise, there is still a factual override and load position dependent path. It does make for a cleaner set of top-level lists though.

python/src/qdk_chemistry/data/registry.py
nabbelbabbel: Yes it should be, I added tests that should fire if it was an issue

wavefunction91 (approved):

PR #634, opened 2026-08-11
Add memory-aware build skill

nabbelbabbel:
Document the 8 GB per compile job requirement and reduce CMake parallelism on memory-constrained machines. Track repository skills under .github.


.github/skills/memory-aware-build/SKILL.md
```diff
@@ -0,0 +1,38 @@
+---
+name: memory-aware-build
+description: Compile QDK Chemistry with memory-aware parallelism. Use this skill whenever building the C++ code, Python package, or dependencies, especially on machines with limited RAM or after an out-of-memory compiler failure.
+---
+
+# Memory-aware builds
+
+QDK Chemistry compilation can require up to 5 GB of RAM per concurrent compile
```
wavefunction91: I like this skill, but I believe the last time I profiled this, one of the TUs is 8GB. Might want to err on the side of caution.

nabbelbabbel: Sure.

mmoerchen (approved):
LGTM

lorisercole (commented):
https://github.com/microsoft/qdk-chemistry/blob/314ba87a06c772727f32428ab4edb45d15889ef1/.github/skills/memory-aware-build/SKILL.md?plain=1#L21-L27
@nabbelbabbel @mmoerchen There is a mistake: 32 / 8 = 4 jobs.

Note that in CI/CD for Windows we use 3.5 or 4 GB per job. The number of jobs is obtained by dividing the _total_ memory by this number. The available RAM (however it is estimated) will be much less than that...
Never had problems in the pipelines, but they might be big runners. On a busy laptop the situation might be more difficult...

PR #635, opened 2026-08-11
`EffectiveHamiltonianConstructor`: don't require `p_indices` ⊆ reference active space

eimrek:
The `EffectiveHamiltonianConstructor` docs and `_validate_inputs` required `P ⊆ W_ref ⊆ W_H`.

That doesn't work for the SW-PT2 implementation, where the target space `P` is independent of the reference wavefunction's active space and is typically a superset of it — e.g. folding onto 10 P-space orbitals from a CAS(2,2) reference.

The check becomes `P ⊆ W_H`, with `W_ref ⊆ W_H` still enforced. This is a pure relaxation: `P ⊆ W_ref` implies `P ⊆ W_H`, so nothing previously valid changes behavior.


cpp/include/qdk/chemistry/algorithms/effective_hamiltonian.hpp
agamshayit-ms: change "nested-space" wording here to reflect the changes

python/src/pybind11/algorithms/effective_hamiltonian.cpp
agamshayit-ms: change "nested-space" wording as well

agamshayit-ms (changes requested):

agamshayit-ms (approved):

PR #636, opened 2026-08-11
Add file-based algorithm serialization

nabbelbabbel:
- add manifest-based serialization for algorithm inputs and outputs
- serialize QDK Chemistry data classes through stable wire type names and HDF5 files
- support nested primitives, containers, algorithm references, and shared-cache entries
- add focused serialization and file-transfer tests

Depends on: #632


python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+    "serialize_outputs",
+]
+
+
+def _atomic_write_json(path: Path, value: Any) -> None:
+    """Write JSON to *path* atomically via a sibling temporary file."""
+    file_descriptor, temporary_path = tempfile.mkstemp(dir=path.parent, suffix=".tmp")
+    try:
+        with os.fdopen(file_descriptor, "w", encoding="utf-8") as file:
+            file_descriptor = -1
```
mmoerchen: Nit: Could we use `NamedTemporaryFile(delete=False)` here? Setting the descriptor to -1 correctly tracks ownership after `os.fdopen()`, but it is subtle. A named temporary file would make ownership and cleanup explicit while preserving the atomic `os.replace()` behavior.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+                and cache.has_data(content_hash)
+            ):
+                return {
+                    "type": "cached",
+                    "dataclass_type": type_name,
+                    "content_hash": content_hash,
+                }
+
+            ext = cls._get_dataclass_extension(type_name)
+            safe_name = name.replace("/", "_").replace("\\", "_").replace("..", "_")
```
mmoerchen: Are these 3 checks sufficient to fully sanitize the file name, or are there additional edge cases we should handle?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+                    directory, item["value"], cache=cache
+                )
+                for item in entry["entries"]
+            }
+
+        raise TypeError(f"Unknown type tag: {type_tag}")
+
+
+def get_serialized_file_names(entry: dict[str, Any]) -> list[str]:
+    """Return every file referenced by a serialized manifest entry."""
```
mmoerchen: Could we expand the docstring to explain the arguments and the structure/meaning of the returned list?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+            json.dump(value, file, indent=2)
+        os.replace(temporary_path, path)
+    except BaseException:
+        if file_descriptor >= 0:
+            os.close(file_descriptor)
+        Path(temporary_path).unlink(missing_ok=True)
+        raise
+
+
+def _jsonable_settings_value(value: Any) -> Any:
```
mmoerchen: How are empty vectors of doubles handled? Could this trigger a `SettingTypeMismatch`?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+        manifest["kwargs"][key] = entry
+
+    files_created = [
+        directory / filename
+        for entries in (manifest["settings"].values(), manifest["args"], manifest["kwargs"].values())
+        for entry in entries
+        for filename in get_serialized_file_names(entry)
+    ]
+
+    manifest_path = directory / "manifest.json"
```
mmoerchen: Should "manifest.json" be a constant instead of being hardcoded?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+    _atomic_write_json(manifest_path, manifest)
+    files_created.insert(0, manifest_path)
+
+    return files_created
+
+
+def deserialize_inputs(directory: str | Path, *, cache: Any = None) -> dict:
```
mmoerchen: Is there a reason cache is not typed as `CacheBackend`?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+    """Deserialize algorithm inputs from a directory.
+
+    Args:
+        directory: Directory containing the input files.
```
mmoerchen: TODO: Is there a way to document "*"?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+            return {"type": "float", "value": value}
+
+        if isinstance(value, str):
+            return {"type": "str", "value": value}
+
+        # Handle lists
+        if isinstance(value, list):
+            items = []
+            for i, item in enumerate(value):
+                items.append(cls.serialize_value(directory, f"{name}_item_{i}", item))
```
mmoerchen: Is there a reason why the cache is not used when serializing `list`, `tuple`, and `dict` objects?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+                    "type": "cached",
+                    "dataclass_type": type_name,
+                    "content_hash": content_hash,
+                }
+
+            ext = cls._get_dataclass_extension(type_name)
+            safe_name = name.replace("/", "_").replace("\\", "_").replace("..", "_")
+            filename = f"{safe_name}{ext}"
+            filepath = directory / filename
+            value.to_hdf5_file(str(filepath))
```
mmoerchen: Only `manifest.json` is written atomically. 
The `.h5` files are written in place here, and the manifest goes last. If a re-run into an existing directory fails half way through, we might be left with the old manifest next to newly overwritten `.h5` files. 
Also, `get_input_files` appears to glob all `*.h5`, so stale files from a previous run get uploaded even if the current manifest doesn't reference them.

python/tests/test_serialization.py
```diff
@@ -0,0 +1,247 @@
...
+    assert (tmp_path / entry["file"]).exists()
+    restored = FileSerializer.deserialize_value(tmp_path, entry)
+    assert isinstance(restored, Orbitals)
+    np.testing.assert_array_equal(restored.get_coefficients(), sample_orbitals.get_coefficients())
+
+
+def test_nested_dataclass_filenames_are_unique(tmp_path, h2_structure):
+    helium = Structure(["He"], np.array([[1.0, 0.0, 0.0]]))
+
+    entry = FileSerializer.serialize_value(tmp_path, "structures", {"/": h2_structure, "_": helium})
```
mmoerchen, line 197: Can we add a test for the filename sanitization? It looks like this path is only reachable through direct serialize_value calls. A case with a name like "../escape" that verifies the file remains inside the target directory would be useful.

python/tests/test_serialization.py
```diff
@@ -0,0 +1,247 @@
...
+    for path in files:
+        shutil.copy2(path, destination / path.name)
+
+    energy, orbitals = deserialize_outputs(destination)
+    assert energy == -75.5
+    assert all(isinstance(item, Orbitals) for item in orbitals)
+
+
+class SharedCache:
+    is_shared = True
```
mmoerchen: Do we also want a test for `is_shared = False`? The fake currently hardcodes `is_shared = True`, so the non-shared path isn't exercised.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+                    f"Cache miss for content_hash={entry['content_hash']!r} (type={entry.get('dataclass_type', '?')})"
+                )
+            return data
+
+        if type_tag == "dataclass":
+            dataclass_type_name = entry["dataclass_type"]
+            dataclass_type = cls._get_dataclass_type(dataclass_type_name)
+            if dataclass_type is None:
+                raise TypeError(f"Unknown DataClass type: {dataclass_type_name}")
+            filepath = directory / entry["file"]
```
mmoerchen: `entry["file"]` comes directly from the manifest and can point outside the job directory (e.g. `../../tmp/x.structure.h5` or an absolute path). Would it make sense to validate that the resolved path stays within the job directory?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,593 @@
...
+        The deserialized result (tuple or single value).
+
+    """
+    directory = Path(directory)
+    manifest_path = directory / "manifest.json"
+
+    with open(manifest_path) as f:
+        manifest = json.load(f)
+
+    results = [FileSerializer.deserialize_value(directory, entry) for entry in manifest["results"]]
```
mmoerchen: Related to the cache discussion above: there is no cache parameter here either, unlike `deserialize_inputs`.

mmoerchen (changes requested):

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Added a fix and a tests

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Yes, it recurses, nested objects are unrolled into their primitives (DataClass etc) and then the cache is called.

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: See above

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: The cache thing was not fully included though so I opted to unify this across the entire serialization layer such that cashes can be use ahead of this serializations on all calls.

python/tests/test_serialization.py
nabbelbabbel: Added that branch

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Fixed

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Fixed that

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Updated to hashes for names to escape any and all edge cases

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,800 @@
...
+
+    @classmethod
+    def is_dataclass(cls, value: Any) -> bool:
+        """Check if a value is a QDK Chemistry DataClass."""
+        return isinstance(value, CoreDataClass)
+
+    @classmethod
+    def is_cacheable(cls, value: Any) -> bool:
+        """Check if a value can be stored in a shared cache."""
+        return cls.is_dataclass(value) or isinstance(value, list)
```
mmoerchen: Could `is_cacheable()` match the value graph supported by the cache backend?

Right now some values that work with the manifest serializer (e.g. lists containing dictionaries `[{"x": 1"}]`) are considered cacheable but cannot actually be stored in the cache.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,800 @@
...
+                            item,
+                            cache=cache,
+                            content_hash=_item_content_hash(item) if cls.is_cacheable(item) else None,
+                            seed_cache=seed_cache,
+                        ),
+                    }
+                )
+            return {"type": "dict", "entries": entries}
+
+        raise TypeError(f"Cannot serialize object of type {type(value).__name__}")
```
mmoerchen, line 438: Could this serializer support NumPy arrays and scalars before reaching this fallback?

For example, the qubit Hamiltonian solvers return tuple[float, np.ndarray]. Those results currently reach this TypeError, so they cannot be transported through serialize_outputs().

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,800 @@
...
+
+    Returns:
+        List of all files created (for upload).
+
+    """
+    directory = Path(directory)
+    directory.mkdir(parents=True, exist_ok=True)
+
+    # Build manifest
+    manifest: dict[str, Any] = {
```
mmoerchen, line 617: Would it make sense to add a schema version to the manifest as well, since `Job` and `DataClass` already version their serialized output?

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Unified that

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Added a version to it

python/src/qdk_chemistry/remote/serialization.py
nabbelbabbel: Good catch, added that.

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,850 @@
...
+
+    Returns:
+        Deserialized inputs containing algorithm metadata, settings, arguments, and cache metadata.
+
+    """
+    directory = Path(directory)
+    manifest_path = directory / _MANIFEST_FILENAME
+
+    with open(manifest_path) as f:
+        manifest = json.load(f)
```
mmoerchen: Could we validate the manifest version number after loading it?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,850 @@
...
+
+    Returns:
+        The deserialized result (tuple or single value).
+
+    """
+    directory = Path(directory)
+    manifest_path = directory / _MANIFEST_FILENAME
+
+    with open(manifest_path) as f:
+        manifest = json.load(f)
```
mmoerchen: Could we validate the manifest version number after loading it?

python/src/qdk_chemistry/remote/serialization.py
```diff
@@ -0,0 +1,850 @@
...
+        if path.exists():
+            raise FileExistsError(f"Serialization file path is not a file: {path}")
+        os.replace(temporary_path, path)
+    finally:
+        Path(temporary_path).unlink(missing_ok=True)
+
+
+def _atomic_write_array(path: Path, value: np.ndarray) -> None:
+    """Write a NumPy array atomically unless its content-addressed file already exists."""
+    if value.dtype.hasobject:
```
mmoerchen: Should we reject structured dtypes here as well, e.g. if `value.dtype.hasobject or value.dtype.fields is not None`?

mmoerchen (approved):

PR #638, opened 2026-08-12
Fix conflict between BLAS inner thread pool and the preset thread pool calling GauXC

BoqinZhang:
The conflict will appear when a DFT calculation is done by more than 1 OpenMP thread.
In build_JK function, a OpenMP thread pool is built to calculate the ERI; these threads will call GauXC. GauXC will call multi-threaded BLAS function, which will create its own thread pool conflicting the pre-built one.
Set `openblas_set_num_threads(1)` to resolve the problem.


lorisercole:
Do I understand correctly that we _never_ want GauXC to run multi-threaded BLAS functions here?

(Additional note: currently when `OMP_NUM_THREADS=1`, BLAS could still be multi-threaded if it was built with pthreads backend. It's an uncommon scenario, but with this PR we'll force BLAS to be single-threaded no matter what. If that's what we want, then that's fine.)

BoqinZhang:
@lorisercole Hi Loris, yes, we should NOT make GauXC run its own multi-thread BLAS functions: the work distribution over threads should be done on the upper level, when build_JK function is called. GauXC should just run sequentially.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
```diff
@@ -17,6 +17,8 @@
 #ifdef QDK_CHEMISTRY_ENABLE_GPU
 #include <qdk/chemistry/scf/util/gpu/cuda_helper.h>
 #endif
+#include <qdk/chemistry/scf/util/blas_threads.h>
```
agamshayit-ms, line 20: Could we keep this implementation private and fixed-purpose? Every call site
requests one thread, and the reported issue is specifically OpenBLAS nested
inside GauXC's OpenMP loop. A local guard like this avoids `BlasVendor`,
configurable counts, runtime loading, `_WIN32`, and the BLIS ABI assumption:

```cpp
#if defined(QDK_CHEMISTRY_HAS_OPENBLAS_THREAD_CONTROL)
#include <cblas.h>
#include <omp.h>
#include <mutex>
#endif

namespace {

#if defined(QDK_CHEMISTRY_HAS_OPENBLAS_THREAD_CONTROL)
class ScopedSerialOpenBlas {
 public:
  ScopedSerialOpenBlas() {
    if (omp_get_max_threads() <= 1) return;

    auto& state = get_state();
    std::lock_guard<std::mutex> lock(state.mutex);
    if (state.depth++ == 0) {
      state.saved_threads = openblas_get_num_threads();
      openblas_set_num_threads(1);
    }
    active_ = true;
  }

  ~ScopedSerialOpenBlas() {
    if (!active_) return;

    auto& state = get_state();
    std::lock_guard<std::mutex> lock(state.mutex);
    if (--state.depth == 0)
      openblas_set_num_threads(state.saved_threads);
  }

  ScopedSerialOpenBlas(const ScopedSerialOpenBlas&) = delete;
  ScopedSerialOpenBlas& operator=(const ScopedSerialOpenBlas&) = delete;

 private:
  struct State {
    std::mutex mutex;
    unsigned depth = 0;
    int saved_threads = 1;
  };

  static State& get_state() {
    static State state;
    return state;
  }

  bool active_ = false;
};
#else
struct ScopedSerialOpenBlas {};
#endif

}  // namespace
```

The four call sites would become:

```cpp
ScopedSerialOpenBlas blas_thread_guard;
```

This should also delete:

- `include/qdk/chemistry/scf/util/blas_threads.h`
- `src/util/blas_threads.cpp`
- the three `BlasThreadsTest` tests and their includes in `tests/util_tests.cpp`

Could the implementation tests be replaced with behavior-level SNK coverage in
`scf_test.cpp`: start with OpenBLAS using more than one thread, execute SNK with
multiple OpenMP threads, verify the result, and verify the original BLAS count is
restored?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,85 @@ install(
...
+  message(FATAL_ERROR
+    "Invalid QDK_CHEMISTRY_BLAS_THREAD_API='${QDK_CHEMISTRY_BLAS_THREAD_API}'. "
+    "Expected AUTO, NONE or one of: ${_qdk_blas_backends}.")
+endif()
+
+if(_qdk_blas_thread_api)
+  # Consumed by config.h.in as #cmakedefine QDK_CHEMISTRY_BLAS_LINKS_<backend>.
+  set(QDK_CHEMISTRY_BLAS_LINKS_${_qdk_blas_thread_api} TRUE)
+  # Already linked transitively through GauXC; name it for the symbols we call.
+  target_link_libraries(chemistry PRIVATE ${_qdk_blas_link_libs})
```
agamshayit-ms, line 237: Could the installed package preserve the selected BLAS provider? The static
archive references `openblas_*`, but exported `BLAS::BLAS` may resolve to MKL or
BLIS, causing unresolved symbols. This requires updating
`cpp/cmake/qdkConfig.cmake.in` or avoiding provider-specific symbols in the
installed archive.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,85 @@ install(
   COMPONENT qdk_chemistry_resources
 )
 
+###############################################################################
```
agamshayit-ms: Could the five-backend runtime-discovery block be replaced by a configure-time
check for the reproduced OpenBLAS/OpenMP case?

```cmake
if(QDK_ENABLE_OPENMP AND BLAS_VENDOR STREQUAL "OpenBLAS")
  include(CheckCXXSourceCompiles)
  include(CMakePushCheckState)

  cmake_push_check_state(RESET)
  set(CMAKE_REQUIRED_LIBRARIES BLAS::BLAS)
  set(CMAKE_REQUIRED_INCLUDES ${BLAS_INCLUDE_DIRS})
  check_cxx_source_compiles([=[
    #include <cblas.h>
    int main() {
      const int threads = openblas_get_num_threads();
      openblas_set_num_threads(threads);
      return 0;
    }
  ]=] QDK_CHEMISTRY_HAS_OPENBLAS_THREAD_CONTROL)
  cmake_pop_check_state()

  if(QDK_CHEMISTRY_HAS_OPENBLAS_THREAD_CONTROL)
    target_compile_definitions(
      chemistry PRIVATE QDK_CHEMISTRY_HAS_OPENBLAS_THREAD_CONTROL)
    target_link_libraries(chemistry PRIVATE BLAS::BLAS)
  else()
    message(WARNING
      "OpenBLAS thread control unavailable; nested threading remains enabled")
  endif()
endif()
```

This also removes `util/blas_threads.cpp` from `target_sources` and removes
`QDK_CHEMISTRY_BLAS_VENDOR` plus all `QDK_CHEMISTRY_BLAS_LINKS_*` definitions
from `config.h.in`.

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
BoqinZhang: Supporting different types of BLAS vendor (OpenBLAS, MKL, BLIS, ...) is required, so the implementation of ScopedBlasThreads must be a long code, and setting its implementation inside `gauxc_impl.cpp` is not a good idea. Also, I think it is better to test the BLAS thread control individually. If we test it together with DFT calculation, it may be hard to judge whether the problem is due to DFT or the thread control code.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
BoqinZhang: I am afraid we cannot do it like this. It is required that the code support a series of BLAS vendors, not only OpenBLAS.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
BoqinZhang: I modified `qdkConfig.cmake.in` to check if the BLAS vendor supports BLAS symbols designated by qdk-chemistry

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,128 @@ install(
...
+  set(QDK_CHEMISTRY_BLAS_THREAD_SYMBOL "" CACHE INTERNAL
+      "BLAS thread-control symbol bound into the library")
+  set(_qdk_blas_thread_api "none (runtime lookup only)")
+endif()
+
+message(STATUS "QDK Chemistry BLAS vendor: '${QDK_CHEMISTRY_BLAS_VENDOR}', "
+               "thread-control API: ${_qdk_blas_thread_api}")
+
+# dlsym() for the runtime fallback; empty on Windows and macOS.
+target_link_libraries(chemistry PRIVATE ${CMAKE_DL_LIBS})
```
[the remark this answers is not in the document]
BoqinZhang: This comment is invalid

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/util/blas_threads.h
```diff
@@ -0,0 +1,90 @@
...
+ * guard to start changes it, and only the last one to finish restores it. This
+ * keeps concurrent/recursive use safe.
+ *
+ * If the BLAS backend in use exposes no thread-control API, the guard is a
+ * no-op (a one-time warning is logged).
+ */
+class ScopedBlasThreads {
+ public:
+  /// @param num_threads Thread count to pin BLAS to while the guard is alive.
+  explicit ScopedBlasThreads(int num_threads = 1);
```
agamshayit-ms: This accepts arbitrary counts, but overlapping guards honor only the first one.
Since every caller passes 1, please remove the parameter and always pin to one
thread.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
```diff
@@ -294,6 +296,10 @@ void GAUXC::free_device_buffer_async_(cudaStream_t stream) {
 void GAUXC::build_XC(const double* D, double* XC, double* xc_energy) {
   QDK_LOG_TRACE_ENTERING();
 
+  // GauXC parallelizes over grid batches with OpenMP and calls BLAS from
+  // each thread; pin BLAS to one thread to avoid nested threading.
+  util::ScopedBlasThreads blas_thread_guard(1);
```
agamshayit-ms: The unit tests do not cover the failing GauXC path. Add one multi-threaded SNK
regression and verify the original BLAS thread count is restored.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/util/blas_threads.cpp
```diff
@@ -0,0 +1,263 @@
...
+    // Prefer the API bound at link time: that BLAS is by construction the one
+    // our calls go to, and it is the only option when BLAS is linked
+    // statically (no module exports the symbols to look up at runtime).
+    if (BlasThreadApi linked = linked_blas_thread_api(); linked.valid()) {
+      QDK_LOGGER().debug("Using {} BLAS thread control (bound at link time)",
+                         to_string(linked.vendor));
+      return linked;
+    }
+
+    // Fall back to whichever backend is loaded in this process.
```
agamshayit-ms: If configure-time binding fails, this scan may change a BLAS loaded by another
Python package instead of the one GauXC uses. In that case, leave the BLAS thread
count unchanged and log that automatic thread control is unavailable.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,128 @@ install(
...
+# the count as dim_t, 32- or 64-bit depending on its build; the 64-bit form is
+# safe for both, as a 32-bit callee reads the low half of the argument register.
+set(_qdk_blas_backends FLEXIBLAS OPENBLAS INTELMKL BLIS NVPL)
+set(_qdk_blas_api_FLEXIBLAS
+    flexiblas_set_num_threads flexiblas_get_num_threads int FlexiBLAS "FlexiBLAS")
+set(_qdk_blas_api_OPENBLAS
+    openblas_set_num_threads openblas_get_num_threads int OpenBLAS "OpenBLAS")
+set(_qdk_blas_api_INTELMKL
+    MKL_Set_Num_Threads MKL_Get_Max_Threads int IntelMKL "Intel MKL")
+set(_qdk_blas_api_BLIS
```
agamshayit-ms, line 130: BLIS dim_t may be 32-bit, so declaring it as int64_t is ABI-incompatible. Use
BLIS's actual dim_t and disable runtime BLIS probing when its ABI is unknown.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,128 @@ install(
...
+  # Already linked transitively through GauXC; name it for the symbols we call.
+  target_link_libraries(chemistry PRIVATE ${_qdk_blas_link_libs})
+  # Link-time binding leaves vendor-specific symbols unresolved in the installed
+  # archive while the exported dependency is the vendor-agnostic BLAS::BLAS, so
+  # cmake/qdkConfig.cmake.in steers and verifies the consumer's BLAS search.
+  # Cache variables because this is a subdirectory scope.
+  set(QDK_CHEMISTRY_REQUIRED_BLAS_VENDOR "${QDK_CHEMISTRY_BLAS_VENDOR}"
+      CACHE INTERNAL
+      "BLAS vendor whose thread-control symbols are bound into the library")
+  set(QDK_CHEMISTRY_BLAS_THREAD_SYMBOL "${QDK_CHEMISTRY_BLAS_LINKED_SET_FN}"
```
agamshayit-ms: qdk-chemistry calls both `QDK_CHEMISTRY_BLAS_LINKED_SET_FN` and
`QDK_CHEMISTRY_BLAS_LINKED_GET_FN`, but this exports only the setter for
consumer validation. Please export and check both function names.

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/util/blas_threads.h
BoqinZhang: The `num_thread` input variable is removed

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/util/blas_threads.cpp
BoqinZhang: The fall-back probe is removed

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
BoqinZhang: Setting it as int64_t is a fallback safe choice. It can be a wider datatype, but cannot be a short one. As for checking `dim_t`, it looks like a complex task and may fail.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
BoqinZhang: No need to make it now: the `num_thread` variable has been removed

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
BoqinZhang: `GET_FN` is added in export

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
```diff
@@ -101,6 +102,127 @@ install(
...
+# the count as dim_t, 32- or 64-bit depending on its build; the 64-bit form is
+# safe for both, as a 32-bit callee reads the low half of the argument register.
+set(_qdk_blas_backends FLEXIBLAS OPENBLAS INTELMKL BLIS NVPL)
+set(_qdk_blas_api_FLEXIBLAS
+    flexiblas_set_num_threads flexiblas_get_num_threads int FlexiBLAS "FlexiBLAS")
+set(_qdk_blas_api_OPENBLAS
+    openblas_set_num_threads openblas_get_num_threads int OpenBLAS "OpenBLAS")
+set(_qdk_blas_api_INTELMKL
+    MKL_Set_Num_Threads MKL_Get_Max_Threads int IntelMKL "Intel MKL")
+set(_qdk_blas_api_BLIS
```
agamshayit-ms, line 130: A wider declaration is not ABI-safe. GCC LTO rejects both functions when
BLIS uses 32-bit dim_t and warns they may be misoptimized. Please use BLIS's
actual dim_t from its header, or disable BLIS support when that type is
unavailable.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
```diff
@@ -396,6 +406,10 @@ void GAUXC::get_gradients(const double* D, double* dXC) {
 void GAUXC::build_snK(const double* D, double* K) {
   QDK_LOG_TRACE_ENTERING();
 
+  // GauXC parallelizes over grid batches with OpenMP and calls BLAS from
+  // each thread; pin BLAS to one thread to avoid nested threading.
+  util::ScopedBlasThreads blas_thread_guard;
```
agamshayit-ms, line 408: Removing the count parameter does not cover this call site. The existing
"SNK" test logs fock_alg=TradJK because the non-addons fixture ignores k_eri.
Please make that branch honor k_eri == "SNK" and verify the BLAS count before
and after the calculation.

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/exc/GauXC/gauxc_impl.cpp
BoqinZhang: Resolved the problem of missing `k_eri == "SNK"` in scf testsuit

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/CMakeLists.txt
BoqinZhang: Added code to find and apply BLIS's dim_t

cpp/src/qdk/chemistry/algorithms/microsoft/scf/tests/scf_test.h
```diff
@@ -162,8 +163,11 @@ class SCFTest
 #else
     cfg.eri.method =
         eri == "CPU" ? ERIMethod::Libint2Direct : ERIMethod::Incore;
-    cfg.k_eri.method =
-        eri == "CPU" ? ERIMethod::Libint2Direct : ERIMethod::Incore;
+    // SnK is not gated by the HGP/RYS/LibintX options, so it is available here
+    // too; without this the "SNK" rows silently fall back to TradJK.
+    cfg.k_eri.method = k_eri == "SNK" ? ERIMethod::SnK
+                       : eri == "CPU" ? ERIMethod::Libint2Direct
+                                      : ERIMethod::Incore;
```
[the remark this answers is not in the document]
agamshayit-ms: This comment is reasonable. Should we just use `k_eri == "CPU"`? If so, that also implicates the pre-existing code in line 158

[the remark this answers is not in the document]
BoqinZhang: Fixed

agamshayit-ms (approved):

PR #649, opened 2026-08-18
Add Discovery Plugin

nabbelbabbel:


python/src/qdk_chemistry/plugins/discovery/backend.py
```diff
@@ -0,0 +1,584 @@
...
+        try:
+            manifest_path = f"{output_dir}/manifest.json"
+            manifest_local = resolved_dir / "manifest.json"
+            self.download(manifest_path, manifest_local)
+            output_paths.append(manifest_path)
+            manifest = json.loads(manifest_local.read_text(encoding="utf-8"))
+            for entry in manifest.get("results", []):
+                if filename := entry.get("file"):
+                    remote_path = f"{output_dir}/{filename}"
+                    self.download(remote_path, resolved_dir / filename)
```
wavefunction91: Could we validate each manifest filename before downloading it, using the same containment rule that #608 applies during deserialization? `resolved_dir / filename` discards `resolved_dir` when `filename` is absolute, so a worker or Blob writer can make `fetch()` replace any file writable by the client before `deserialize_outputs()` gets a chance to reject the manifest. The validated target should be proven to remain under `resolved_dir` before `download()` creates its parent or writes it.

python/src/qdk_chemistry/plugins/discovery/backend.py
```diff
@@ -0,0 +1,584 @@
...
+        results: list[Any] = []
+        for entry in job.output_hashes:
+            if "value" in entry:
+                results.append(entry["value"])
+                continue
+            value = cache.get_data(entry["hash"])
+            if value is None:
+                raise LookupError(f"Shared cache is missing output {entry['hash']} for run {run_hash}")
+            results.append(value)
+        return results[0] if len(results) == 1 else tuple(results)
```
wavefunction91: Could `_fetch_cached_result()` use the persisted `job.output_is_tuple` from #608 instead of inferring result shape from the number of entries? The current return converts a singleton tuple such as `(42,)` to `42`, while `not job.output_hashes` rejects an empty tuple. Both are supported result shapes that #608 preserves explicitly; the two-item tuple in this PR’s test does not exercise either case.

python/src/qdk_chemistry/remote/worker.py
```diff
@@ -0,0 +1,168 @@
...
+    cache, run_hash = _load_remote_cache(input_path)
+    result = _get_cached_result(cache, run_hash)
+
+    if result is _CACHE_MISS:
+        inputs = deserialize_inputs(input_path, cache=cache)
+        algorithm = create_algorithm(inputs["algorithm_type"], inputs["algorithm_name"])
+        for key, value in inputs["settings"].items():
+            algorithm.settings().set(key, value)
+        result = algorithm.run(*inputs["args"], **inputs["kwargs"])
+        _store_cached_result(cache, run_hash, inputs, result, required=cache_transport)
```
wavefunction91: How should cache transport handle output shapes that remote serialization accepts but `CacheBackend` cannot store? For example, with `(1, (2, 3))`, `collect_content_hashes()` treats the nested tuple as a non-primitive result, `FolderCache.put_data()` rejects that tuple, and `required=True` turns a successfully computed result into a failed job. Cache transport needs to preserve every result shape accepted by the remote output contract.

python/src/qdk_chemistry/plugins/discovery/backend.py
```diff
@@ -0,0 +1,584 @@
...
+                    remote_path = f"{output_dir}/{filename}"
+                    self.download(remote_path, resolved_dir / filename)
+                    output_paths.append(remote_path)
+            result = deserialize_outputs(resolved_dir)
+            succeeded = True
+            return result
+        finally:
+            if own_tmp:
+                shutil.rmtree(resolved_dir, ignore_errors=True)
+            if self.cleanup and (succeeded or self.cleanup_on_failure):
```
wavefunction91: Could this use #608’s caller-driven cleanup lifecycle rather than deleting artifacts inside `fetch()`? #608 retains artifacts until `Job.cleanup()` or `Job.fetch(cleanup=True)`, both of which call `cleanup_job()`, but `DiscoveryBackend` does not override that method. Those APIs therefore raise `NotImplementedError`, while a deletion error here can replace an otherwise successful fetch result. Moving idempotent deletion into `cleanup_job()` and removing the `cleanup`/`cleanup_on_failure` settings would match the accepted backend contract.

python/src/qdk_chemistry/plugins/discovery/backend.py
```diff
@@ -0,0 +1,584 @@
...
+            payload,
+            cache=payload.get("remote_cache_backend"),
+        )
+        input_dir = f"{self.remote_workdir}/job_{job_id}/input"
+        output_dir = f"{self.remote_workdir}/job_{job_id}/output"
+        input_paths: list[str] = []
+        try:
+            for local_file in input_files:
+                remote_path = f"{input_dir}/{local_file.name}"
+                self.upload(local_file, remote_path)
```
wavefunction91, line 362: Could the staging path roll back everything it has created when upload or submission fails before a `Job` exists? If `serialize_inputs()` raises, the outer `mkdtemp()` directory remains; if a later upload or `begin_run()` fails, already-uploaded input blobs remain and no `backend_state` is returned for #608’s `cleanup_job()` path. These pre-handle failures need to remove the local directory and any known uploaded paths before re-raising.

python/src/qdk_chemistry/remote/worker.py
wavefunction91: Could cache transport skip `serialize_outputs()` after the required cache write succeeds? In this mode `_submit()` creates no output mount and `fetch()` reads only `_fetch_cached_result()`, so `output_path` is never consumed. Serializing here writes a full second copy into container `/tmp` after the cache write, repeats output hashing and I/O, and can turn a successful computation into a failed job if temporary storage fills.

python/src/qdk_chemistry/plugins/discovery/backend.py
```diff
@@ -0,0 +1,584 @@
...
+        polling = OperationIdPolling()
+        self._client.tools.begin_run(
+            project_name=str(self.project_name),
+            tool_id=str(self.tool_id),
+            node_pool_ids=[str(self.node_pool_id)],
+            command=command,
+            inline_files=inline_files,
+            input_data=input_data,
+            output_data=output_data,
+            infra_overrides=InfraOverrides(
```
wavefunction91: What behavior is `use_entire_node` intended to select here? The option is read, persisted, and defaults to `True`, but `_submit()` builds identical `InfraOverrides` whether it is true or false. Could we either wire it to the actual resource request or remove the public setting until that behavior is supported?

python/src/qdk_chemistry/plugins/discovery/backend.py
nabbelbabbel: Yes, updated this post merge of #608 into this via main.

python/src/qdk_chemistry/plugins/discovery/backend.py
nabbelbabbel: Updated this post merge of https://github.com/microsoft/qdk-chemistry/pull/608 into this via main.

python/src/qdk_chemistry/remote/worker.py
nabbelbabbel: Fixed

python/src/qdk_chemistry/plugins/discovery/backend.py
nabbelbabbel: Removed it

python/src/qdk_chemistry/remote/cache/folder.py
```diff
@@ -163,7 +163,7 @@ def put_data(self, content_hash: str, data: Any, *, shared_only: bool = False) -
         self._validate_key(content_hash, "content_hash")
         if not is_cacheable(data):
             raise TypeError(f"FolderCache does not support caching values of type {type(data).__name__}")
-        if isinstance(data, list):
+        if isinstance(data, list | tuple):
```
wavefunction91, line 166: Could tuple support avoid deleting shared content-addressed children when one parent entry is removed? For example, two tuple entries containing the same NumPy array both reference the same `<hash>.ndarray.npy`; deleting the first parent recursively unlinks that file, leaving the second parent with `has_data(...) == True` but `get_data(...) is None`. Since this line newly admits root tuples, their deletion path needs to preserve children still referenced by other cache entries.

nabbelbabbel: Fixed that.

wavefunction91 (approved):

PR #653, opened 2026-08-18
Preserve atom-specific ECP assignments

### LINKED ISSUE #651: Mixed ECP treatment is lost for atoms of the same element
**Describe the bug**

A `BasisSet` can represent different basis/ECP treatment for individual atoms, but converting it to a PySCF molecule does not preserve that distinction when two atoms have the same element symbol.

For an Ag2 system where the first Ag uses `def2-svp` (28 core electrons replaced by an ECP) and the second Ag uses the all-electron `ano-rcc` basis, the QDK basis correctly reports `[28, 0]`. After conversion, PySCF applies the ECP to both Ag atoms. This silently removes 28 extra electrons.

**To Reproduce**

```python
import numpy as np

from qdk_chemistry.data import BasisSet, Structure
from qdk_chemistry.plugins.pyscf.conversion import basis_to_pyscf_mol

structure = Structure(
    ["Ag", "Ag"],
    np.array(
        [
            [0.0, 0.0, 0.0],
            [5.0, 0.0, 0.0],
        ]
    ),
)

basis = BasisSet.from_index_map(
    {0: "def2-svp", 1: "ano-rcc"},
    structure,
)
mol = basis_to_pyscf_mol(basis)

print("QDK ECP electrons:", basis.get_ecp_electrons())
print("PySCF ECP keys:", list(mol.ecp))
print("PySCF atom charges:", mol.atom_charges().tolist())
print("PySCF electron count:", mol.nelectron)
```

Observed output:

```text
QDK ECP electrons: [28, 0]
PySCF ECP keys: ['Ag']
PySCF atom charges: [19, 19]
PySCF electron count: 38
```

**Expected behavior**

The ECP should apply only to atom 0. Atom 1 should remain all-electron:

```text
PySCF atom charges: [19, 47]
PySCF electron count: 66
```

The conversion should preserve the per-atom ECP assignment already present in the QDK `BasisSet`.

**System**

- OS: WSL2 Linux, x86_64
- qdk-chemistry: 2.1.0 / current main
- PySCF: 2.12.1

**Additional context**

Homogeneous cases, where every atom of a given element uses the same ECP treatment, do not expose the problem. The failure requires multiple atoms of the same element with different ECP or all-electron treatment.

ConradJohnston:
## Summary

- key PySCF ECP data by unique atom labels so same-element centers can use different ECP treatment
- preserve per-atom ECP electron counts through Microsoft SCF serialization and round trips
- retain the existing per-element representation for backward compatibility when treatment is uniform
- add regression coverage for mixed ECP and all-electron Ag centers

Closes #651

## Testing

- `cpp/build/tests/test_scf --gtest_brief=1` (42 passed)
- mixed Ag ECP Python reproduction (66 electrons, atom charges `[19, 47]`)
- `ruff check --ignore RUF043 src/qdk_chemistry/plugins/pyscf/conversion.py tests/test_pyscf_plugin.py`
- `ruff format --check src/qdk_chemistry/plugins/pyscf/conversion.py tests/test_pyscf_plugin.py`


cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/core/basis_set.cpp
```diff
@@ -568,15 +579,30 @@ std::shared_ptr<BasisSet> BasisSet::from_serialized_json(
...
     int atomic_num = bs->mol->atomic_nums[i];
-    if (bs->element_ecp_electrons.count(atomic_num)) {
-      int ecp_electrons = bs->element_ecp_electrons[atomic_num];
-      bs->mol->atomic_charges[i] = atomic_num - ecp_electrons;
-      bs->n_ecp_electrons += ecp_electrons;
-    }
+    int ecp_electrons = bs->atom_ecp_electrons[i];
+    bs->mol->atomic_charges[i] = atomic_num - ecp_electrons;
+    bs->n_ecp_electrons += ecp_electrons;
   }
```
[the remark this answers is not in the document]
ConradJohnston: Addressed in 64da47a5. Deserialization now rejects per-atom ECP counts outside [0, atomic number] before computing unsigned atomic charges. Added regression coverage for negative and oversized values.

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
```diff
@@ -236,14 +237,10 @@ qdk::chemistry::data::BasisSet convert_basis_set_to_qdk(
...
-      int atomic_num = basis_set.mol->atomic_nums[i];
-      auto it = basis_set.element_ecp_electrons.find(atomic_num);
-      if (it != basis_set.element_ecp_electrons.end()) {
-        qdk_ecp_electrons[i] = static_cast<size_t>(it->second);
-      }
+    std::vector<size_t> qdk_ecp_electrons;
+    qdk_ecp_electrons.reserve(basis_set.atom_ecp_electrons.size());
+    for (int ecp_electrons : basis_set.atom_ecp_electrons) {
+      qdk_ecp_electrons.push_back(static_cast<size_t>(ecp_electrons));
     }
```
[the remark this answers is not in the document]
ConradJohnston: Addressed in 64da47a5. Conversion now validates vector size and each ECP count before casting to size_t, including bounds against the corresponding atomic number. Added regression coverage for both negative and oversized values.

mmoerchen (changes requested):
It seems like having both a per-element and a per-atom map for the number of ECP electrons could lead to invalid states, especially when two atoms of the same element have a different number of ECP electrons.

Would it make sense to remove `element_ecp_electrons` altogether? 
Are per-element values even required anywhere, or could we just use the new per-atom map?

ConradJohnston:
Thanks, Max. I agreed that keeping both representations as mutable state could allow them to diverge. Addressed in 03bc1e42: `atom_ecp_electrons` is now the sole live representation, the ECP constructor accepts per-atom values directly, and ASAHF equality/hash and conversion paths no longer depend on an element map. The legacy `element_ecp_electrons` JSON field is derived on write only when treatment is uniform and is read into a temporary map solely for backward compatibility with older serialized data. Added coverage for legacy JSON fallback and mixed same-element serialization. The affected utility and SCF suites passed before rebasing onto the latest `main`, and all pre-commit hooks pass after the rebase.

python/src/qdk_chemistry/plugins/pyscf/conversion.py
```diff
@@ -169,6 +169,7 @@ def basis_to_pyscf_mol(basis: BasisSet, charge: int = 0, multiplicity: int = 1)
     if basis.has_ecp_shells() and basis.has_ecp_electrons():
         # Build PySCF ECP structure from QDK ECP shells
         ecp_dict = {}
+        ecp_data_by_atom = {}
```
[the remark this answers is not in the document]
ConradJohnston: Addressed in aed81fd8. The count-only fallback now builds element or atom-label keyed ECP entries from the per-atom electron counts, so mixed same-element metadata such as [28, 0] applies the named ECP only to Ag1. Added a regression covering charges [19, 47] and 66 electrons.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/core/basis_set.h
```diff
@@ -70,8 +69,8 @@ class BasisSet {
   uint64_t num_atomic_orbitals;  ///< Total number of atomic orbitals
 
   std::vector<Shell> ecp_shells;  ///< Effective core potential shells
-  std::unordered_map<int, int>
-      element_ecp_electrons;  ///< Map from atomic number to ECP electron number
+  std::vector<int>
+      atom_ecp_electrons;   ///< Number of ECP electrons for each atom
```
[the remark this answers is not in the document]
ConradJohnston: This SCF BasisSet is an internal implementation type rather than an installed public C++ API. Its header is under cpp/src/.../scf/include, while installation only exports cpp/include/**/*.hpp, and it is not exposed through pybind11. Retaining duplicate mutable state would reintroduce the inconsistency this PR fixes, so no compatibility member is added.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/core/basis_set.cpp
```diff
@@ -415,10 +434,27 @@ nlohmann::ordered_json BasisSet::to_json() const {
     json_ecp_shells.push_back(sh.to_json(true /*is_ecp*/));
   }
 
+  std::map<int, int> element_ecp_electrons;
```
mmoerchen: Is there a reason why we still serialize `element_ecp_electrons`? Could we not directly serialize `atom_ecp_electrons`?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/core/basis_set.cpp
```diff
@@ -547,36 +584,53 @@ std::shared_ptr<BasisSet> BasisSet::from_serialized_json(
...
-    // Backward compatibility: support old key name
-    element_ecp_electrons_json = json["elem2ecpcore"].get<std::vector<int>>();
-  }
+  if (json.contains("atom_ecp_electrons")) {
+    bs->atom_ecp_electrons = json["atom_ecp_electrons"].get<std::vector<int>>();
+    if (bs->atom_ecp_electrons.size() != bs->mol->n_atoms) {
+      throw std::runtime_error(
+          "atom_ecp_electrons must contain one entry per atom.");
+    }
+  } else {
```
mmoerchen, line 565: Do we want to keep this backwards compatibility? If so, could we add a comment that this is legacy behavior and log a warning when it is encountered?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/asahf.cpp
```diff
@@ -140,24 +139,19 @@ std::shared_ptr<BasisSet> make_atom_basis_set(size_t index,
...
 
-  // get element from mol to get the ecp electrons from map
   auto atomic_number = mol->atomic_nums[0];
-  if (basis_set.element_ecp_electrons.find(atomic_number) !=
-      basis_set.element_ecp_electrons.end()) {
-    ecp_electrons[atomic_number] =
-        basis_set.element_ecp_electrons.at(atomic_number);
-    total_ecp_electrons = ecp_electrons[atomic_number];
+  if (index < basis_set.atom_ecp_electrons.size()) {
+    total_ecp_electrons = basis_set.atom_ecp_electrons[index];
```
mmoerchen: Is `total_ecp_electrons` still used anywhere, or can it be removed?

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/core/basis_set.h
```diff
@@ -70,8 +69,8 @@ class BasisSet {
...
   int n_ecp_electrons = 0;  ///< Total number of core electrons replaced by ECPs
```
mmoerchen: Do we need this as well, or can we just use `atom_ecp_electrons` as the source of truth?

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
```diff
@@ -254,19 +255,28 @@ qdk::chemistry::data::BasisSet convert_basis_set_to_qdk(
   }
 
   // Handle ECP (Effective Core Potential) information if present
-  if (basis_set.n_ecp_electrons != 0 || !basis_set.ecp_shells.empty() ||
-      !basis_set.element_ecp_electrons.empty()) {
+  if (basis_set.n_ecp_electrons != 0 || !basis_set.ecp_shells.empty()) {
```
mmoerchen: Why are we not using the per atom vector as source of truth here?

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
```diff
@@ -360,24 +370,34 @@ nlohmann::ordered_json convert_to_json(
...
   }
 
-  // Build element_ecp_electrons map from ecp_electrons vector
+  // Retain the element map for backward compatibility when the ECP treatment
+  // is uniform for every atom of an element.
   auto& structure = basis_set.get_structure();
   auto nuclear_charges = structure->get_nuclear_charges();
   auto ecp_electrons = basis_set.get_ecp_electrons();
 
   std::map<int, int> element_ecp_electrons;
```
mmoerchen: Is there a reason why we still serialize `element_ecp_electrons`? Could we not directly serialize `atom_ecp_electrons`?

python/src/qdk_chemistry/plugins/pyscf/conversion.py
```diff
@@ -213,16 +214,48 @@ def basis_to_pyscf_mol(basis: BasisSet, charge: int = 0, multiplicity: int = 1)
...
             mol.ecp = ecp_dict
             # Store ECP name as attribute for roundtrip conversion
             mol.qdk_ecp_name = basis.get_ecp_name()
     elif basis.has_ecp_electrons():
         # Fallback: only ECP name available, no shells
-        mol.ecp = basis.get_ecp_name()
+        ecp_name = basis.get_ecp_name()
+        ecp_electrons = basis.get_ecp_electrons()
+        ecp_dict = {}
+        named_ecp_indices_by_element: dict[str, list[int]] = {}
```
mmoerchen: Nit: Could we extract the element-vs-atom key selection into a helper? It looks like the same logic is implemented in both the explicit-shell and named-ECP paths.

mmoerchen (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/core/basis_set.cpp
ConradJohnston: Addressed in ac6b74ee: `BasisSet::to_json` now writes only `atom_ecp_electrons`; the element-level field is retained only as a legacy read fallback.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/core/basis_set.cpp
ConradJohnston: Addressed in ac6b74ee: the legacy fallback is retained, documented as legacy behavior, and logs a warning when either legacy key is read.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/src/scf_algorithm/asahf.cpp
ConradJohnston: Addressed in ac6b74ee: the `total_ecp_electrons` local was removed in favor of the atom-specific value read with bounds checking.

cpp/src/qdk/chemistry/algorithms/microsoft/scf/include/qdk/chemistry/scf/core/basis_set.h
ConradJohnston: Addressed in ac6b74ee: the cached `n_ecp_electrons` member was removed; totals are now derived from `atom_ecp_electrons`.

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
ConradJohnston: Addressed in ac6b74ee: conversion validates the per-atom vector and derives ECP presence and totals directly from it.

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
ConradJohnston: Addressed in ac6b74ee: `convert_to_json` now writes only `atom_ecp_electrons`; legacy element data is read-only compatibility.

python/src/qdk_chemistry/plugins/pyscf/conversion.py
ConradJohnston: Addressed in ac6b74ee: element-vs-atom ECP key selection is shared in one helper, with additional consistency validation.

mmoerchen (approved):
lgtm

PR #659, opened 2026-08-19
Fix ECP-adjusted nuclear charges and repulsion energies

### LINKED ISSUE #658: ECP calculations use incorrect nuclear repulsion
**Description**
For ECP systems, native SCF correctly uses ECP-adjusted nuclear charges, but Hamiltonian construction used the physical charges stored in `Structure`.

For AgH/def2-SVP, this gave:

Physical Ag charge: 47
ECP-adjusted Ag charge: 47−28=19
Incorrect Hamiltonian core term: 47/R
Correct effective nuclear repulsion: 19/R

Consequently, the SCF total energy disagreed with the mean-field Ansatz energy.

**Expected behavior**
`Structure` should retain physical atomic charges. ECP-adjusted charges and nuclear repulsion should be derived from `BasisSet` and used by Hamiltonian construction and ECP-dependent electron counting. Non-ECP behavior should remain unchanged.

Originally identified while working on #583 by @vbarandun

mmoerchen:
Fixes #658


cpp/src/qdk/chemistry/data/basis_set.cpp
```diff
@@ -1230,6 +1265,17 @@ bool BasisSet::_is_consistent_with_structure() const {
     return true;  // No structure to validate against
   }
 
+  if (_ecp_electrons.size() != _structure->get_num_atoms()) {
+    return false;
+  }
+  const auto& nuclear_charges = _structure->get_nuclear_charges();
+  for (size_t atom_idx = 0; atom_idx < _ecp_electrons.size(); ++atom_idx) {
+    if (static_cast<double>(_ecp_electrons[atom_idx]) >
+        nuclear_charges(static_cast<Eigen::Index>(atom_idx))) {
```
ConradJohnston, line 1274: This validation is reasonable, but the existing `make_ecp_basis()` fixture in `cpp/tests/test_cube_generator.cpp` still creates hydrogen (`Z = 1`) with 2 or 28 ECP electrons. The constructor now throws before three `CubeGeneratorTest` cases run, which is why the macOS, Linux, and clang-cl test jobs are failing. Please keep this check and update the fixture to use a single high-Z atom, such as Ag, while retaining the synthetic valence and projector shells.

ConradJohnston (changes requested):
The ECP charge and energy changes look sound overall, but this currently breaks the C++ test suite. Please update the cube-generator ECP fixture described in the inline comment while keeping the new production validation.

python/tests/test_pyscf_plugin.py
```diff
@@ -2931,8 +2931,29 @@ def test_agh_def2svp_roundtrip(self):
         # Create QDK Structure
         structure = Structure(symbols=["Ag", "H"], coordinates=mol1.atom_coords())
 
+        # Run native QDK SCF once and compare its ECP-adjusted quantities with PySCF
```
agamshayit-ms: Why does this need to go through the PySCF test?

cpp/src/qdk/chemistry/algorithms/microsoft/scf.cpp
```diff
@@ -108,16 +108,15 @@ ScfCalculationResult ScfSolver::_run_with_options(
...
 
-  // Compute sum of nuclear charges
-  int nuclear_charge = 0;
-  for (auto i = 0; i < structure->get_num_atoms(); ++i) {
-    nuclear_charge += structure->get_atom_nuclear_charge(i);
-  }
+  const auto effective_nuclear_charges =
+      qdk_raw_basis_set->get_effective_nuclear_charges();
+  const int effective_nuclear_charge =
+      static_cast<int>(effective_nuclear_charges.sum());
```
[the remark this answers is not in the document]
mmoerchen: Valid, because `structure.hpp` is supporting fractional charges

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
```diff
@@ -269,11 +269,10 @@ std::shared_ptr<qcs::BasisSet> convert_basis_set_from_qdk(
...
 
-  // remove number of ecp electrons from atomic charges
-  auto ecp_electrons = qdk_basis_set.get_ecp_electrons();
+  const auto effective_charges = qdk_basis_set.get_effective_nuclear_charges();
   for (size_t i = 0; i < mol->n_atoms; ++i) {
-    int n_core_electrons = static_cast<int>(ecp_electrons[i]);
-    mol->atomic_charges[i] = mol->atomic_nums[i] - n_core_electrons;
+    mol->atomic_charges[i] =
+        static_cast<uint64_t>(effective_charges(static_cast<Eigen::Index>(i)));
   }
```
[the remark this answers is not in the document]
mmoerchen: Valid, because structure.hpp is supporting fractional charges

cpp/tests/test_scf.cpp
```diff
@@ -765,10 +767,34 @@ TEST_F(ScfTest, AgHDef2SvpWithEcp) {
...
+
+  auto hamiltonian_constructor = HamiltonianConstructorFactory::create();
+  auto hamiltonian = hamiltonian_constructor->run(orbitals);
+  EXPECT_NEAR(hamiltonian->get_core_energy(),
+              expected_effective_nuclear_repulsion,
+              testing::scf_energy_tolerance);
+
+  Ansatz mean_field_ansatz(hamiltonian, wfn);
+  EXPECT_NEAR(mean_field_ansatz.calculate_energy(), energy,
               testing::scf_energy_tolerance);
```
ConradJohnston, line 798: Could we also keep an absolute total-energy reference for this AgH calculation? The PySCF comparison is useful, but PySCF versions can drift or contain their own bugs. A fixed reference would give us an independent baseline alongside the analytical checks and cross-code comparison. Please pin the correctly defined total energy for this exact geometry and basis, with an appropriate tolerance and a note describing where the reference value came from.

cpp/src/qdk/chemistry/algorithms/microsoft/scf.cpp
```diff
@@ -108,16 +109,17 @@ ScfCalculationResult ScfSolver::_run_with_options(
...
 
-  // Compute sum of nuclear charges
-  int nuclear_charge = 0;
-  for (auto i = 0; i < structure->get_num_atoms(); ++i) {
-    nuclear_charge += structure->get_atom_nuclear_charge(i);
-  }
+  const auto effective_nuclear_charges =
+      utils::microsoft::to_integral_nuclear_charges(
+          qdk_raw_basis_set->get_effective_nuclear_charges());
+  const int effective_nuclear_charge = static_cast<int>(
```
ConradJohnston: Could we rename this scalar to `total_effective_nuclear_charge`? It is the sum over the per-atom `effective_nuclear_charges` vector, so including `total` would make that distinction clear. This is only a local naming suggestion; the public `get_effective_nuclear_charges()` name makes sense for both all-electron and ECP calculations.

ConradJohnston:
Non-blocking: The ECP cases here all remove an even number of electrons. Could we add a small test using an ECP with an odd core-electron count, if one is readily available in the supported basis data? That would exercise the electron-count and multiplicity paths in a way the 28-electron Ag ECP does not. If no practical example exists, I do not think this should block the PR.

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
```diff
@@ -269,11 +292,10 @@ std::shared_ptr<qcs::BasisSet> convert_basis_set_from_qdk(
...
-  auto ecp_electrons = qdk_basis_set.get_ecp_electrons();
+  const auto effective_charges = to_integral_nuclear_charges(
+      qdk_basis_set.get_effective_nuclear_charges());
   for (size_t i = 0; i < mol->n_atoms; ++i) {
-    int n_core_electrons = static_cast<int>(ecp_electrons[i]);
-    mol->atomic_charges[i] = mol->atomic_nums[i] - n_core_electrons;
+    mol->atomic_charges[i] = effective_charges[i];
   }
 
   auto basis_json = convert_to_json(qdk_basis_set);
```
agamshayit-ms, line 301: `data::BasisSet` stores ECP counts per atom, but
`qdk::chemistry::scf::BasisSet` stores one count per element in
`element_ecp_electrons`. For two Ag atoms with counts `[28, 0]`,
`from_serialized_json()` therefore overwrites the correct charges `[19, 47]`
with `[19, 19]`. Could this conversion preserve the per-atom counts, or reject
same-element atoms with different counts?

python/src/pybind11/data/basis_set.cpp
```diff
@@ -844,6 +844,30 @@ Get the ECP (Effective Core Potential) electrons vector.
     >>> print(f"ECP electrons per atom: {ecp_electrons}")
 )");
 
+  basis_set.def("get_effective_nuclear_charges",
```
agamshayit-ms, line 847: Could these docstrings mention the `RuntimeError` raised when no `Structure` is
associated? Both methods are available on a public shell-only
`BasisSet(name, shells)`, but require `get_structure()` internally.

cpp/tests/test_scf.cpp
```diff
@@ -765,10 +767,38 @@ TEST_F(ScfTest, AgHDef2SvpWithEcp) {
...
+  EXPECT_NEAR(hamiltonian->get_core_energy(),
+              expected_effective_nuclear_repulsion,
+              testing::scf_energy_tolerance);
+
+  Ansatz mean_field_ansatz(hamiltonian, wfn);
+  EXPECT_NEAR(mean_field_ansatz.calculate_energy(), energy,
+              testing::scf_energy_tolerance);
+  EXPECT_NEAR(mean_field_ansatz.calculate_energy(), -146.62430815169887,
+              testing::scf_energy_tolerance);
+  EXPECT_NEAR(mean_field_ansatz.calculate_energy(), -146.62430815169887,
```
agamshayit-ms: Why the same `EXPECT_NEAR` twice?

agamshayit-ms (changes requested):

cpp/src/qdk/chemistry/algorithms/microsoft/utils.cpp
mmoerchen: This is addressed in #653 and is out of scope for this bug fix.

ConradJohnston: ECPs that remove an odd number of core electrons do exist, but they appear to be limited to specialized 4f-in-core lanthanide pseudopotentials, such as those described by Dolg et al. (*Theoretical Chemistry Accounts* 75, 173–194, 1989, https://doi.org/10.1007/BF00528565). Since covering one here would require a fairly artificial setup to test, I do not think we need to add this test in this PR. The implementation does not assume an even count, so we can add coverage later if one of these ECPs becomes part of our supported basis data. Happy to consider my suggestion resolved.

ConradJohnston (approved):

agamshayit-ms (approved):

Issue #628, opened 2026-08-08
Generate stable circuit image assets from QDK/Chemistry tutorial workflows


sobolevnrm:
## Problem

The ground-state molecular-QPE tutorial renders logical circuits interactively
with `qdk.widgets.Circuit`. Interactive rendering is useful in notebooks, but
Sphinx documentation also benefits from stable in-page circuit images that can
be reviewed, versioned, described with alt text, and read without executing a
notebook.

The current Chapter 5 and Chapter 6 circuit figures were prepared from manually
captured widget output. Manual screenshots are workable, but they are difficult
to reproduce consistently and may include editor or widget-interface artifacts.
They also make future circuit changes harder to detect and regenerate.

## Desired outcome

Investigate and implement, where practical, a reproducible path from the existing
QDK/Chemistry circuit workflows to stable documentation assets such as SVG or PNG.

The workflow should use the same circuit data generated by the tutorial scripts
and notebooks. It should not require manually capturing Visual Studio Code,
Jupyter, browser, or widget UI.

Candidate approaches include:

- exporting SVG or PNG directly from the QDK circuit representation or widget;
- rendering circuit JSON through a supported headless renderer;
- adding a small documentation-maintenance exporter outside the Sphinx build; or
- contributing export support to the owning QDK widget package if the capability
  does not belong in QDK/Chemistry.

## Scope

Evaluate the circuit visuals used by the ground-state molecular-QPE tutorial:

- one- and two-determinant state-preparation logical circuits; and
- the power-one IQPE iteration circuit.

Determine whether the same export path can support other QDK/Chemistry
documentation and examples without introducing tutorial-specific rendering logic.

## Constraints

- Do not capture Visual Studio Code, notebook, browser, or widget control UI.
- Keep generation outside the Sphinx build.
- Use deterministic circuit inputs and stable framing, dimensions, fonts, and
  colors where supported.
- Preserve complete wire and gate labels needed for interpretation.
- Avoid adding a heavyweight browser or graphics dependency to ordinary
  QDK/Chemistry installations.
- Keep interactive notebook rendering; static documentation assets complement it
  rather than replace it.

## Acceptance criteria

- The available export paths and their owning repository/package are documented.
- If feasible in QDK/Chemistry, a reproducible maintenance command generates the
  tutorial's circuit image assets from tested circuit data.
- Repeated generation from unchanged inputs produces byte-identical or
  semantically stable assets.
[...]

PR #670, opened 2026-08-22
Make molecular QPE tutorial figures theme-safe

### LINKED ISSUE #666: Document source settings for the Chapter 2 orbital figures
The following two figures in the Chapter 2 “Describing the molecule” tutorial require their generation settings:

Image

We need the original basis/orbital indices, isocontour values, cube-generation settings, and camera/view settings so they can be reproduced with transparent backgrounds while preserving the existing scientific visuals.

If possible, please add a small generation script or document the settings used to create them.

sobolevnrm:
## Summary

- generate the five Graphviz diagrams and two Matplotlib plots as locally reproducible, accessible SVGs on opaque white canvases
- preserve the four original orbital and circuit screenshots as non-published source assets under `docs/figure_sources/ground_state_qpe`
- regenerate the two Chapter 2 orbital figures as transparent PNGs with lossless color-to-alpha conversion
- restore the original Chapter 5 and 6 circuit screenshots as opaque PNGs on white backgrounds
- document every maintenance command and add focused source-to-output regression tests

## Asset policy

Every published tutorial image has a documented local generation or post-processing path:

- `generate_tutorial_qpe_graphviz.py` creates committed SVGs from the tutorial DOT sources.
- The phase-wrapping and orbital-entropy generators create deterministic vector SVGs from tutorial constants and calculations.
- `generate_tutorial_qpe_screenshot_images.py` creates the four screenshot-derived PNGs from source screenshots retained outside the published Sphinx tree.

All assets with fixed dark foreground colors use a shared opaque white canvas for maximum contrast and simple maintenance. Orbital surfaces retain transparent backgrounds because their colors remain legible against light and dark themes. Generated SVGs include authored titles, descriptions, ARIA labels, and source hashes.

## Circuit screenshots

This PR preserves the existing circuit screenshots as source inputs and deterministically normalizes their white backgrounds. It does **not** generate circuit images from tested circuit data and therefore does not complete the requirements of issue #628. `qdk.widgets.Circuit` does not currently provide a supported Python or command-line exporter; microsoft/qdk#3632 tracks that prerequisite.

## Resolution of #666

The requested historical cube-generation, isocontour, orbital-index, and camera settings cannot be recovered. Inventing settings would misstate the figures' scientific provenance. This PR closes #666 by retaining the original screenshots as explicit source assets and reproducibly generating transparent published copies without changing their appearance when composited on white.

## Validation

- configured pre-commit and pre-push hooks pass
- molecular-QPE regressions plus image tests: 26 passed, 1 expected PySCF visualization skip
- all 12 generated outputs are byte-identical across repeated complete regeneration passes
- clean Sphinx HTML build completes with no tutorial-image warnings
- all seven SVG figures render through standard `` elements with authored alt text and captions

Closes #666


docs/source/_static/diagrams/README.md
```diff
@@ -0,0 +1,66 @@
...
+composited on white. The opaque source assets are preserved in Git commit
+`e128a7a40d7645dee8a0bbf854c10d823b79ee46`. Recover either source with
+`git show`, then run the converter:
+
+```console
+git show e128a7a40d7645dee8a0bbf854c10d823b79ee46:docs/source/_static/diagrams/tutorial_qpe_atomic_basis_functions.png \
+  > /tmp/tutorial_qpe_atomic_basis_functions.png
+python docs/source/_static/diagrams/convert_white_to_alpha.py \
+  /tmp/tutorial_qpe_atomic_basis_functions.png \
+  docs/source/_static/diagrams/tutorial_qpe_atomic_basis_functions.png
```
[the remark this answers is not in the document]
sobolevnrm: Addressed in 74ac9de2. The example now uses two standalone commands with no backslash continuations, so readers can copy each command independently.

docs/source/conf.py
```diff
@@ -154,6 +154,9 @@
     "show-inheritance",  # Show base classes in documentation
 )
 
+# Render Graphviz diagrams as scalable assets with transparent backgrounds.
+graphviz_output_format = "svg"
```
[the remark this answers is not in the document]
sobolevnrm: Addressed in da68fa7b. The comment now states that Sphinx selects SVG output while each DOT source controls its own canvas background.

docs/source/_static/diagrams/README.md
```diff
@@ -0,0 +1,51 @@
+# Ground-state QPE tutorial figures
+
+The ground-state molecular-QPE tutorial uses Graphviz diagrams, generated plots,
+orbital renders, and logical-circuit diagrams. Generation is a documentation
+maintenance task and does not run during a Sphinx build.
```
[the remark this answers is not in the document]
sobolevnrm: Addressed in da68fa7b. The introduction now distinguishes manually regenerated PNG assets from DOT sources rendered by Sphinx/Graphviz during the build.

python/tests/test_docs_tutorial_images.py
```diff
@@ -0,0 +1,165 @@
...
+
+    configuration = ast.parse((REPOSITORY_ROOT / "docs" / "source" / "conf.py").read_text(encoding="utf-8"))
+    output_formats = [
+        node.value.value
+        for node in configuration.body
+        if isinstance(node, ast.Assign)
+        and any(isinstance(target, ast.Name) and target.id == "graphviz_output_format" for target in node.targets)
+        and isinstance(node.value, ast.Constant)
+    ]
+    assert output_formats == ["svg"]
```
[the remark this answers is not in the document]
sobolevnrm: Addressed in da68fa7b. The AST check now accepts both ast.Assign and ast.AnnAssign while still requiring the configured value to be the constant "svg".

docs/source/conf.py
```diff
@@ -154,6 +154,9 @@
     "show-inheritance",  # Show base classes in documentation
 )
 
+# Render Graphviz diagrams as SVG; each DOT source controls its canvas background.
+graphviz_output_format = "svg"
```
ConradJohnston: Switching `graphviz_output_format = "svg"` changes Sphinx output from `` to an `<object>` whose authored `:alt:` text is only fallback content. When the SVG loads successfully, readers instead get the embedded Graphviz document, whose title is just the generic DOT graph name. This removes the detailed accessible names written for six tutorial diagrams and one non-tutorial diagram.

Please retain an `img`-based PNG/SVG path, or customize the Graphviz HTML/SVG output so the object has `role="img"` and the authored description is supplied through `aria-label`, `aria-labelledby`, or SVG `<desc>`. The remaining directives without explicit alt text should also receive descriptions.

ConradJohnston: The global SVG switch also changes font resolution for four DOT files using `fontname="Arial Bold"`. Graphviz emits `font-family="Arial Bold"` without `font-weight`; browsers treat that face name as an unknown family and render regular fallback text. Eleven cluster headings lose their bold hierarchy, and fallback metrics caused three headings to overflow during browser verification.

Please use `fontname="Arial"` with HTML-like `<B>...</B>` cluster labels, which emits `font-weight="bold"`, and add a regression check covering both tutorial and non-tutorial DOT files.

docs/source/_static/diagrams/generate_tutorial_qpe_screenshot_images.py
```diff
@@ -0,0 +1,83 @@
...
+ORBITAL_SCREENSHOTS = (
+    "tutorial_qpe_atomic_basis_functions.png",
+    "tutorial_qpe_example_molecular_orbitals.png",
+)
+CIRCUIT_SCREENSHOTS = (
+    "tutorial_qpe_state_preparation_comparison.png",
+    "tutorial_qpe_power_one_circuit_overview.png",
+)
+
+CIRCUIT_BACKGROUND = 242
```
ConradJohnston: `CIRCUIT_BACKGROUND = 242` changes the circuit canvas from white to `#F2F2F2` without changing the solid `#8F8F8F` wires and outlines. Their contrast drops from 3.234:1 to 2.889:1. The 3:1 threshold comes from [WCAG 2.2 SC 1.4.11, Non-text Contrast](https://www.w3.org/TR/WCAG22/#non-text-contrast), which applies to graphical objects needed to understand content. These opaque images contain the circuit structure readers are explicitly asked to trace, so CSS cannot repair the contrast.

Please use at least `#F8F8F8` for the panel or darken the structural strokes, update the corner-color assertion, and add an explicit contrast check.

docs/source/_static/diagrams/tutorial_qpe_workflow.dot
```diff
@@ -7,7 +7,7 @@ digraph TutorialQpeWorkflow {
     overlap=false;
     splines=spline;
     pad=0.2;
-    bgcolor="#FAFAFA";
+    bgcolor="transparent";
```
ConradJohnston: Changing the canvas from `#FAFAFA` to transparent leaves the `#7B1FA2` flow arrows unchanged. Their contrast is only 2.16:1 on `#111827` and 2.56:1 even on black. [WCAG 2.2 SC 1.4.11, Non-text Contrast](https://www.w3.org/TR/WCAG22/#non-text-contrast) uses 3:1 for graphical parts needed to understand content. The surrounding captions and prose may mitigate a strict conformance failure, but these arrows cross transparent gaps and encode required flow in four tutorial diagrams, so this remains a concrete dark-theme readability regression.

Please retain an opaque light canvas for the affected diagrams or use theme-adaptive, higher-contrast arrows, and validate the rendered output on dark backgrounds.

ConradJohnston (changes requested):

docs/source/conf.py
sobolevnrm: Addressed in 3e29d0fa. The tutorial no longer uses Sphinx Graphviz SVG objects. A local generator now creates committed SVGs with role="img", aria-labelledby, authored <title>/<desc> metadata, and source hashes; standard `figure` directives render them as `` elements with the original alt text and captions.

sobolevnrm: Addressed in 3e29d0fa. All seven affected cluster labels now use `fontname="Arial"` with HTML `<B>` labels. Generated SVGs contain real `font-weight="bold"`, and the regression test rejects `Arial Bold` in tutorial DOT sources.

docs/source/_static/diagrams/generate_tutorial_qpe_screenshot_images.py
sobolevnrm: Addressed in 3e29d0fa. Circuit panels now use #F8F8F8, raising #8F8F8F structure contrast above 3:1. The test computes the WCAG relative-luminance ratio and confirms the structural stroke color is present in both generated circuit PNGs.

docs/source/_static/diagrams/tutorial_qpe_workflow.dot
sobolevnrm: Addressed in 3e29d0fa. The five tutorial DOT canvases are now opaque #FAFAFA. They generate committed SVGs outside Sphinx, preserving SVG scaling while keeping arrow contrast predictable on dark themes. The rendered docs were visually checked and tests require the opaque canvas.

docs/source/_static/diagrams/generate_tutorial_qpe_screenshot_images.py
sobolevnrm: Follow-up in 0886754a: the final canvas policy uses white rather than #F8F8F8. This further increases #8F8F8F circuit-structure contrast while giving all fixed-color opaque assets one clearly documented background value.

docs/source/_static/diagrams/tutorial_qpe_workflow.dot
sobolevnrm: Follow-up in 0886754a: the final Graphviz canvases use opaque white rather than #FAFAFA. Arrow contrast remains predictable and higher than on the earlier transparent dark-theme presentation, while all fixed-color opaque assets now share one background value.

docs/source/_static/diagrams/generate_tutorial_qpe_screenshot_images.py
```diff
@@ -0,0 +1,83 @@
...
+        for index in range(1, num_components + 1)
+        if index in border_components or areas[index] >= MIN_BACKGROUND_COMPONENT_AREA
+    }
+    background = np.isin(components, list(background_components))
+    offset = 255 - CIRCUIT_BACKGROUND
+    rgb[background] = np.maximum(
+        rgb[background].astype(np.int16) - offset,
+        0,
+    ).astype(np.uint8)
+    return Image.fromarray(rgb, mode="RGB")
```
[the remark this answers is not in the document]
sobolevnrm: Addressed in 48f62ae9. Detected background components are now assigned directly to the canonical white value. A synthetic near-white border regression verifies 254 becomes 255 while a non-background center pixel remains unchanged, and both committed circuit PNGs were regenerated.

python/tests/test_docs_tutorial_images.py
```diff
@@ -0,0 +1,277 @@
...
+import importlib.util
+import re
+import sys
+import xml.etree.ElementTree as ET
+from hashlib import sha256
+from pathlib import Path
+
+import numpy as np
+from PIL import Image
+
```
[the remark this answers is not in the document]
sobolevnrm: Pillow is already guaranteed by the declared dependency graph: QDK/Chemistry directly requires matplotlib>=3.10.0, and every supported Matplotlib release declares Pillow as a required (non-optional) dependency; the current 3.11.1 metadata requires pillow>=9. Base, docs, and test installs therefore all include it. Adding the same transitive requirement again would be redundant, so no dependency change is needed.

docs/source/_static/diagrams/generate_tutorial_qpe_screenshot_images.py
```diff
@@ -0,0 +1,79 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from pathlib import Path
+
+import numpy as np
+from PIL import Image
+from scipy import ndimage
+
```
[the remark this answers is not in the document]
sobolevnrm: Same answer as the companion thread: the repository directly requires matplotlib>=3.10.0, whose supported releases require Pillow; current Matplotlib 3.11.1 declares pillow>=9. The documented contributor environments therefore already install Pillow, and duplicating that requirement in an optional group would not change installation behavior.

docs/source/_static/diagrams/generate_tutorial_qpe_orbital_entropy.py
```diff
@@ -8,6 +8,8 @@
 import sys
 from pathlib import Path
 
+import matplotlib as mpl
```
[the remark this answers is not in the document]
sobolevnrm: Won't fix.

python/tests/test_docs_tutorial_images.py
```diff
@@ -0,0 +1,277 @@
...
+        title = root.find(f"{{{SVG_NAMESPACE}}}title")
+        description = root.find(f"{{{SVG_NAMESPACE}}}desc")
+        assert title is not None
+        assert title.text
+        assert description is not None
+        assert description.text == alt_match.group("alt")
+        assert "Arial Bold" not in dot_source
+        svg_text = svg_path.read_text(encoding="utf-8")
+        assert 'fill="#ffffff"' in svg_text
+        assert 'font-weight="bold"' in svg_text
```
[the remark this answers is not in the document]
sobolevnrm: Nope.

docs/source/tutorials/ground_state_molecular_energies_with_qpe/index.rst
```diff
@@ -7,10 +7,11 @@ It is intended for advanced undergraduate and early-stage graduate students who
 
 .. _tutorial-qpe-workflow:
 
-.. graphviz:: /_static/diagrams/tutorial_qpe_workflow.dot
+.. figure:: /_static/diagrams/tutorial_qpe_workflow.svg
```
ConradJohnston, line 10: Seven tutorial figures now reference explicit `.svg` files. HTML works, but `make clean all SPHINX_FORMAT=latexpdf` emits raw SVG paths into the LaTeX output and then fails because standard LaTeX graphics cannot include them. At base, Graphviz generated PDF for LaTeX and the two plots were PNG. The repository documents HTML and Texinfo failure was not established, which limits the scope.

Could we provide PNG/PDF sibling assets and reference `basename.*` so Sphinx selects the appropriate format per builder, or configure and declare a supported SVG conversion extension?

python/tests/test_docs_tutorial_images.py
```diff
@@ -0,0 +1,277 @@
...
+def test_opaque_png_figures_use_white_backgrounds():
+    """Keep every fixed-color PNG on the shared opaque white canvas."""
+    for name in OPAQUE_PNG_FIGURES:
+        with Image.open(DIAGRAMS_DIR / name) as image:
+            rgba = image.convert("RGBA")
+            assert rgba.getchannel("A").getextrema() == (255, 255), name
+            assert rgba.getpixel((0, 0)) == (255, 255, 255, 255), name
+
+
+def test_circuit_png_figures_meet_non_text_contrast():
```
ConradJohnston: The contrast test calculates 3.234:1 from hardcoded white and `#8F8F8F` values. The image only needs one `#8F8F8F` pixel, while the canvas check samples `(0,0)`, so low-contrast semantic structure can still pass. The current assets meet the targeted contrast, but the assertion is not connected to their actual boundaries.

Could we detect real interior/background colors and measure contrast across representative structure/background boundaries with substantial compliant coverage?

python/tests/test_docs_tutorial_images.py
ConradJohnston: The generated-SVG style guards are whole-file substring checks. Unrelated bold node labels satisfy `font-weight="bold"` after cluster headings lose bold, and a white plot marker satisfies `fill: #ffffff` after both canvases turn gray. Regeneration updates source hashes, so wrong-but-fresh output still passes. Current SVGs are correct.

Could we assert the specific cluster-heading elements are bold and the specific figure/axes background patches are white, or compare regenerated normalized output?

python/tests/test_docs_tutorial_images.py
```diff
@@ -0,0 +1,277 @@
...
+    assert references == {name.replace(".dot", ".svg") for name in dot_files}
+    assert references == svg_files
+    assert len(directives) == 5
+    for rst_path, match in directives:
+        svg_name = match.group("source")
+        dot_name = svg_name.replace(".svg", ".dot")
+        dot_path = DIAGRAMS_DIR / dot_name
+        dot_source = dot_path.read_text(encoding="utf-8")
+        assert 'bgcolor="#FFFFFF"' in dot_source
+        if dot_name != "tutorial_qpe_wavefunction_hierarchy.dot":
```
ConradJohnston: The Graphviz test rewrite removed the prior hierarchy checks that enforced the approved spaced `μ`/`HF` subscript forms, rejected `χ[μ]` and `Φ(HF)`, and banned other raw `<SUB>` usage. The hierarchy source is exempt from the generic ban, and regeneration updates its hash, so known typography regressions now pass. The current artifact remains correct.

Could we restore those source assertions and add focused checks that the approved subscript structure survives in the generated SVG?

docs/source/_static/diagrams/generate_tutorial_qpe_phase_grid.py
```diff
@@ -162,14 +173,36 @@ def main() -> None:
...
-            evolution_time
+    with mpl.rc_context(
+        {
+            "svg.fonttype": "path",
+            "svg.hashsalt": "qdk-chemistry-phase-wrapping",
+        }
+    ):
+        figure = generate_phase_wrapping_figure(evolution_time)
+        output_path = Path(__file__).with_name(SVG_FILENAME)
+        figure.savefig(
```
ConradJohnston, line 185: This generator and the entropy generator overwrite the tracked SVG with raw Matplotlib output before validating descriptions. Missing or malformed alt text can therefore raise after the valid SVG has already lost its role, title, description, and source-hash metadata. The failure is loud, Git-recoverable, and CI catches it, so this is maintenance-only.

Could we validate descriptions first, or render in memory and write the destination only after metadata injection succeeds?

docs/source/_static/diagrams/generate_tutorial_qpe_graphviz.py
```diff
@@ -0,0 +1,70 @@
...
+    "tutorial_qpe_wavefunction_hierarchy": "Wavefunction representation hierarchy",
+    "tutorial_qpe_orbital_partition": "Active-space orbital partition",
+    "tutorial_qpe_jordan_wigner_parity": "Jordan-Wigner parity mapping",
+    "tutorial_qpe_iqpe_iteration": "Iterative phase-estimation iteration",
+}
+
+
+def accessible_svg(dot_path: Path, title: str, description: str) -> str:
+    """Render one DOT source and add stable accessibility metadata."""
+    result = subprocess.run(
```
ConradJohnston, line 29: `subprocess.run(..., text=True)` captures Graphviz's UTF-8 stdout without specifying `encoding="utf-8"`. Under a non-UTF default locale this can produce mojibake or fail after earlier SVGs were rewritten, while current tests can accept successfully corrupted labels. Native Windows is outside the documented Linux/macOS/WSL maintenance scope, limiting the impact.

Could we set `encoding="utf-8"` explicitly and regenerate the five SVGs so their generator-source hashes remain current?

ConradJohnston (commented):
Some nitpicks/polish that could be worth adding.

ConradJohnston (dismissed):

docs/source/tutorials/ground_state_molecular_energies_with_qpe/index.rst
sobolevnrm: The published and CI-validated documentation target is HTML. LaTeX/PDF output is neither documented nor tested and requires an undeclared external TeX toolchain, so preserving that builder is outside this PR's scope. Supporting it would require a separate project-level decision and broader validation beyond these tutorial assets.

python/tests/test_docs_tutorial_images.py
sobolevnrm: Thanks, agreed that the original assertion did not prove its stated claim. I explored boundary-based validation against the real image pixels; both assets passed, but the result depended on an arbitrary coverage threshold and antialiasing details. Since the current assets already meet the targeted contrast, I removed the misleading contrast test in 81e19faf rather than expand this documentation PR into image-analysis policy. The existing white-canvas and deterministic source/output checks remain.

docs/source/_static/diagrams/generate_tutorial_qpe_graphviz.py
sobolevnrm: Addressed in 63e0cf93. The Graphviz subprocess now decodes stdout with explicit UTF-8. I regenerated all five SVGs; after normalizing data-source-sha256, they are byte-identical to the prior artifacts, so the only generated changes are the expected source-hash updates.

docs/source/_static/diagrams/generate_tutorial_qpe_phase_grid.py
sobolevnrm: Addressed in e0e2ff9d using the smaller prevalidation option. Both Matplotlib generators now resolve their authored description before computation or any file write. I verified with a forced description-validation failure that both tracked SVGs and the phase-grid table remain byte-identical; normal regeneration changed only the expected SVG source hashes.

python/tests/test_docs_tutorial_images.py
sobolevnrm: Agreed that the whole-file substring checks overstated what they validated. Removed those three weak SVG style assertions in 1eebb98f rather than add element-specific SVG parsing to this documentation PR. The focused DOT background setting, generated-source hashes, accessibility metadata, and no-embedded-raster checks remain; current artifacts are unchanged.

python/tests/test_docs_tutorial_images.py
sobolevnrm: Addressed narrowly in 015d2b7f. Restored source-level guards for the two exact spaced μ/HF subscript forms, rejected the legacy χ[μ]/Φ(HF) spellings, and required exactly two `<SUB>` elements while retaining the ban in every other DOT file. I kept this at the authored-source level rather than add generated-SVG parsing; mutation checks confirmed the known spacing and typography regressions fail the guards.

wavefunction91 (approved):

PR #689, opened 2026-08-28
Add MCP

nabbelbabbel:


python/src/qdk_chemistry/ui/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# Licensed under the MIT License. See LICENSE.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import importlib
+
+__copyright__ = """"""
+
+__version__ = "1.0.0"
+
+__all__ = ["app", "cli"]
```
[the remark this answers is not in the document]
nabbelbabbel: Resolved

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+        return algorithm.run(*args, **kwargs)
+
+    # If remote is a name string and remote_config is provided,
+    # create a pre-configured backend instance so the SDK receives
+    # the full configuration.
+    resolved_remote = remote
+    if isinstance(remote, str) and remote_config:
+        from qdk_chemistry.remote.backends import get_backend  # noqa: PLC0415
+
+        resolved_remote = get_backend(remote, **remote_config)
```
wavefunction91, line 437: `remote_config` is untrusted MCP input, but it is forwarded verbatim to the backend constructor. With the built-in `local` backend, `{"python_path": "/workspace/evil"}` becomes `subprocess.Popen`'s executable, so an executable in the opened workspace can run as the MCP server user; `overwrite=true` forces the execution path even when a cache entry exists. Could this surface exclude executable-selection fields and allowlist only the backend options that are safe for MCP clients to control?

python/src/qdk_chemistry/ui/validation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    """
+    # Convert projects_dir to Path if it's a string
+    if isinstance(projects_dir, str):
+        projects_dir = Path(projects_dir)
+    elif not isinstance(projects_dir, Path):
+        return False, f"Projects dir should be a Path or string but it's {type(projects_dir)}"
+
+    # Check if project exists
+    project_path = projects_dir / project_name
```
wavefunction91: `project_name` is joined without a containment check. Absolute names discard `projects_dir`, while `..` or an escaping symlink reaches outside it; the validator then creates and enters that location and all decorated tools operate there. This is distinct from the existing global-`cwd` thread. Could project names be restricted to one component and the resolved candidate be required to remain beneath the resolved projects root, including the direct `create_project` and `list_project_files` paths?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+def _get_default_cache() -> FolderCache:
+    """Return the default FolderCache backed by ``config.cache_dir``."""
+    return FolderCache(path=config.cache_dir)
+
+
+def _discover_cached_jobs() -> list[Job]:
+    """Discover job files in the jobs directory with cache fallback."""
+    jobs = []
+    seen: set[str] = set()
+    for jobs_dir in (config.jobs_dir, config.cache_dir):
```
wavefunction91, line 3243: `_discover_cached_jobs` scans the process-wide job/cache directories, and the public job tools never verify that a job belongs to `project_name` or the bound workspace. One workspace can therefore list another workspace's job IDs and then fetch its outputs, read its logs, or cancel it using the persisted backend configuration. Could each job record carry a workspace/project owner that all four job-management tools enforce?

python/src/qdk_chemistry/ui/validation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    """
+    markers = {}
+    for name in data.__all__:
+        obj = getattr(data, name, None)
+        if obj is None or not inspect.isclass(obj):
+            continue
+
+        # Get _data_type_name if it exists and is not None
+        type_name = getattr(obj, "_data_type_name", None)
```
wavefunction91: QDK data classes expose the public `data_type_name()` API, not `_data_type_name`, so this map currently excludes every ordinary type except the special-cased `MajoranaMapping`. Output-producing tools such as `run_scf` consequently reject `Wavefunction` before execution, and `_get_loadable_data_classes` repeats the same assumption. Could both paths use the existing data-type registry/helpers and add a positive test that a known type is accepted rather than asserting the current failure?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+        resolved_remote.connect()
+
+    sdk_kwargs: dict[str, Any] = {
+        "remote": resolved_remote,
+        "force_rerun": overwrite,
+    }
+    if cache is not None:
+        sdk_kwargs["cache"] = resolved_cache
+    else:
+        sdk_kwargs["local_cache"] = resolved_cache
```
wavefunction91: `qdk_chemistry.remote.proxy.run` accepts `cache`, not `local_cache`. On the default path the proxy therefore sees no cache and forwards `local_cache` as an algorithm keyword, so ordinary `run_*` calls fail instead of using the default `FolderCache`; the timed submission callback is also bypassed. Could `resolved_cache` always be passed through the proxy's `cache` parameter?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+
+    # If remote is a name string and remote_config is provided,
+    # create a pre-configured backend instance so the SDK receives
+    # the full configuration.
+    resolved_remote = remote
+    if isinstance(remote, str) and remote_config:
+        from qdk_chemistry.remote.backends import get_backend  # noqa: PLC0415
+
+        resolved_remote = get_backend(remote, **remote_config)
+        resolved_remote.connect()
```
wavefunction91: This path constructs and connects the backend, but passing the instance to `proxy.run` makes the proxy treat it as caller-owned and skip `disconnect()`. Repeated MCP calls with `remote_config` therefore leave connection-scoped clients and local backend work directories alive. Could this layer either retain ownership and disconnect in a `finally`, or pass configuration through a path where the proxy owns the backend lifecycle?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+        cache=cache,
+        remote=remote,
+        remote_config=remote_config,
+        remote_timeout=remote_timeout,
+        overwrite=overwrite,
+    )
+    if isinstance(result, str):
+        return result
+
+    energy, optimized_structure, wavefunction, hessian = result
```
wavefunction91: The geometry-optimizer result contract is `(energy, structure, hessian, wavefunction)`, but this assignment swaps the last two values. With the default `compute_hessian=False`, the wavefunction is lost and reported as an available Hessian; when both output paths are supplied, each object is written under the other's file type. Could this unpack in the contract order used by the binding, geomeTRIC implementation, and v2.1 example?

python/src/qdk_chemistry/ui/cli.py
```diff
@@ -0,0 +1,2565 @@
...
+
+def energy_from_phase(phase_fraction: float, *, evolution_time: float) -> float:
+    """Convert a measured QPE phase fraction to an energy."""
+    if evolution_time == 0:
+        raise ValueError("evolution_time must be nonzero")
+
+    phase_angle = (float(phase_fraction) % 1.0) * (2.0 * math.pi)
+    if phase_angle > math.pi:
+        phase_angle -= 2.0 * math.pi
+    return phase_angle / float(evolution_time)
```
wavefunction91: Why is `resolve-phase-energy` implementing its own phase inversion instead of delegating to the unitary container's `eigenvalue_from_phase()`? The mapping is representation-dependent: a Pauli product formula uses `-angle/t`, while an LCU quantum walk uses `lambda*cos(2*pi*phase)`, and a raw block encoding intentionally has no phase mapping until it is wrapped as a walk. This helper both gets the product-formula sign wrong and cannot cover the qubitization path. Could the command load the relevant `UnitaryRepresentation` and use its container's canonical inversion, applying alias/branch resolution appropriate to that representation?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+        elif masses is not None:
+            structure = data.Structure(coordinates=coordinates, symbols=symbols, masses=masses)
+        elif nuclear_charges is not None:
+            structure = data.Structure(coordinates=coordinates, symbols=symbols, nuclear_charges=nuclear_charges)
+        else:
+            structure = data.Structure(coordinates=coordinates, symbols=symbols)
+    except RuntimeError as e:
+        return f"There was a problem creating a qdk/chemistry Structure objects from input: {e}"
+
+    save_path = config.projects_dir / project_name / filename_to_save
```
wavefunction91: The existence check and save resolve different paths here: `check_output_exists` strips directories, but `save_path` preserves `..` and absolute filenames. A call using `../victim.structure.json` can therefore overwrite outside the project even with `overwrite=False`. Could the filename be normalized once, constrained beneath the project, and that same resolved path used for both checking and saving?

python/src/qdk_chemistry/ui/visualization.py
```diff
@@ -0,0 +1,1107 @@
...
+            "</script>\n"
+        )
+
+        return (
+            "<!DOCTYPE html>\n"
+            '<html lang="en">\n'
+            "<head>\n"
+            '<meta charset="utf-8"/>\n'
+            '<meta name="viewport" content="width=device-width,initial-scale=1"/>\n'
+            f"<title>{title}</title>\n"
```
wavefunction91: The scatter-plot arguments are MCP-controlled but are interpolated directly into HTML/SVG and JavaScript. A title such as `</title><script>...</script>` executes when this `text/html;profile=mcp-app` resource is rendered; axis labels also enter JavaScript source above. Could this use DOM construction/`textContent` and context-specific escaping rather than string-built executable markup?

python/src/qdk_chemistry/ui/visualization.py
```diff
@@ -0,0 +1,1107 @@
...
+        self._title = title
+        self._timeout = timeout
+        self._min_height = min_height
+        self._data: dict = {}
+        self._ready = threading.Event()
+
+    # Called by the tool
+    def send(self, payload: dict) -> CallToolResult:
+        """Store *payload*, wake the resource, and return a ``CallToolResult``."""
+        self._data = dict(payload)
```
wavefunction91: Every call writes to the same process-global payload behind a fixed resource URI. If client A calls `visualize_molecule(A)` and client B calls `visualize_molecule(B)` before A fetches the resource, B overwrites the slot and A renders B's molecule. The payload is also never cleared, so a later HTTP/SSE client can retrieve the previous client's QASM, molecule, cube, or entanglement data. Could each tool call return a unique, single-use resource URI whose payload is removed after that request consumes it? The scatter resource has the same shared-slot behavior.

cpp/src/qdk/chemistry/data/orbitals.cpp
```diff
@@ -853,7 +853,7 @@ bool Orbitals::is_restricted() const {
 bool Orbitals::has_active_space() const {
   QDK_LOG_TRACE_ENTERING();
   return !_active_space_indices.first.empty() ||
-         !_active_space_indices.second.empty();
+         !_active_space_indices.second.empty() || has_inactive_space();
```
wavefunction91, line 856: Treating an empty active list plus non-empty inactive orbitals as a defined partition makes sense, but the corresponding HDF5 update is incomplete for `ModelOrbitals`. The base `Orbitals` loader detects the zero-length active datasets by presence; `ModelOrbitals::from_hdf5` detects them by vector emptiness, so it now falls through to the full-active constructor and drops the inactive partition. Could the `ModelOrbitals` loader use dataset presence like the base class and add the same empty-active round-trip case?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,4958 @@
...
+            "for example run_time_evolution_builder(..., settings={'power': 4, 'power_strategy': 'repeat'})."
+        )
+
+    # Create the circuit mapper algorithm
+    circuit_mapper = algorithms.create("controlled_circuit_mapper", algorithm_name)
+    _apply_settings(circuit_mapper, settings)
+    circuit_mapper.settings().set("control_indices", control_indices)
+
+    try:
+        circuit = circuit_mapper.run(time_evolution_unitary)
```
wavefunction91: This tool advertises `cache`, `remote`, `remote_config`, and `remote_timeout`, but bypasses `_run_algorithm` and invokes the mapper directly. A caller requesting remote execution therefore gets a local run without any error, and the requested cache is ignored. Could this route through the common execution path or remove/reject unsupported parameters?

copilot-plugins/qdk-chemistry/skills/qdk-chemistry-mcp/references/qpe-and-state-prep.md
```diff
@@ -0,0 +1,148 @@
+# QPE and State Preparation
```
wavefunction91: The MCP/CLI/tool surface is one coherent deliverable; the prescriptive knowledge layer under `copilot-plugins/qdk-chemistry/skills/**/references/` is a separate user-facing scientific product. It defines active-space choices, QPE branch and error policy, resource budgets, feasibility thresholds, and canonical chemistry examples, none of which is required for an agent to discover or call these tools. Shipping that comprehensive policy inside the initial MCP PR also makes those independent scientific claims part of this release contract. Could this PR limit the bundled assets to tool discovery, parameter/schema reference, and mechanical usage, and remove the prescriptive chemistry/QPE/resource playbooks from the initial MCP scope?

wavefunction91 (changes requested):
I found several correctness, security, and contract issues that need resolution before merge. The inline comments are based on the integrated `main + #649 + #689` tree; no builds or tests were run as part of this review.

copilot-plugins/qdk-chemistry/skills/qdk-chemistry-mcp/references/qpe-and-state-prep.md
nabbelbabbel: Made some edits there. Without in-depth knowledge some set of additional information though. That being said it need not be as perscriptive.

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+            break
+        except (RuntimeError, ValueError, FileNotFoundError, OSError):
+            continue
+    if input_object is None:
+        return f"Failed to load {input_filename} as a Structure or Wavefunction."
+
+    analyzer = algorithms.create("population_analyzer", algorithm_name)
+    _apply_settings(analyzer, settings)
+
+    populations = _run_algorithm(
```
wavefunction91, line 2653: `PopulationAnalyzer` has the fixed run signature `run(Wavefunction)`, but this tool accepts a `Structure` and forwards `input_object`, `charge`, `spin_multiplicity`, and `n_inactive_orbitals`. Both the C++ binding and the PySCF implementation accept only the wavefunction argument, so the real call raises `TypeError`; the current MCP test hides that by mocking `_run_algorithm` and asserting these extra arguments.

Could this tool accept/load a `Wavefunction` and pass that single argument, or introduce an explicit preprocessing step if structure input is intended?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+        allowed_message = ", ".join(sorted(allowed)) or "none"
+        raise ValueError(
+            f"remote_config contains options that MCP clients cannot control for backend '{name}': "
+            f"{', '.join(sorted(unsupported))}. Allowed options: {allowed_message}."
+        )
+
+
+def _strip(filename: str) -> str:
+    """Strip directory path from a filename, keeping only the base name."""
+    return filename.rsplit("/", maxsplit=1)[-1]
```
wavefunction91: `load_data_object` and `save_data_object` now preserve exact paths, but this helper only recognizes `/`. On Windows, an MCP filename such as `..\victim.wavefunction.json` or `C:\outside\victim.wavefunction.json` survives `_strip`; `_prepare_output` checks only the basename through `check_output_exists`, then the tool loads or saves the original escaping path. This reintroduces both project-boundary access and the check/write mismatch on every tool that relies on `_strip`.

Could the MCP boundary normalize or reject both POSIX and Windows path components, with regression cases for parent-relative and absolute Windows paths, so checking and I/O always use the same project-local target?

python/src/qdk_chemistry/remote/proxy.py
```diff
@@ -246,34 +260,39 @@ def run(
         return _run_uncached(algorithm, remote, args, kwargs)
 
     payload = _build_payload_for(algorithm, args, kwargs)
+    if _owner is not None:
+        payload["owner"] = _owner
     run_hash = payload.get("run_hash")
     if run_hash is None:
         return _run_uncached(algorithm, remote, args, kwargs)
+    job_cache_key = _job_cache_key(run_hash, _owner)
+    payload["job_cache_key"] = job_cache_key
```
wavefunction91, line 269: `payload["job_cache_key"]` is set here, but it is not serialized to the compute node: `serialize_inputs()` has no corresponding parameter, and neither the Local nor Discovery backend passes one. The worker therefore falls back to the bare `run_hash`; Discovery’s cache-result fetch also uses that bare key. Owner scoping consequently stops at the client boundary: the worker can reuse or overwrite an unscoped record from another owner, while the client’s scoped shared-cache lookup cannot hit.

Could `job_cache_key` be carried through `serialize_inputs()` and both backend paths, with a two-owner shared-cache transport test confirming that worker lookup, storage, and client retrieval all use the same key?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+            completed, _ = concurrent.futures.wait(
+                (future, submitted),
+                return_when=concurrent.futures.FIRST_COMPLETED,
+            )
+            if future in completed:
+                result = future.result()
+            else:
+                job = submitted.result()
+                result = future.result(timeout=remote_timeout)
+        except concurrent.futures.TimeoutError:
```
wavefunction91, line 510: `concurrent.futures.TimeoutError` is the built-in `TimeoutError`, which is also what `Job.wait()` raises when the backend’s own timeout expires. Because this handler surrounds both branches, a real backend timeout from `future.result()` is treated as the MCP return-window timeout. Before `_on_job_submitted` runs, `job` is unbound and the original failure becomes `UnboundLocalError`; after submission, the same failure is reported as `status="submitted"` even though the waiter has already failed.

Could the timeout handler wrap only `future.result(timeout=remote_timeout)`, so backend failures propagate normally? The timed-out path should also use a bounded/managed background executor rather than creating one waiter thread per submitted call until the remote job finishes.

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+        out_wavefunction_filename, "Wavefunction", data.Wavefunction, overwrite=overwrite
+    )
+    if _err:
+        return _err
+
+    wavefunction, _err = _load_or_error(wavefunction_filename, data.Wavefunction, "wavefunction")
+    if _err:
+        return _err
+
+    if charge is not None and algorithm_name == "qdk_valence":
```
wavefunction91, line 1231: The explicit `qdk_valence` path never applies the caller’s `settings`; it creates a fresh selector and overwrites the active electron/orbital counts with the automatically computed values. A request such as `settings={"num_active_orbitals": 6}` therefore succeeds while silently ignoring the scientific override.

Could this apply the supplied settings and auto-fill only values the caller did not provide, or reject the conflicting combination explicitly?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+        unitary_builder_settings = getattr(unitary_builder, "settings", None)
+        evolution_time = unitary_builder_settings.get("time") if unitary_builder_settings is not None else 0.0
+    except (KeyError, RuntimeError):
+        evolution_time = 0.0
+    if evolution_time == 0.0:
+        return (
+            "You need to set settings.qpe_circuit_builder.unitary_builder.time for QPE. "
+            "The default value of 0.0 is invalid. "
+            "Choose a value based on the eigenvalue range of your Hamiltonian. "
+            "A good starting point is evolution_time = 2*pi / (E_max - E_min), "
```
wavefunction91: This error message embeds scientific parameter-selection policy in the MCP layer. Please remove the evolution-time formula and the `0.1–10.0` recommendation. The tool should validate that the selected implementation received a valid explicit value and report the missing/invalid setting; it should not prescribe that value.

python/src/pybind11/algorithms/geometry_optimization.cpp
```diff
@@ -95,6 +95,8 @@ Optimize a molecular structure.
       R"(Internal settings replacement hook for Python subclasses.)");
   optimizer.def("name", &GeometryOptimizer::name,
                 R"(Return the implementation name.)");
+  optimizer.def("aliases", &GeometryOptimizer::aliases,
```
wavefunction91, line 103: `Algorithm::hash` exists for every C++ algorithm type and is exposed by the sibling bindings, but `GeometryOptimizer` and `PopulationAnalyzer` do not bind it. Their Python implementations subclass those pybind classes rather than the Python `Algorithm` base, so they do not inherit the Python fallback either.

`_run_algorithm` requires `hash()` before every timed remote submission, and `remote_timeout` defaults to 120, so `run_geometry_optimization(remote=...)` and `run_population_analysis(remote=...)` fail before reaching the backend. Could both bindings expose their existing C++ hash methods and be included in an all-registered-implementations hash test?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+
+# =========================
+# Helpers
+# =========================
+
+
+def _validate_mcp_remote_config(name: str, remote_config: dict[str, Any]) -> None:
+    """Reject backend constructor options that MCP clients may not control."""
+    allowed = get_mcp_safe_config_options(name)
+    unsupported = remote_config.keys() - allowed
```
wavefunction91, line 216: The MCP allowlist validates option names but not the values of the newly exposed `timeout` and `poll_interval` controls. For example, JSON `1e309` becomes `+Inf`, so `Job.wait()` can never reach its timeout; a negative polling interval instead fails in `time.sleep()` after the remote job has already been submitted.

Could MCP-provided timing values be required to be non-boolean numeric, finite, and positive at this trust boundary, without changing the general SDK backend contract?

python/src/qdk_chemistry/ui/cli.py
```diff
@@ -0,0 +1,2560 @@
...
+    if not projects_dir.exists():
+        _print_result([])
+        return
+    projects = sorted(d.name for d in projects_dir.iterdir() if d.is_dir())
+    _print_result(projects)
+
+
+def cmd_utils_create_project(args):
+    """Create a new project directory."""
+    project_dir = config.projects_dir / args.project_name
```
wavefunction91: The direct CLI handlers still construct project paths with `config.projects_dir / args.project_name` instead of using the containment logic in `resolve_project_path()`. Values such as `..` or an absolute path therefore let `create-project`, `list-files`, the data inspectors, and `resolve-phase-energy` operate outside the configured projects directory, even though the equivalent MCP paths reject them.

Could these handlers share the same project resolver rather than maintaining a second, unconstrained path implementation?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3572 @@
...
+        mapping,
+        cache=cache,
+        remote=remote,
+        remote_config=remote_config,
+        remote_timeout=remote_timeout,
+        overwrite=overwrite,
+    )
+
+    # save to file
+    save_data_object(qubit_hamiltonian, out_qubit_hamiltonian_filename)
```
wavefunction91, line 2214: QDK qubit mapping intentionally excludes the fermionic Hamiltonian’s core-energy constant; existing SDK examples add `hamiltonian.get_core_energy()` back when reporting total molecular energies. This MCP handoff returns only the `QubitHamiltonian` filename, while the downstream solver, estimator, and QPE tools return the resulting electronic-frame values without carrying or identifying that omitted offset.

Could `run_qubit_mapper` preserve the companion `core_energy` in its result and document the mapped-energy convention, so downstream MCP results cannot be mistaken for total molecular energies?

python/src/qdk_chemistry/ui/visualization.py
```diff
@@ -0,0 +1,923 @@
...
+        app: MCP server application that receives the widget resources and
+            tools.
+
+    """
+    if not _WIDGETS_AVAILABLE:
+        return
+
+    # ── Minimal test tool ─────────────────────────────────────────
+    @app.resource(
+        "ui://qdk-chem-mcp/test-square",
```
wavefunction91: Why is the MCP Apps test-square fixture part of the production tool catalog? It is not used by the chemistry UI, and its page inserts the caller-controlled label through `innerHTML`, making arbitrary markup executable in the iframe.

Could this test resource and `visualize_test_square` tool be removed from the shipped surface rather than documented as a public visualization?

python/src/qdk_chemistry/ui/validation.py
```diff
@@ -0,0 +1,242 @@
...
+            project_name: Project name inside default project directory
+            *args: Additional positional arguments passed to the decorated function
+            **kwargs: Additional keyword arguments passed to the decorated function
+
+        Returns:
+            String with error information if validation fails, otherwise
+            returns the result of the decorated function
+
+        """
+        with _PROJECT_CWD_LOCK:
```
wavefunction91: `_PROJECT_CWD_LOCK` is held for the entire decorated tool body, and synchronous MCP tools are dispatched on worker threads, so one long calculation serializes every other project-bound call. This also blocks the asynchronous job interface itself: `check_remote_job`, `list_remote_jobs`, and `cancel_remote_job` use absolute configured job paths and do not need the process CWD, but they cannot run while another decorated calculation holds this lock.

Could the job-management tools avoid the CWD-changing decorator, and could the lock otherwise be narrowed so polling and cancellation remain available during long-running work?

python/src/pybind11/algorithms/factory_bindings.hpp
```diff
@@ -275,4 +275,111 @@ Check if an algorithm name exists in the registry.
...
+ *
+ * Example usage:
+ * @code
+ * py::class_<ScfSolver, ...> scf_solver(m, "ScfSolver");
+ * // ... bind other methods ...
+ * add_on_remote_method(scf_solver);
+ * @endcode
+ */
+template <typename PyClass>
+void add_on_remote_method(PyClass& py_class) {
```
wavefunction91: These `add_on_remote_method()` and `add_hash_method()` templates have no call sites. The remote helper would also fail if wired because `qdk_chemistry.remote.RemoteAlgorithmProxy` does not exist; remote execution is currently exposed through the registry wrapper/module-level API instead.

Could these unused templates be removed rather than retaining an untested second API design in the binding layer?

copilot-plugins/qdk-chemistry/skills/qdk-chemistry-coding/SKILL.md
```diff
@@ -0,0 +1,53 @@
...
+
+Algorithms are obtained from the registry:
+
+- `available(algorithm_type)` returns registered implementations.
+- `show_default(algorithm_type)` returns the registered default.
+- `create(algorithm_type, algorithm_name, **settings)` creates an algorithm.
+- `algorithm.settings()` returns its settings object.
+- `algorithm.run(...)` executes it.
+- `algorithm.hash(...)` computes the content hash for a call.
+- `algorithm.on_remote(...)` returns a remote-bound algorithm.
```
wavefunction91: `algorithm.on_remote()` is not part of the implemented Python API: the pybind helper that mentions it has no call sites, `RemoteAlgorithmProxy` does not exist, and remote execution is exposed through `run(..., remote=...)` and the module-level `qdk_chemistry.remote.run()` / `submit()` functions. The retained `remote-execution` skill repeats the same method-level API and also describes an SSH backend that this package does not ship.

Could both retained skills be updated to describe the remote API and registered backends that actually exist?

docs/source/user/agents.rst
```diff
@@ -0,0 +1,147 @@
...
+Getting started
+---------------
+
+1. **Install QDK/Chemistry** into a virtual environment (see :doc:`quickstart`).
+
+2. **Deploy agent configs** into your project directory:
+
+   .. code-block:: bash
+
+      # Install the QDK/Chemistry plugin and deploy its agents and skills.
```
wavefunction91: The published plugin no longer contains an `agents/` directory or an `agents` manifest entry, but this page still instructs users to deploy agent configs, says installation creates agent definitions, lists `.github/agents/` as an installed component, and documents editing the five removed agents. `README.md` retains the same multi-agent claim.

Could the agent-specific instructions and README statements be removed or rewritten to describe the skill-and-MCP plugin that now actually ships?

wavefunction91 (commented):
The nuclear-derivative changes are orthogonal to adding the MCP surface. This PR changes the public default from `qdk_finite_difference` to `qdk` for both derivative calculations and geometry optimization, and adds a new gradient-difference Hessian implementation. The `qdk` implementation was already selectable by name, while the prior default already supplied Hessians, so none of these semantic changes is required by the MCP tools.

The switch also changes existing multi-reference behavior: the new default rejects any `energy_calculator` other than `scf_solver/qdk`, where the previous default accepted CASCI/ASCI/MCSCF paths, and the change is absent from the release notes. Could the default switch and new Hessian implementation be removed from this PR and reviewed as their own scientific change with the corresponding compatibility notes and numerical tests?

wavefunction91 (commented):
QRE lives in upstream QDK, outside this repository. `run_resource_estimation` adds a partial, hard-coded wrapper here: two architectures, fixed QEC/factory choices, selected timing/error controls, and an incomplete set of the assumptions that determine the result.

Could the QRE helper, MCP tool, catalog/skill entries, and tests be removed from this PR? If QRE needs an MCP surface, that integration should be designed in upstream QDK rather than exposed through a narrow QDK Chemistry compatibility layer.

nabbelbabbel:
> QRE lives in upstream QDK, outside this repository. `run_resource_estimation` adds a partial, hard-coded wrapper here: two architectures, fixed QEC/factory choices, selected timing/error controls, and an incomplete set of the assumptions that determine the result.
> 
> Could the QRE helper, MCP tool, catalog/skill entries, and tests be removed from this PR? If QRE needs an MCP surface, that integration should be designed in upstream QDK rather than exposed through a narrow QDK Chemistry compatibility layer.

This is getting confusing on what the boundaries are supposed to be. 

I understand that the implemented version of `run_resource_estimation` mirroring what is in the [examples](https://github.com/microsoft/qdk-chemistry/blob/main/examples/qpe_stretched_n2.ipynb) but then scoping down possible choices may not be what we want. But it seems to me that `data.circuit.estimate()` does the same thing conceptually, it just uses another (older) technical route. It too uses QRE from QDK and allows only a subset of functionality/features that QRE from QDK has to date.

While `data.circuit.estimate()` exists it only makes sense that the MCP can at least use/expose that functionality. If we want to remove all QRE here (meaning `data.circuit.estimate()` too) that seems like a clean way of resolving this all up. But this would be something for another PR and probably warrant a deprecation warning.

For now I have removed the `run_` version and added a smaller helper that does estimation via `data.circuit.estimate()`, I have added a skill explaining this helper and also the fully fledged code based QDK/QRE path, so that in the future we can just edit that skill, keeping any agentic system around the MCP to have the same knowledge we have now, just using different tools. After all, these MCP tools should not be something that anyone hard codes against and us explicitly pointing at a feature set in a QDK MCP or alike in a future version should be an acceptable change.

python/src/qdk_chemistry/ui/validation.py
nabbelbabbel: The latest solution should get rid of it alltogether.

nabbelbabbel:
> QRE lives in upstream QDK, outside this repository. `run_resource_estimation` adds a partial, hard-coded wrapper here: two architectures, fixed QEC/factory choices, selected timing/error controls, and an incomplete set of the assumptions that determine the result.
> 
> Could the QRE helper, MCP tool, catalog/skill entries, and tests be removed from this PR? If QRE needs an MCP surface, that integration should be designed in upstream QDK rather than exposed through a narrow QDK Chemistry compatibility layer.

See #696

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3445 @@
...
+
+    structure, err = _load_or_error(structure_filename, data.Structure, "structure")
+    if err:
+        return err
+
+    seed, err = _resolve_seed_or_basis(seed_or_basis)
+    if err:
+        return err
+
+    calculator = algorithms.create("nuclear_derivative_calculator", algorithm_name)
```
wavefunction91, line 2724: Moving the nuclear-derivative implementation work out restored the `qdk_finite_difference` default, but the MCP surface was not reconciled with that revert. I reproduced the current head in a dev-session: both registered nuclear-derivative implementations fail the all-registered `hash()` and `aliases()` tests, `describe_algorithm(..., "qdk")` returns an error, and the MCP discovery test still expects `qdk` instead of the restored default.

Could the generic `hash`/`aliases` bindings be retained for `NuclearDerivativeCalculator` and the MCP tests be updated to the actual `qdk_finite_difference` default?

python/src/qdk_chemistry/ui/cli.py
```diff
@@ -0,0 +1,2570 @@
...
+            summary = obj.get_summary() if hasattr(obj, "get_summary") else str(obj)
+            _print_success(type=cls.__name__, summary=summary)
+            return
+        except (RuntimeError, ValueError, FileNotFoundError, OSError):
+            continue
+
+    _print_error(f"Could not load '{filename}' as any known data type.")
+
+
+def cmd_data_convert(args):
```
wavefunction91, line 796: Project names now use the shared resolver, but the direct CLI data handlers still strip filenames with `split("/")`. On Windows, `..\secret.json`, `..\victim.hdf5`, and absolute drive paths survive unchanged. These handlers run outside the MCP project context, so `load_data_object` / `save_data_object` preserve the surviving path and read or write outside the validated project after `chdir`.

Could the direct CLI handlers use the same cross-platform filename confinement as the MCP tools, with Windows parent-relative and drive-path tests?

python/src/qdk_chemistry/ui/tools.py
```diff
@@ -0,0 +1,3446 @@
...
+    remote_timeout: int = 120,
+    overwrite: bool = False,
+) -> str | dict[str, str | float]:
+    """Save a mapped QubitHamiltonian excluding core energy and return its filename with the companion offset."""
+    # Strip filenames in case full path is passed
+    hamiltonian_filename = _strip(hamiltonian_filename)
+    mapping_filename = _strip(mapping_filename)
+    out_qubit_hamiltonian_filename = _strip(out_qubit_hamiltonian_filename)
+    out_qubit_hamiltonian_filename, _err = _prepare_output(
+        out_qubit_hamiltonian_filename, "QubitHamiltonian", data.QubitHamiltonian, overwrite=overwrite
```
wavefunction91: This output path is unreachable in the current build. `_DATA_TYPE_MARKERS` is generated from registered data classes, where the `qubit_hamiltonian` wire type is registered as `QubitOperator`; `QubitHamiltonian` is a distinct deprecated subclass alias and is therefore absent from the marker map. Passing `"QubitHamiltonian"` here makes `_prepare_output` reject every filename before the mapper runs. `create_spin_model_hamiltonian` has the same call at line 1550.

I reproduced both public tools unmocked: each returns `Invalid output filename: Unrecognized data type QubitHamiltonian`. The current mapper test masks this by patching `_prepare_output`. Could both call sites use the registered `QubitOperator` type/name and add an unmocked output-validation regression test?

nabbelbabbel: Done

wavefunction91 (approved):
Re-reviewed the amended head. The remaining qubit-output blocker is fixed and reproduced unmocked; the targeted nuclear/CLI regressions pass. No hard blocker remains.

PR #695, opened 2026-09-01
Make phase inversion correct for powered unitary representations

### LINKED ISSUE #694: Make phase inversion correct for powered unitary representations
## Problem

`UnitaryContainer.eigenvalue_from_phase()` promises to recover the Hamiltonian eigenvalue for the unitary represented by the container, but the current implementations only invert unit-power representations.

- For a product formula built with repeat power `p`, the circuit represents `exp(-i H t p)` while the container retains `scale=t`; inversion therefore returns `pE`, and the alias period is also wrong by `p`.
- `LCUWalkContainer` stores `power`, but `QuantumWalkContainer.eigenvalue_from_phase()` ignores it and evaluates `lambda * cos(p * acos(E/lambda))` rather than recovering `E`.

This surfaced while #689 delegated its `resolve-phase-energy` command to the container API. The caller should remain representation-agnostic; the general container contract needs to handle the represented power.

## Expected behavior

Phase-to-eigenvalue conversion should be correct for every valid powered representation carried by a `UnitaryContainer`, with any unavoidable multiple branches represented explicitly rather than silently returning a plausible wrong value.

## Acceptance criteria

- Preserve enough total-time/power metadata to invert repeated product formulas correctly.
- Define and implement the inverse-branch contract for powered quantum walks.
- Add round-trip tests for power one and power greater than one, including serialized representations.

RushiGong:
This pull request updates the Quantum Phase Estimation (QPE) workflow and its documentation to correctly handle cases where the phase-to-energy mapping is not unique—specifically, when the underlying unitary is raised to a power, causing several energies to correspond to the same measured phase. The changes ensure that all candidate energies are properly reported and handled throughout the codebase and documentation. The most important changes are:

**API and Data Model Updates:**
* The `eigenvalue_from_phase` method in all relevant `UnitaryContainer` classes now returns a tuple of all possible energy candidates for a given phase, not just a single value. This allows the code to handle ambiguous cases where multiple energies map to the same phase. 
* The `QpeResult` data structure and its factory methods are updated to store all candidate energies in the `branching` field, and to clarify which candidate is selected as the "raw" energy.

**Phase Estimation Algorithm Changes:**
* The phase estimation post-processing now collects all candidate energies from the container, storing them in `branching`, and selects the lowest as the default `resolved_energy`.

**Container Construction and Scaling:**
* When constructing powered time-evolution containers, the total evolution time (`scale`) is now correctly multiplied by the number of repetitions, ensuring phase-to-energy conversion is accurate.

**Documentation Improvements:**
* The user documentation for QPE results is updated to explain the meaning of the `branching` field, the handling of ambiguous phase-to-energy mappings, and the correct interpretation of the evolution time. 

**Bugfixes and Logic Corrections:**
* The amplitude amplification logic now correctly checks all energy candidates when marking phase bins, not just the first. 

These changes ensure that QPE results are accurate and unambiguous, even when the unitary is powered, and that both the API and documentation reflect the correct handling of multiple energy candidates.

Fixes #694.


wavefunction91 (commented):
I think there are two separate cases here: a saved representation of `U^p` must invert the phase of `U^p`, while standard and iterative QPE still invert the base `U` because their powers are only part of the query schedule. The scale, powered-walk inversion, and `power=1` fixes all make sense under that distinction.

Could we keep `eigenvalue_from_phase()` scalar and restore a separate branch-aware method for powered walks? Changing the existing method to return a tuple pushes this concern through ordinary QPE callers even though they now explicitly use `power=1`, and `QpeResult.from_phase_fraction()` then has to choose `min(candidates)` as `raw_energy` without a physical reason. The earlier shape—scalar inversion raises for an ambiguous `W^p`, while a separate method returns all candidates—keeps the powered-container behavior at the container and CLI boundary without changing normal QPE result semantics.

One more place needs the new scale invariant: `PauliProductFormulaContainer.reorder_terms()` rebuilds the container without passing `scale`, so it resets to `1.0`. Reordering a representation with any other evolution time—including a powered one—then leaves the circuit unchanged but makes phase inversion wrong. Can this preserve `self.scale` and get a regression test?

python/src/qdk_chemistry/ui/cli.py
```diff
@@ -1166,7 +1166,8 @@ def cmd_utils_resolve_phase_energy(args):
 
     unitary = load_data_object(filename, qdk_data.UnitaryRepresentation)
     container = unitary.get_container()
-    raw_energy = container.eigenvalue_from_phase(args.phase_fraction)
+    branching = tuple(sorted(container.eigenvalue_from_phase(args.phase_fraction)))
```
wavefunction91: `--reference-energy` is still ignored for powered walks here. For example, a `W^2` representation at phase `0.1` has candidates near `-5.7063` and `+5.7063`, but a reference energy of `5.8` still resolves to the negative value because only the product-formula path uses the reference. Can we select the nearest walk branch and cover this with a serialized `W^p` CLI test? That would exercise the motivating path from #694 end to end.

feisuxiaozhu (approved):
all looks good!

PR #696, opened 2026-09-01
Use analytic gradients by default for nuclear derivatives

nabbelbabbel:
- Make the QDK analytic nuclear derivative calculator the default.
- Use analytic SCF gradients for nuclear-gradient calculations.
- Compute nuclear Hessians using central finite differences of analytic gradients.
- Update geometry optimization to use the analytic derivative calculator by default.
- Expose derivative calculator hashing and aliases through Python bindings.
- Update C++ and Python tests for the new behavior.


nabbelbabbel:
@copilot some of the new tests added to this PR by merging in main need to be updated to the new default too, do that.

nabbelbabbel:
@copilot looking at the tests it seems like you missed an instance.

PR #702, opened 2026-09-02
Add QCE 2026 tutorial landing page to the published docs site

Rohit-2002-28:
## Summary

Adds the companion landing page for our QCE 2026 / IEEE Quantum Week tutorial to the published `qdk-chemistry` docs site.

**Tutorial:** *From Molecules to Logical Qubits: A Practical Tutorial on the Microsoft Quantum Development Kit (QDK)*
**When:** Wednesday 16 September 2026
**Where:** Room 715B, Metro Toronto Convention Centre

The page lists the two 90-minute sessions, the agenda, the target audience, and offers `.ics` downloads plus Google Calendar links so attendees can add either session (or both) to their calendar.

It is currently hosted on a personal GitHub Pages repo. We'd like it to live under an official Microsoft repo instead, which is what this PR does.

**Resulting URL:** `https://microsoft.github.io/qdk-chemistry/qce26-tutorial/`

## Why `html_extra_path` and not a commit to `gh-pages`

GitHub Pages for this repo serves from the `gh-pages` branch (`/docs` folder), but that branch is **100% generated Sphinx output** — its commits are titled "Update docs" / "Update docs to 2.0.0" and its tree is build artifacts. Anything committed there by hand would be destroyed by the next docs publish.

So the page has to enter the site *as part of the Sphinx build*, from source on `main`. Sphinx's [`html_extra_path`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path) is exactly the right mechanism: it copies files verbatim into the root of the HTML output with **no template or reST processing**, so the page is published byte-for-byte as authored.

That's a one-line addition to `docs/source/conf.py`, right next to the existing `html_static_path`:

```python
html_extra_path: list[str] = ["_extra"]  # Files copied verbatim into the output root
```

Everything under `docs/source/_extra/` is then mirrored into the output root, so `_extra/qce26-tutorial/index.html` becomes `/qce26-tutorial/`.

## Why this PR touches `.pre-commit-config.yaml`

This is the non-obvious bit, so flagging it up front — reviewers will otherwise wonder why a lint config is being changed in a docs PR.

RFC 5545 **requires CRLF line endings** in iCalendar files, and Outlook rejects `.ics` files that violate it. Our `.pre-commit-config.yaml` sets `files: &linted_files .*` and enables `mixed-line-ending` with `args: [--fix=lf]`, which would physically rewrite the `.ics` files to LF and break the "Add to calendar" buttons. `trailing-whitespace` and `end-of-file-fixer` also match `.*` and are similarly risky for these files.

Hence the one-token change on line 6:

```diff
-exclude: \.(xyz|fcidump|wfn|qasm)$
+exclude: \.(xyz|fcidump|wfn|qasm|ics)$
```

This follows the existing precedent in that line, which already exempts other byte-sensitive data formats (`.xyz`, `.fcidump`, `.wfn`, `.qasm`).

There is also a scoped `docs/source/_extra/qce26-tutorial/.gitattributes` with `*.ics -text`. **Both defences are needed and they solve different problems:** `.gitattributes` stops *git's own* CRLF normalisation on checkin/checkout (the repo root sets `* text=auto eol=lf`), whereas pre-commit rewrites the file on disk *before git ever sees it*. Neither one alone is sufficient.

Only `.ics` is exempted — `index.html` is still fully linted and normalised like any other file in the repo.

## Changes

7 files: 5 added, 2 modified.

| File | Change |
|---|---|
| `docs/source/_extra/qce26-tutorial/index.html` | added (822 lines) |
| `docs/source/_extra/qce26-tutorial/session1.ics` | added (49 lines) |
| `docs/source/_extra/qce26-tutorial/session2.ics` | added (49 lines) |
| `docs/source/_extra/qce26-tutorial/both-sessions.ics` | added (73 lines) |
| `docs/source/_extra/qce26-tutorial/.gitattributes` | added (3 lines) |
| `docs/source/conf.py` | +1 line |
| `.pre-commit-config.yaml` | 1 line changed |

No existing file is reordered or reformatted, and there is zero functional change to the landing page itself.

## The page is fully self-contained

Verified: no `<base>` tag, no absolute `href="/…"` / `src="/…"` paths, no protocol-relative URLs, and no external CSS/JS/font loads (the favicon is an inline `data:` SVG and all CSS/JS is inline). The only outbound URLs are ordinary navigation links (Google Calendar, github.com, quantum.microsoft.com). The `.ics` links are relative (`session1.ics`, etc.).

That's what makes it safe to serve from a subpath like `/qce26-tutorial/`.

## Verification performed

A full `make all` docs build needs Doxygen, Graphviz and the compiled `qdk-chemistry` package, so instead:

- **`html_extra_path` really does copy verbatim** — built a throwaway minimal Sphinx project (Sphinx 9.1.0) in a temp dir with its own tiny `conf.py`, containing copies of all four files. All four landed at `<output>/qce26-tutorial/` **byte-identical** (sha256 match), with CRLF intact in the `.ics` files.
- **CRLF survives the commit** — read the committed blobs back out of git: `session1.ics` 49 CRLF / 0 bare LF, `session2.ics` 49 / 0, `both-sessions.ics` 73 / 0, and sha256 of each blob matches the source file exactly.
- **`conf.py` still parses** — `ast.parse` OK.
- **`.pre-commit-config.yaml` still parses** — `yaml.safe_load` OK, new `exclude` regex compiles, and simulating pre-commit's `files`-AND-NOT-`exclude` filter confirms the three `.ics` paths are skipped while `index.html`, `conf.py` and the config itself remain linted.

## :warning: Notes for reviewers

- **The page will not appear until the docs are published to `gh-pages` again.** That branch is produced by a separate publish step rather than by CI on merge, so merging this PR alone will not make the URL live. Could a maintainer confirm how and when that publish runs, and whether anything else is needed to trigger it?
- **Please confirm the final URL before merge.** A QR code on a printed flyer currently points at the old personal URL; it will be regenerated to point at the new one once this lands, so we want to be sure `https://microsoft.github.io/qdk-chemistry/qce26-tutorial/` is the path we're committing to.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_stages: [pre-commit, pre-push]
 # Define the list of directories to lint
 # You can extend this list to include more directories over time
 files: &linted_files .*
-exclude: \.(xyz|fcidump|wfn|qasm)$
+exclude: \.(xyz|fcidump|wfn|qasm|ics)$
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/docs/source/_extra/qce26-tutorial/.gitattributes
+++ b/docs/source/_extra/qce26-tutorial/.gitattributes
@@ -1,0 +1,3 @@
+# RFC 5545 requires CRLF line endings in iCalendar files.
+# Treat .ics as binary so git never normalises them.
+*.ics -text
--- a/docs/source/_extra/qce26-tutorial/both-sessions.ics
+++ b/docs/source/_extra/qce26-tutorial/both-sessions.ics
@@ -1,0 +1,73 @@
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Microsoft Quantum Development Kit//QCE 2026 Tutorial//EN
+CALSCALE:GREGORIAN
+METHOD:PUBLISH
+X-WR-CALNAME:Microsoft QDK Tutorials - QCE 2026
+X-WR-TIMEZONE:America/Toronto
+BEGIN:VTIMEZONE
+TZID:America/Toronto
+BEGIN:DAYLIGHT
+TZOFFSETFROM:-0500
+TZOFFSETTO:-0400
+TZNAME:EDT
+DTSTART:19700308T020000
+RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=2SU
+END:DAYLIGHT
+BEGIN:STANDARD
+TZOFFSETFROM:-0400
+TZOFFSETTO:-0500
+TZNAME:EST
+DTSTART:19701101T020000
+RRULE:FREQ=YEARLY;BYMONTH=11;BYDAY=1SU
+END:STANDARD
+END:VTIMEZONE
+BEGIN:VEVENT
+UID:qdk-qce2026-tutorial-session1@microsoft.com
+DTSTAMP:20260811T000000Z
+DTSTART;TZID=America/Toronto:20260916T100000
+DTEND;TZID=America/Toronto:20260916T113000
+SUMMARY:QDK Tutorial (Session 1): From Molecules to Logical Qubits
+DESCRIPTION:Session 1 of the Microsoft Quantum Development Kit tutorial at 
+ IEEE Quantum Week (QCE 2026).\n\nFrom Molecules to Logical Qubits: A Pract
+ ical Tutorial on the Microsoft Quantum Development Kit (QDK)\n\nFull detai
+ ls: https://qce.quantum.ieee.org/2026/qce26-schedule/tutorial-schedule/#12
+ 97-1
+LOCATION:Room 715B\, Metro Toronto Convention Centre\, 255 Front St W\, Tor
+ onto\, ON\, Canada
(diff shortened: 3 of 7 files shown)


Rohit-2002-28:
________________________________

nabbelbabbel:
A few questions:

@Rohit-2002-28 is there anything here that actually requires extra html things? Could this be standard .rst as the rest of the docs instead?

Is this ephemeral and supposed to be removed entirely after the event? I dislike these types of ephemeral things in the repo. Should this instead be a more general page for events that simply tracks upcoming and past ones?

Rohit-2002-28:
> @nabbelbabbel 

> Nothing *functional* needs HTML — the content is text and links, the `.ics` files are plain anchors, and the only JS is a 25-line nav toggle. It'd convert cleanly.
It's the visual design that needs it. This is the QR-code destination on a printed flyer we're handing out at the IEEE conference, so it's built to match that flyer. As a themed RST page it'd render with the API-docs sidebar and breadcrumbs — right for reference material, wrong for a landing page.

>This is meant to be ephemeral, as it is just meant to be for the IEEE conference tutorial session. For it to be a more general events page, I'd have to check with Sasha and the design team first. Open to that suggestion as well

wavefunction91:
Integrated via direct deployment.

PR #703, opened 2026-09-03
Add azure quantum as circuit executor backend

RushiGong:


python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,291 @@
...
+    Returns:
+        A ``(bitstring_counts, loss_bitstrings)`` tuple of label-to-count dicts; the latter is empty absent qubit loss.
+
+    """
+    counts: dict[str, int] = {}
+    loss: dict[str, int] = {}
+    for entry in raw_results.values():
+        outcome = entry["outcome"]
+        count = entry["count"]
+        outcome_bits = outcome if isinstance(outcome, (list | tuple)) else (outcome,)
```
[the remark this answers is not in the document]
RushiGong: This was enforced by pre-commit. Won't fix.

python/src/qdk_chemistry/algorithms/phase_estimation/circuit_builder/iterative_builder.py
```diff
@@ -200,6 +222,73 @@ def _create_circuit_from_qsharp_op(
...
+
+        """
+        num_system_qubits = qubit_hamiltonian.num_qubits
+
+        # One controlled unitary per round, mirroring the per-bit path: round k uses U^(2^(num_bits-1-k)).
+        ctrl_unitary_circuits: list[Circuit] = []
+        num_ancilla_qubits = 0
+        for k in range(num_bits):
+            circuit, num_ancilla_qubits = self._create_controlled_circuit(qubit_hamiltonian, 2 ** (num_bits - 1 - k))
+            ctrl_unitary_circuits.append(circuit)
```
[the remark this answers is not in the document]
RushiGong: IQPE uses one ancilla each round, the num_ancilla_qubits stay the same.

python/src/qdk_chemistry/data/circuit.py
```diff
@@ -253,6 +259,7 @@ def get_qsharp_circuit(self, prune_classical_qubits: bool = False) -> QdkCircuit
                 self._qsharp_factory.program,
                 *self._qsharp_factory.parameter.values(),
                 prune_classical_qubits=prune_classical_qubits,
+                generation_method=generation_method,
```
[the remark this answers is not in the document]
RushiGong: "None" is the default setting of qdk.

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,307 @@
...
+        required = ("subscription_id", "resource_group", "workspace_name", "location", "target_name")
+        coordinates = {key: self._settings.get(key) for key in required}
+        missing = [key for key, value in coordinates.items() if not value]
+        if missing:
+            raise ValueError(
+                "Azure Quantum target cannot be resolved; set "
+                + ", ".join(f"{key} (${_WORKSPACE_ENV_VARS[key]})" for key in missing)
+            )
+
+        qir_string = str(circuit.get_qir())
```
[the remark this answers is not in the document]
RushiGong: The entry point and profile live inside the QIR itself ("entry_point", "qir_profiles"="base_profile"), which str() preserves verbatim, so no metadata is lost.

docs/source/_static/examples/python/circuit_executor.py
```diff
@@ -38,6 +38,32 @@
...
+executor.settings().set("workspace_name", "my-workspace")
+executor.settings().set("location", "my-location")
+executor.settings().set("target_name", "my.emulator.target")
+executor.settings().set("auth_mode", "azure-cli")
+executor.settings().set(
+    "emulation_settings",
+    json.dumps(
+        {"simulationType": "cliffordrounding", "enableNoise": False, "seed": 42}
+    ),
+)
```
wavefunction91: Can this use the position args API?

docs/source/_static/examples/python/circuit_executor.py
```diff
@@ -38,6 +38,32 @@
 # end-cell-configure-qiskit
 ################################################################################
 
+################################################################################
+# start-cell-configure-azure-quantum
+# Configure the Azure Quantum emulator. Each connection setting falls back to the
+# matching AZURE_QUANTUM_* environment variable when omitted.
+import json
+
+executor = create("circuit_executor", "azure_quantum_emulator")
```
wavefunction91: Why is this specifically called emulator? Should it not be general to any AQ backend?

docs/source/user/comprehensive/algorithms/circuit_executor.rst
```diff
@@ -184,6 +191,70 @@ This implementation is available through the :doc:`Qiskit plugin <../plugins>`.
      - Qiskit transpilation optimization level (0–3). Default is 0.
 
 
+Azure Quantum emulator
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. rubric:: Factory name: ``"azure_quantum_emulator"``
+
+Compiles circuits to :term:`QIR` and submits them to an Azure Quantum target, returning the measurement results.
+Available through the :doc:`Azure Quantum plugin <../plugins>`; requires ``azure-quantum`` and ``azure-identity``.
```
wavefunction91: Specify these are pypi packages

docs/source/user/comprehensive/algorithms/circuit_executor.rst
```diff
@@ -184,6 +191,70 @@ This implementation is available through the :doc:`Qiskit plugin <../plugins>`.
...
+
+The connection to the Azure Quantum workspace is established using the provided settings.
+The five workspace settings fall back to the matching ``AZURE_QUANTUM_*`` environment variable, so a configured shell needs no arguments.
```
wavefunction91: Is this standard AQ behaviour or something we're introducing?

python/src/qdk_chemistry/algorithms/phase_estimation/circuit_builder/iterative_builder.py
```diff
@@ -200,6 +222,75 @@ def _create_circuit_from_qsharp_op(
...
         )
 
+    def _create_full_circuit(
+        self,
+        state_preparation: Circuit,
+        qubit_hamiltonian: QubitOperator,
+        *,
+        num_bits: int,
+    ) -> Circuit:
+        """Construct a single circuit implementing the full IQPE with in-circuit feedback.
```
wavefunction91: Would the simpler description be "a single dynamic circuit"?

python/src/qdk_chemistry/algorithms/phase_estimation/iterative_phase_estimation.py
```diff
@@ -105,6 +105,21 @@ def _run_impl(
...
+            return self._run_single_circuit(
+                circuit_builder=circuit_builder,
+                circuit_executor=circuit_executor,
+                state_preparation=state_preparation,
+                qubit_hamiltonian=qubit_hamiltonian,
+                container=container,
+                num_bits=num_bits,
+                noise=noise,
+            )
+
```
wavefunction91: This is fragile and couples to other module's implementation details.

python/pyproject.toml
```diff
@@ -137,6 +137,11 @@ test = [
   "setuptools>=64",
   "wheel"
 ]
+azure-quantum = [
+  # azure-quantum requires azure-identity and azure-storage-blob, which both pull in cryptography; no win-arm64 wheel
+  "azure-identity>=1.19.0,<2; sys_platform != 'win32' or platform_machine != 'ARM64'",
+  "azure-quantum; sys_platform != 'win32' or platform_machine != 'ARM64'"
+]
```
wavefunction91: We need to double check this.

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,307 @@
...
+        Logger.trace_entering()
+        super().__init__()
+        self._set_default(
+            "emulation_settings",
+            "string",
+            json.dumps(_DEFAULT_EMULATION_SETTINGS),
+            "Azure Quantum emulationSettings, as a JSON object string",
+        )
+        self._set_default("job_name", "string", "qdk-chemistry-azure-quantum-emulator", "Name for the submitted job")
+        self._set_default("timeout_secs", "int", 3600, "Maximum seconds to wait for job completion")
```
wavefunction91: what happens if we don't make the timeout? Are these cached to be restarted?

python/src/qdk_chemistry/utils/qsharp/src/CombinedIterationPhaseEstimation.qs
```diff
@@ -0,0 +1,77 @@
...
+
+        mutable results = [Zero, size = numBits];
+
+        for k in 0..numBits - 1 {
+            statePrep(system);
+            // Compute accumulated phase correction from previously measured bits.
+            mutable accumulatePhase = 0.0;
+            for j in 0..k - 1 {
+                if results[j] == One {
+                    set accumulatePhase += 2.0 * PI() / IntAsDouble(1 <<< (k - j + 1));
```
wavefunction91: Could we preserve Adaptive_RI context creation when adding this module? With this change, `create_qsharp_context(TargetProfile.Adaptive_RI)` raises `UseOfDynamicDouble` at this line before any circuit builder or `combine_iterations` setting is used. The same Q# project without this new file initializes successfully. This breaks existing RI workflows, not just combined IQPE, and the builder's profile guard cannot catch a failure during context initialization.

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,261 @@
...
+    loss: dict[str, int] = {}
+    for entry in raw_results.values():
+        outcome = entry["outcome"]
+        count = entry["count"]
+        outcome_bits = outcome if isinstance(outcome, (list | tuple)) else (outcome,)
+        if "-" in outcome_bits:
+            key = "".join("L" if bit == "-" else str(bit) for bit in reversed(outcome_bits))
+            loss[key] = loss.get(key, 0) + count
+        else:
+            key = "".join(str(bit) for bit in reversed(outcome_bits))
```
wavefunction91: Could SDK error outcomes be handled explicitly before converting entries to measurement counts? Azure Quantum 3.12.0's `Job.get_results_histogram()` can return an outcome such as `{"Error": {"Code": "0x20", "Name": "TestErrorThirtyTwo"}}`. The current conversion turns that dictionary into a string key in `bitstring_counts`, reporting a failed shot as a clean measurement even though `CircuitExecutorData` specifies 0/1 bitstrings. I reproduced this with the installed SDK and this parser. Raising with the service error, or another explicit error path, would avoid misclassifying the failed shot.

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,261 @@
...
+
+        destination = Path(output_dir).expanduser().resolve()
+        destination.mkdir(parents=True, exist_ok=True)
+        saved: list[str] = []
+        for name in names:
+            # Attachment names come from the service, so keep them from escaping output_dir.
+            local_path = (destination / Path(name).name).resolve()
+            if local_path.parent != destination:
+                raise ValueError(f"refusing to write attachment outside the output directory: {name!r}")
+            local_path.write_bytes(job.download_attachment(name))
```
wavefunction91: Could we detect destination collisions before writing attachments, or preserve a collision-free mapping? Distinct names such as `trace/rawOutputData` and `diagnostics/rawOutputData` both become `output_dir/rawOutputData` here. In a local SDK-boundary reproduction, the second download overwrote the first while `saved_attachments` reported the same path twice. The traversal protection is useful, but flattening names should not silently discard a requested artifact.

python/src/qdk_chemistry/utils/qsharp/src/CombinedIterationPhaseEstimation.qs
```diff
@@ -0,0 +1,77 @@
...
+            []
+        } else {
+            qs[1 + Length(systems)..Length(qs) - 1]
+        };
+        let allTargets = system + ancillas;
+
+        mutable results = [Zero, size = numBits];
+
+        for k in 0..numBits - 1 {
+            statePrep(system);
```
wavefunction91: Is re-preparing the system each round intentional, and what overlap assumption should the combined mode require? For the commuting two-qubit example

```text
H = (-5 II + 11 IZ - 15 ZI + ZZ) / 64
t = 2*pi, num_bits = 4
psi = sqrt(3/8)|00> + (1/2)|01> + sqrt(3/8)|10>
```

the combined distribution has mode `1010`, giving phase `5/8` and energy `3/8`, which is absent from the full spectrum `{-1/2, -1/8, 0, 5/16}`. The local QDK execution confirms this; standard QPE returns only the prepared eigenphases. This is also a limitation of the legacy one-shot protocol, not a claimed regression relative to it. The public contract says "significant overlap" but does not specify a threshold. Could we preserve the system across rounds, or explicitly document and test the intended validity assumptions rather than presenting this as generic eigenvalue extraction?

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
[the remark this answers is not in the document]
RushiGong: The highest power measures the LSB. `RunFullIQPE` returns LSB-first, and reversing it gives the MSB-first string `_run_single_circuit` expects.

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,261 @@
...
+            raise ValueError("Azure Quantum target cannot be resolved; set " + ", ".join(missing))
+
+        qir_string = str(circuit.get_qir())
+        Logger.debug("QIR compiled")
+
+        workspace = Workspace(
+            subscription_id=coordinates["subscription_id"],
+            resource_group=coordinates["resource_group"],
+            name=coordinates["workspace_name"],
+            location=coordinates["location"],
```
rigidit, line 190: nit: location can be skipped when subscription, rg and name is specified

python/src/qdk_chemistry/plugins/azure_quantum/circuit_executor.py
```diff
@@ -0,0 +1,261 @@
...
+            shots=shots,
+            input_data=qir_string,
+            input_data_format="qir.v1",
+            output_data_format="microsoft.quantum-results.v2",
+            input_params=input_params,
+        )
+        Logger.debug(f"Job submitted: {job.id}")
+
+        timeout = self._settings.get("timeout_secs")
+        raw_results = job.get_results_histogram(timeout_secs=timeout)
```
rigidit, line 213: nit: you can make the job result waiting more explicit by calling `job.wait_until_completed(timeout_secs=timeout)` before calling `get_results_histogram`, otherwise it is called inside when you pass timeout_secs to get_results_histogram

rigidit (approved):

PR #704, opened 2026-09-03
Docs: automate versioned publishing and adopt PyData theme

### LINKED ISSUE #576: Automate the docs build upon release
As we did for packaging the examples in #537, we should automate the docs build and push upon release

eimrek:
Preview: https://eimrek.github.io/qdk-chemistry  
The preview's `2.2` and `2.3` releases are demonstrations only.

Adds automated, versioned documentation publishing and adopts **PyData Sphinx Theme** for its built-in version switcher.

## Publishing

- **`/dev/`** reuses the documentation artifact from successful `Build and Test` push or manual runs on this repository's `main`.
- **`/X.Y/`** contains a release's documentation. Newer patches replace that minor version's content. Manual runs can apply compatible content corrections via `docs_ref`, without a new package release.
- **`/stable/`** is a copy of the newest published final release. Publishing an older maintenance release does not downgrade it.

Release builds use the tagged documentation by default; `docs_ref` can supply corrected `docs/` content. The API sources and dependency requirements remain tied to the release, and builds install the **exact matching package from PyPI**. New themes or build dependencies require a new release.

Package publication is separately approval-gated; if the package is not yet available, rerun `Docs` after it is published. Historical releases retain their original documentation configuration and theme.

## Note

- Builds currently occupy roughly 50–60 MiB each; `/stable/` duplicates the newest release. Publication warns above 800 MiB and fails above GitHub Pages' 1 GiB limit.

## Migration Checklist

Migration of the existing flat `gh-pages/docs` output remains manual.

- [ ] Record the current `gh-pages` SHA for rollback.
- [ ] Publish `v2.1.0` through `Docs` to seed `/2.1/` and `/stable/`.
- [ ] Remove legacy flat output, preserving `.nojekyll`, `index.html`, `404.html`, `switcher.json`, `/dev/`, `/stable/`, and all `X.Y/` directories; commit and push to `gh-pages`.
- [ ] Confirm `/dev/` publishes from a successful `main` build.
- [ ] Verify root and legacy deep-link redirects to `/stable/`, and version switching on builds using the new theme.

Closes #576


docs/source/conf.py
```diff
@@ -41,7 +41,25 @@
...
+    f"{_release_match.group(1)}.{_release_match.group(2)}"
+    if _release_match
+    else release
+)
+
+# Canonical URL of this build; overridden per version directory when publishing.
+html_baseurl = os.environ.get(
+    "QDK_CHEMISTRY_DOCS_BASE_URL", "https://microsoft.github.io/qdk-chemistry/stable/"
+)
+
```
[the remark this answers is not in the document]
eimrek: fixed

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,308 @@
+name: Docs
+
+on:
+  # Build and Test already builds dev from main, so reuse that trusted artifact.
+  # Branch filters cannot take expressions, so this one name is hardcoded; the
+  # job conditions below compare against the repository's default branch.
+  workflow_run:
+    workflows: [Build and Test]
```
nabbelbabbel, line 6: This is a pipeline that publishes, right? I think this label is wrong then.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,308 @@
...
+    types: [completed]
+    branches: [main]
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: Immutable release tag to build (vX.Y.Z)
```
nabbelbabbel: What does the tag of latest head on main resolve to? Does it have this format?

.github/workflows/docs.yaml
eimrek: This part means that this `Docs` workflow is automatically called whenever the `Build and Test` workflow completes on `main`. I could adapt the comment to make this more clear.

.github/workflows/docs.yaml
eimrek: Currently only a release tag is supported here (so, no branches).

The reasoning is that I don't want to build the whole library in this workflow. Therefore, it either takes the library from 1) `Build and Test` workflow (for `dev`), or
2) from PYPI, for a release.

for option 1), one should just manually re-run the whole `Build and Test` (which makes sure the docs artifact is stored), and that triggers this Docs workflow automatically.

Alternatively, I could also allow `main` here, but there is a chance that the docs artifact is not available any more, so i thought it just safer to have users avoid this.

And the main purpose of this manual run is to backfill or fix docs builds for already released versions.

happy to discuss.

.devcontainer/scripts/parallelism.sh
```diff
@@ -8,7 +8,7 @@
 #
 # parallel_jobs_for_memory <memory_per_job_gb>
 #
-# Compute a safe parallel job count based on available RAM and CPU cores.
+# Compute a safe parallel job count based on total RAM and CPU cores.
```
nabbelbabbel: If a machine is under heavy load, then running an additional build will kill things if total RAM is compared, please keep this conservative and work off of the available free slice.

eimrek: sorry. this was a change unrelated to this PR that slipped through. I reverted the comment. The script does, however, check total RAM, so that should be fixed separately.

.github/scripts/docs_site.py
```diff
@@ -0,0 +1,659 @@
...
+    missing_keys = _BUILD_INFO_KEYS - info.keys()
+    if missing_keys:
+        info_file = directory / BUILD_INFO_NAME
+        _fail(
+            f"build metadata in {info_file} is missing: {', '.join(sorted(missing_keys))}"
+        )
+    return info
+
+
+def _run_git(repository: Path, *arguments: str) -> subprocess.CompletedProcess[str]:
```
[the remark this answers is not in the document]
eimrek: False positive: _fail() is annotated NoReturn and always raises SystemExit(1), so this branch cannot fall through. Verified with an injected OSError.

PR #705, opened 2026-09-03
Add bounded CLI and MCP fuzz testing

nabbelbabbel:
## Summary

- Add bounded Hypothesis coverage for CLI parsing, JSON, and dotted overrides.
- Fuzz filesystem-facing MCP tools with direct containment assertions.
- Generate valid MCP calls from published schemas and exercise in-process dispatch.
- Verify stdio recovery from malformed JSON, invalid UTF-8, truncation, and nested payloads.
- Add a weekly/manual 5,000-example campaign with a persisted Hypothesis corpus.
- Reject malformed or scalar-conflicting `--set` paths discovered by fuzzing.

## CI behavior

Regular CI runs 100 examples per property through the existing `test` extra.

The extended workflow runs every Monday at 06:00 UTC or manually with a configurable example count. It uses two build jobs and saves minimized Hypothesis examples.

### DIFF
--- a/.github/workflows/fuzz.yaml
+++ b/.github/workflows/fuzz.yaml
@@ -1,0 +1,111 @@
+name: Extended Fuzz Tests
+
+on:
+  schedule:
+  - cron: 0 6 * * 1
+  workflow_dispatch:
+    inputs:
+      examples:
+        description: Hypothesis examples per property
+        required: false
+        default: 5000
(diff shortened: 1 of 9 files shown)


nabbelbabbel:
@wavefunction91 per offline discussion here a draft and thus a place to discuss fuzz testing.

PR #640, opened 2026-08-13
Use zero-based orbital cube file labels

ConradJohnston:
## Summary

- label generated orbital cube files with zero-based indices by default
- update tutorial and visualization consumers to parse zero-based labels without an offset
- add regression coverage for in-memory labels and on-disk filenames

## Validation

- `python -m ruff check python/src/qdk_chemistry/utils/cubegen.py python/tests/test_utils_cubegen.py examples/utils/viz_utils.py docs/source/_static/examples/python/tutorial_choose_active_space.py`
- `python -m py_compile python/src/qdk_chemistry/utils/cubegen.py python/tests/test_utils_cubegen.py examples/utils/viz_utils.py docs/source/_static/examples/python/tutorial_choose_active_space.py`
- focused runtime verification that orbital index 0 produces `orbital_0000.cube`

Closes #639

PR #717, opened 2026-09-07
Harden Windows wheel runtime prerequisite

wavefunction91:
## Summary

- Document the Microsoft Visual C++ v14 Redistributable as the Windows wheel runtime prerequisite.
- Replace the generic native-extension loader failure with an actionable Windows installation message.
- Preserve the policy-default `ms-ensureconda` test environment on x64 and the existing venv exception on ARM64.
- Add a binary-only base installation and root-package import before test extras.
- Audit every PE import and reject dependencies outside Windows, CPython, the exact release VC++ v14 runtime set, or the wheel itself.

## Rationale

The 2.2.0 Windows wheels link against `MSVCP140.dll` and `VCRUNTIME140*.dll`, but existing CI did not state or validate that runtime boundary. A clean Windows ARM64 VM therefore failed during `import qdk_chemistry` until the VC++ Redistributable was installed.

This keeps `/MD` to match CPython and relies on Microsoft's centrally serviced Redistributable. It does not statically link the CRT or ship a private runtime copy inside the wheel.

## Validation

- Azure release-wheel pipeline run [179669](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=179669&view=results) succeeded for CPython 3.13 on Windows x64 and ARM64.
- x64 used `ms-ensureconda==2026.6.1`; ARM64 retained the existing venv path.
- Both generated wheels passed the exact PE dependency audit and binary-only base install/root import.
- Windows ARM64: 3007 passed, 369 skipped.
- Windows x64: 3220 passed, 188 skipped.
- The new import-diagnostic test passed on both architectures.
- All staged pre-commit hooks passed.

The separately reported active-space tutorial cube-label issue was already fixed by #640 and is not changed here.

### DIFF
--- a/python/src/qdk_chemistry/__init__.py
+++ b/python/src/qdk_chemistry/__init__.py
@@ -51,12 +51,12 @@ def _import_core() -> None:
     try:
         importlib.import_module("qdk_chemistry._core")
     except ImportError as exc:
-        if _sys.platform == "win32" and "DLL load failed" in str(exc):
+        if _sys.platform == "win32":
             raise ImportError(
                 "QDK/Chemistry requires the current Microsoft Visual C++ v14 "
-                "Redistributable on Windows. Install it for "
-                "your Python architecture, restart Python, and try again: "
-                "https://learn.microsoft.com/cpp/windows/latest-supported-vc-redist"
+                "Redistributable on Windows. Install the x64 package (which also "
+                "includes ARM64), restart Python, and try again: "
+                "https://aka.ms/vc14/vc_redist.x64.exe"
             ) from exc
         raise
 
--- a/python/src/qdk_chemistry/__init__.py
+++ b/python/src/qdk_chemistry/__init__.py
@@ -50,6 +50,8 @@ def _import_core() -> None:
     """Import the native extension with an actionable Windows failure message."""
     try:
         importlib.import_module("qdk_chemistry._core")
+    except ModuleNotFoundError:
+        raise
     except ImportError as exc:
         if _sys.platform == "win32":
             raise ImportError(
--- a/python/src/qdk_chemistry/__init__.py
+++ b/python/src/qdk_chemistry/__init__.py
@@ -45,12 +45,30 @@ import subprocess
 import sys
 import warnings
 
+
+def _import_core() -> None:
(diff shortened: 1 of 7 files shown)


lorisercole (dismissed):

python/src/qdk_chemistry/__init__.py
```diff
@@ -45,12 +45,30 @@
 import sys
 import warnings
 
+
+def _import_core() -> None:
+    """Import the native extension with an actionable Windows failure message."""
+    try:
+        importlib.import_module("qdk_chemistry._core")
+    except ImportError as exc:
+        if _sys.platform == "win32" and "DLL load failed" in str(exc):
```
billti: Does the message definitely contain "DLL load failed" across all platforms and locales?

python/src/qdk_chemistry/__init__.py
```diff
@@ -45,12 +45,30 @@
...
+            raise ImportError(
+                "QDK/Chemistry requires the current Microsoft Visual C++ v14 "
+                "Redistributable on Windows. Install it for "
+                "your Python architecture, restart Python, and try again: "
+                "https://learn.microsoft.com/cpp/windows/latest-supported-vc-redist"
```
billti: It might be nice to directly link to the redist for the platform to save them having to read the docs and choose the right package. There are permalinks for x64 and arm64, e.g.

```python
from sysconfig import get_platform

arch = "arm64" if get_platform() == "win-arm64" else "x64"
permalink = f"https://aka.ms/vc14/vc_redist.{arch}.exe"
```

python/src/qdk_chemistry/__init__.py
billti: (ChatGPT says the substring in the message is hard-coded English in the Python codebase, but still recommends against string-matching on it as it isn't part of any "contract" and may change across versions. I tend to agree, but could be pursuaded either way).

billti (dismissed):

python/src/qdk_chemistry/__init__.py
wavefunction91: That's a good point, will fix!

agamshayit-ms (approved):

PR #718, opened 2026-09-08
Fix local Copilot plugin installation and MCP Apps

nabbelbabbel:
## Summary

- Support Copilot CLI live local plugins when no `installed-plugins` copy is created.
- Deploy workspace skills and MCP configuration without modifying the source plugin.
- Register visualizers through MCP 2.0's first-party `Apps` extension.
- Replace tokenized UI resources with static MCP App resources receiving structured tool results.
- Clarify that workspace binding occurs once per MCP server process.

## Fixes

Local `qc plugin install --target-dir` incorrectly failed after Copilot successfully enabled a local plugin because the installer expected a copied plugin directory.

Visualization tools also advertised tokenized resource templates that VS Code could not resolve, causing `Unknown resource` errors instead of mounting widgets.


python/src/qdk_chemistry/ui/visualization.py
```diff
@@ -201,141 +222,109 @@ def _build_html(
...
+    let minY=Math.min(...points.map(point=>point.y)),maxY=Math.max(...points.map(point=>point.y));
+    const padX=(maxX-minX)*.08||1,padY=(maxY-minY)*.08||1;minX-=padX;maxX+=padX;minY-=padY;maxY+=padY;
+    const left=75,top=45,width=595,height=340,tx=value=>left+(value-minX)/(maxX-minX)*width,ty=value=>top+height-(value-minY)/(maxY-minY)*height;
+    add('rect',{x:left,y:top,width,height,fill:'var(--color-background-secondary,#222)',stroke:'var(--color-border-primary,#666)'});
+    add('text',{x:350,y:28,'text-anchor':'middle','font-size':16},data.title||'Scatter Plot');
+    add('text',{x:372,y:430,'text-anchor':'middle','font-size':12},data.x_label||'X');
+    add('text',{x:18,y:215,'text-anchor':'middle','font-size':12,transform:'rotate(-90 18 215)'},data.y_label||'Y');
+    const colors=['#4f8cff','#e05263','#38a169','#d69e2e','#805ad5','#319795'];
+    for(const point of points){const dot=add('circle',{cx:tx(point.x),cy:ty(point.y),r:5,fill:colors[point.group%colors.length]});const title=document.createElementNS(ns,'title');title.textContent=`${point.label}${point.label?' - ':''}${point.rawX}, ${point.rawY}`;dot.append(title);}
```
wavefunction91: Could we adapt the existing scatter renderer to receive tool results rather than replace it? This version drops the numeric axis ticks and named-series legend that the previous renderer provided, leaving no numeric axis scale or visible key for the series colors. The static-resource change does not require losing those features. `_build_plotly_html` is also still present below, but its only caller has been removed. Preserving its rendering behavior while changing the data delivery, then removing the obsolete wrapper, would keep this focused on the MCP Apps fix.

wavefunction91 (approved):

